Issue #331, opened 2026-08-05
[Bug]: bundled BCL assemblies duplicate Unity 6000's own at a different major version, breaking consumer compilation


wallstop:
## Summary

The package ships four BCL assemblies in `Runtime/Binaries/`. **Three of them duplicate assemblies Unity 6000.x provides itself, at a different major version.** Which copy the compiler binds is left to Unity's dedup, and when it resolves the family inconsistently the package fails to compile in the consuming project:

```text
Packages\com.wallstop-studios.unity-helpers\Runtime\Core\Serialization\JsonConverters\Vector4Converter.cs(15,57):
error CS0012: The type 'JavaScriptEncoder' is defined in an assembly that is not referenced.
You must add a reference to assembly 'System.Text.Encodings.Web, Version=9.0.0.0, ...'.

Packages\com.wallstop-studios.unity-helpers\Runtime\Core\Extension\EnumExtensions.cs(526,30):
error CS0103: The name 'Unsafe' does not exist in the current context
```

Reported against `3.5.1` on Unity `6000.5.2f1`, intermittent.

## Measured, from the live editor

Probed a Unity `6000.5.2f1` editor over the MCP bridge.

**What we ship** (`Runtime/Binaries`, all from the same NuGet drop, `9.0.9` → assembly version `9.0.0.0`):

| Assembly | Our version | Unity 6000.5.2f1 ships it? |
| --- | --- | --- |
| `System.Text.Json` | 9.0.0.0 | **Yes — 8.0.0.0** |
| `System.Text.Encodings.Web` | 9.0.0.0 | **Yes — 8.0.0.0** |
| `Microsoft.Bcl.AsyncInterfaces` | 9.0.0.0 | **Yes — 8.0.0.0** |
| `System.IO.Pipelines` | 9.0.0.0 | No |
| `System.Runtime.CompilerServices.Unsafe` | **not shipped** | Yes — 6.0.0.0 |

**What the editor actually binds** for `WallstopStudios.UnityHelpers`:

```text
C:/…/6000.5.2f1/Editor/Data/BCLExtensions/TargetingPacks/netstandard2.1/ref/Microsoft.Bcl.AsyncInterfaces.dll
C:/…/6000.5.2f1/Editor/Data/BCLExtensions/TargetingPacks/netstandard2.1/ref/System.Runtime.CompilerServices.Unsafe.dll
[...]

Issue #326, opened 2026-08-01
chore/research: CI Grouping Optimization


wallstop:
See https://github.com/Ambiguous-Interactive/DxMessaging/issues/337, but for this repo

wallstop:
Measured this against Unity Tests run [`30227874347`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/30227874347) — the last full green run on `main` (commit `3d4bdd1c`). Posting the numbers before anyone regroups jobs, because they point somewhere other than grouping.

## Where the 108 minutes go

Wall clock 108 min. Sum of job durations 113 min. There are **two** self-hosted Unity runners (`ELI-MACHINE`, `DAD-MACHINE`), so the floor for 14 Unity jobs is ~57 min of perfectly packed runner time.

| Job | Duration | Runner |
| --- | ---: | --- |
| 2022.3.45f1 editmode | 5 min | ELI |
| 2021.3.45f1 editmode | 9 min | DAD |
| 6000.3.16f1 editmode | 10 min | DAD |
| 6000.5.2f1 editmode | 13 min | ELI |
| 2021.3.45f1 playmode | 3 min | ELI |
| 2022.3.45f1 playmode | 4 min | DAD |
| 6000.3.16f1 playmode | 3 min | ELI |
| 6000.5.2f1 playmode | 3 min | ELI |
| standalone 2022.3.45f1 | 8 min | ELI |
| standalone 2021.3.45f1 | 11 min | DAD |
| standalone 6000.3.16f1 | 11 min | DAD |
| standalone 6000.5.2f1 | 8 min | ELI |
| 6000.3.16f1 editmode SINGLE_THREADED | 11 min | DAD |
| 6000.3.16f1 playmode SINGLE_THREADED | 10 min | ELI |
| unitypackage-smoke | 4 min | hosted |

## The largest single line item is idle time, not setup

Sorting the jobs by start time leaves gaps where **both** runners sat idle:

| Gap | Duration |
| --- | ---: |
| 00:44:42 → 00:59:19 (inside the fast tier) | **14.6 min** |
| 01:17:53 → 01:24:23 (inside the fast tier) | 6.5 min |
| 01:57:44 → 02:07:11 (standalone tier → SINGLE_THREADED tier) | **9.5 min** |

That is **~31 min of the 108**, or 28%, with no job running. Grouping editmode and playmode into one job per Unity version cannot recover any of it — the gaps are scheduling, not setup.

[...]

PR #335, opened 2026-08-05
fix(ci): restore Unity validation on pull requests, and green main

### LINKED ISSUE #332: [Bug]: CI Does not run Unity Tests
### Description

Latest CI/PRs are not running any Unity tests - this is incredibly broken and we need to remedy this

### Steps to Reproduce

Open a PR with unity changes

### Expected Behavior

Relevant Unity CI is ran

### Actual Behavior

No Unity CI is ran

### Severity

Critical (crash, data loss)

### Unity Version

Unity 6.2 (6000.2)

### Package Version

3.5.1

### Platform

- [ ] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Regression

### Additional Context

_No response_

wallstop:
`main` went red within twenty minutes of #328 merging, on two independent causes, and #332 — "CI does not run Unity tests" — turned out to be a third, structural one. All three are fixed here. Both regressions were parts #328 itself flagged as unverified, and the matrix answered both on its first run.

## 1. #332 — every Unity leg reports failure without running a test

Not a runner outage and not a licensing problem: a **superseded** pull-request run keeps its place in the self-hosted queue.

Each licensed leg guards itself with `require-current-pr-head`, twice — before setup and immediately before lock acquisition. Both fire only **after** the leg has been dispatched, and a dispatch means waiting in line for the single self-hosted Unity seat. So a stale iteration queues eight legs, each of which eventually gets a runner, discovers the head has moved, and exits in ~12 seconds having done nothing.

That would be merely wasteful if the run could be cancelled. It cannot: this workflow's concurrency group is deliberately `cancel-in-progress: false`, because a leg may be holding the org Unity license and cancellation would skip its return and lock-release cleanup — a constraint `test-pr-workflow-concurrency.ps1` enforces. The successor run therefore sits `pending` behind the stale one until its queue drains.

Measured on #328 itself:

| Run | Created | Outcome |
| --- | --- | --- |
| `30960311691` | 23:32 | 2 legs failed on the head guard |
| `30962575177` | 00:12 | all 8 legs failed on the head guard |
| `30963202461` | 00:24 | 6 legs failed on the head guard, **2 still queued 45 min later** |
| `30964852134` | 00:56 | `pending` behind the above; never started before the merge |

#328 merged with no Unity validation at all.

**Fix.** `matrix-config` — hosted, ~15 s — resolves "has this PR moved on?" once and exposes it as an output. The four licensed tiers skip when it says yes, so a superseded run finishes in seconds and hands the concurrency group to the run that matters. The per-leg guards stay as defense in depth for a push that lands after that step has already answered.

Three properties, each with a mutation proof in `test-unity-workflow-matrix-contract.ps1`:

- **Fails open.** A non-pull-request run, an unresolvable head, and a head that has not moved all report `superseded=false`. An API hiccup costs a redundant run, never a skipped validation. The contract binds value to branch position, not to occurrence counts — swapping which branch writes which value leaves the counts identical.
- **Validates shape, not emptiness.** `gh api --jq '.head.sha'` prints the literal `null` and exits 0 when the field is absent; `null` is non-empty and unequal to any real SHA, so an emptiness check would have read an API anomaly as a moved head and skipped every tier.
- **Supersession waives only the four licensed results.** `matrix-config` and `runner-preflight` still have to pass. Both run to completion regardless — `matrix-config` goes on to lint the test-project module manifest — and a failed job still publishes its outputs, so short-circuiting on the output alone would report green for a bogus UPM module id or an offline runner fleet.

A superseded run reports Unity CI Success as a clean no-op rather than a failure. Its check belongs to the head SHA it was queued for, so it can neither gate nor authorize a merge of the current head; reporting failure would mark "you pushed again" as a defect.

## 2. Main red — the bundled BCL boundary was one release too wide

#328 constrained `System.Text.Json`, `System.Text.Encodings.Web`, and `Microsoft.Bcl.AsyncInterfaces` to `!UNITY_6000_0_OR_NEWER` from a `6000.5.2f1` probe, and said in its own commit message that the `6000.0` boundary was unverified. **`6000.3.16f1` then stopped compiling entirely** — `CS0246` on every `Utf8JsonReader`, `Utf8JsonWriter`, `JsonSerializerOptions`, `JsonConverter<>`, `JsonInclude`, and `JsonIgnore` in the package. Excluding our copies on an editor that has none of its own leaves nothing behind.

Probed the live editor rather than guessing the replacement. Unity supplies these from `Editor/Data/BCLExtensions/TargetingPacks/netstandard2.1/ref`. Across all 21 Unity editors installed on the probe machine:

| Editors | `BCLExtensions` |
| --- | --- |
| 2022.3.51, 2022.3.62, 2023.2.5, 2023.2.13 | absent |
| 6000.0.23 / .28 / .32 / .58 / .65, 6000.1.3 / .14 / .15, 6000.2.1 / .6 / .8 | absent |
| 6000.3.0 / .10 / .14, 6000.4.5, 6000.4.6 | absent |
| **6000.5.2** | **present** |

So the boundary is **6000.5**. Also confirmed `UNITY_6000_5_OR_NEWER` is a real define there by reading the editor's own compilation define set — a `#if` probe inside a dynamically compiled assembly reports every Unity define as undefined, including `UNITY_EDITOR`, so it cannot answer this. The matrix now pins the boundary from both sides: `6000.3.16f1` proves the package's copies are still required, `6000.5.2f1` proves excluding them is what stops the `CS0012`/`CS0103` mixed resolution in #331.

## 3. Main red — a new fixture skipped the cross-fixture tripwire

`AssetChangeDetectionEnablementTests` landed in #328 without calling `AssetPostprocessorTestHandlers.AssertCleanAndClearAll()`. It drives `DetectAssetChangeProcessor` directly, one of the handler-involvement tokens `AssetContextFixturesCallCrossFixturePollutionTripwire` gates on, so the contract matched it and failed — one failure out of 4639 tests, identically on the 2021.3, 2022.3, and 6000.5.2 EditMode legs.

Swept the rest of `Tests/Editor/`: no other fixture is missing the call, and none places it after `base.BaseSetUp(`.

## 4. The stuck-job watchdog has never recovered a run

Found while waiting on this PR's own CI, and it corrects something I posted on #332 earlier: I claimed #328's watchdog fix "works". It does not. Reading the 02:20 UTC run's own log:

```text
WARN: org-level runner list unavailable (likely 403). Falling back to repo runners.
ERROR: repo-level runner list also failed; cannot evaluate idle runners. No action issued.
```

That is step 3 of the audit, before any queued run is examined, and the handler exits **0** there. `orgs/{owner}/actions/runners` needs `admin:org`, which the workflow's `GITHUB_TOKEN` does not have; the repo-scoped fallback does not list org-level runners, which is the only kind this org has. The audit has therefore never evaluated a single queued run, and has reported success on every cycle while doing nothing.

#328's CRLF fix moved the failure from "red at checkout" to "green and blind" — strictly worse, because a green check is what stops anyone from looking. Run `30566478129` sat queued from July 30 to August 5, six days, against a five-minute schedule. It dispatched on its own.

**Fix.** The org query now uses the build-lock reader App (`BUILD_LOCK_READER_APP_ID`), the same identity `check-unity-runner-availability` already uses to decide whether a Unity runner is online — exactly the grant this needs. It falls back to the workflow token, then repo scope, and a missing App credential never fails the audit on its own. An unreadable inventory now **fails closed**.

If the App turns out not to hold read access to organization self-hosted runners, the watchdog will go red every five minutes with a message naming that grant. That is intended — it is currently broken and silent, and the remedy is then a permissions change, not a code change.

## 5. `validate:prepush` could not pass locally

`test-portable-cleanup-classifier.js` hard-required `BUILD_LOCK_POLICY_ROOT`, a separate checkout of the central build-lock policy that only CI provides. Unset, the path resolved to the repo root, the classifier runtime was missing, and the assertion took down the gate CLAUDE.md names as the pre-push requirement on every developer machine. It now skips with a message naming the variable when the checkout is absent, and fails closed under GitHub Actions, where its absence means the checkout step broke.

## 6. #331 was misdiagnosed — the real cause is name deduplication

The `CS0012 JavaScriptEncoder` in #331 is not this package competing with Unity 6000. Confirmed against a live Unity `6000.4.6f1` project that reproduces it.

Unity deduplicates precompiled assemblies **by simple name**: exactly one file wins `System.Text.Encodings.Web` project-wide. Ours is auto-referenced (`isExplicitlyReferenced: 0`), so when ours wins, everything gets it. `com.unity.ai.assistant` ships all four names under `Plugins/Shared/` with `isExplicitlyReferenced: 1`. When one of *those* wins, the winner is auto-referenced by nothing and our copy is gone — so the name is referenced by **zero** assemblies.

```text
GetPrecompiledAssemblyPathFromAssemblyName
  System.Text.Json           -> unity-helpers/Runtime/Binaries/System.Text.Json.dll   (ours won)
  System.Text.Encodings.Web  -> com.unity.ai.assistant/Plugins/Shared/...             (theirs won)

editor assemblies referencing System.Text.Json = 168 / 245
editor assemblies referencing Encodings.Web    =   0 / 245
```

That asymmetry is why the error names `Encodings.Web` and not `System.Text.Json`.

The remedy is the AI Assistant's own opt-out defines. Adding `EXCLUDE_TEXT_ENCODINGS_WEB`, `EXCLUDE_BCL_ASYNCINTERFACES`, and `EXCLUDE_COMPILER_SERVICES_UNSAFE` took `Encodings.Web` from **0/245 to 221/221** and recompiled the project clean, with all ten `Unity.AI.Assistant.*` assemblies still building — they resolve against our copy, the same `9.0.0.0` assembly.

Two installed packages contesting a name is not something this package can settle alone, so this ships as `docs/guides/bundled-assembly-conflicts.md` — symptom, mechanism, how to ask Unity which file won a name, and the fix — plus a corrected CHANGELOG entry.

## 7. CHANGELOG entries are now required to be short

Entries had drifted into RCA territory: mechanism, Unity internals, run IDs, "verified on…". That is the right detail for a commit body and the wrong detail for someone skimming a release to find out whether anything affects them.

The rule now lives in the three places an agent reads — `.llm/context.md` Documentation Rules, `ship-changes` Step 4, and `update-documentation`'s entry-writing section, which gains explicit rules and a worked counter-example that is accurate, well written, and still wrong for a CHANGELOG. The Unreleased section is rewritten to match; the worst entry went from 147 words to two sentences.

Four entries describing Unity CI provisioning and license-return internals were dropped: the same file already says CI/CD and infrastructure do not belong in a CHANGELOG, and users do not run this package's CI.

## Verification

- `test-unity-workflow-matrix-contract.ps1` passes, plus eight mutation proofs that each fail it (ungated tier, fail-closed detection, missing shape check, waived hosted gates, aggregator short-circuit removed, fail-open inversion, watchdog exit-0 regression, watchdog reader token unwired)
- `actionlint`, `yamllint`, `npm run lint:yaml`, `npm run test:shell-portability` clean
- `npm run validate:prepush` green end to end
- Both `main` failures confirmed fixed on CI: 6000.3.16 EditMode went from 1896 `CS0246` to green, and the three tripwire legs (2021.3, 2022.3, 6000.5.2) went green
- BCL boundary and define availability probed against a live `6000.5.2f1` editor

Unity licensing is not configured on this machine, so the C# and `.meta` changes were not compiled or run locally — CI is the gate for those, and for both of this session's regressions it was the only thing that could have been.

Closes #332
Refs #331, #326

---

> [!NOTE]
> **Medium Risk**
> Touches licensed Unity CI gating and org-runner watchdog credentials—high operational impact—but changes are contract-tested and the BCL boundary is a targeted compile fix for Unity 6000.3–6000.4 vs 6000.5+.
> 
> **Overview**
> Restores **PR Unity CI** after stale runs blocked the single self-hosted seat: **`matrix-config`** now detects a **superseded PR head** on a hosted runner (fail-open), skips all **licensed Unity tiers** when the head moved, and **`unity-ci-success`** treats that case as a clean no-op while still requiring **matrix-config** and **runner-preflight** to pass.
> 
> Fixes **main compile breaks** by tightening bundled **`System.Text.Json`** / related DLL **`.meta`** constraints to **`!UNITY_6000_5_OR_NEWER`** (Unity 6000.0–6000.4 still need the package copies). Adds **`docs/guides/bundled-assembly-conflicts.md`**, shortens **CHANGELOG** entries, and documents brevity rules in **`.llm`**.
> 
> **Stuck-job watchdog** mints a **build-lock reader App** token for org runner inventory and **fails the workflow** when inventory cannot be read (no more silent green audits).
> 
> Also: **`AssetChangeDetectionEnablementTests`** calls the cross-fixture pollution tripwire; **`test-portable-cleanup-classifier.js`** skips locally when **`BUILD_LOCK_POLICY_ROOT`** is unset; workflow **contract tests** cover supersession and watchdog behavior; issue templates add version **1.0.1**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84035580cfaf075002d578730289878f7521f7a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wallstop:
Two pieces of live evidence from this PR's own runs, which is a better demonstration than the contract test.

**The mechanism, reproduced and then broken by hand.** The intermediate push (`00a39e30`) started Unity run `30966586037`. When I force-pushed `555115b1`, run `30969689881` was created and immediately went `pending` — because `30966586037` still had three legs sitting in the self-hosted queue and the concurrency group cannot cancel in progress. I cancelled the superseded run manually (safe: every leg was either completed or still queued, so none held the org Unity license), and `30969689881` moved `pending -> in_progress` within twenty seconds. That manual step is exactly what this PR automates, minus the manual part and minus the ~45 minutes of queue drain.

**Red -> green on the tripwire fix.** Run `30966586037` carried the tripwire fix but not yet the BCL fix, so it isolates the two causes cleanly:

| Leg | Before (`30965190533`, main) | After (`30966586037`) |
| --- | --- | --- |
| 2021.3.45f1 editmode | failure | **success** |
| 2022.3.45f1 editmode | failure | **success** |
| 6000.5.2f1 editmode | failure | **success** |
| 2021.3.45f1 playmode | success | success |
| 6000.3.16f1 editmode | failure | failure (BCL, fixed in `2425b0fb`) |

So the single `AssertCleanAndClearAll()` call accounts for all three EditMode failures, and the remaining `6000.3.16f1` failure is entirely the define constraint — 1896 `CS0246` and no `results.xml`, versus 4639 tests collected on the legs that compile.

For the third cause: the last fully green `main` run (`30801483300`, Aug 3) had all 18 jobs green including the standalone and SINGLE_THREADED tiers, which have not been reached since. This PR's run is the first opportunity to re-exercise them.

Issue #327, opened 2026-08-04
[Bug]: DetectAssetChangeProcessor's all-types reflection scan crashes Unity in batchmode — the guard that exists for it only covers play mode


wallstop:
### Description

`DetectAssetChangeProcessor.OnPostprocessAllAssets` runs `BuildWatchers`, an **all-types / all-methods
reflection scan**, on every asset postprocess pass. In a `-batchmode` editor run this crashes Unity with
`STATUS_ACCESS_VIOLATION` (`0xC0000005`) inside Mono's `GetMethodsByName_native`.

**The package already knows this scan can do exactly this, and guards against it — but on the wrong
condition.** `OnPostprocessAllAssets` ([`Editor/AssetProcessors/DetectAssetChangeProcessor.cs:308-324`](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/Editor/AssetProcessors/DetectAssetChangeProcessor.cs#L308-L324))
carries this comment:

> The asset-change watcher is an editor-authoring concern, not a play-mode one. `EnsureInitialized()`
> below runs `BuildWatchers`, an all-types/all-methods reflection scan; if a play-mode test mutates the
> AssetDatabase, Unity invokes this callback recursively inside the import phase and that scan
> destabilizes the asset pipeline (**a native mono crash on some Unity versions**, multi-minute importer
> stalls on others), aborting the play-mode test run with a zero-count results.xml. **Skip while
> playing.**

and then:

```csharp
if (EditorApplication.isPlayingOrWillChangePlaymode)
{
    return;
}

EnsureInitialized();
```

A `-batchmode -nographics -quit -executeMethod ...` CI run is **not** play mode, so the guard passes and
the scan runs. The crash the comment describes then happens through the other door.

The comment's own justification — *"an editor-authoring concern"* — applies at least as strongly to
batchmode, where there is no author at all and nothing can act on a file-watcher callback.

### Steps to Reproduce

1. Add `com.wallstop-studios.unity-helpers` 3.5.1 to a project on Unity `6000.5.2f1`.
2. Run the editor headless against that project, e.g. the shape game-ci uses:

   ```
   Unity.exe -batchmode -nographics -quit -projectPath . \
             -executeMethod <YourCompileCheck>.RunFromCI \
             -disable-assembly-updater -logFile -
   ```
3. Let it get far enough to compile the project's assemblies and trigger an asset postprocess pass.
4. Unity compiles everything, then dies before `-executeMethod` can finish.

[...]

Issue #318, opened 2026-07-29
SingleThreadedThreadPool.Dispose() drops queued work: it cancels the worker before awaiting it (measured 197/200)


wallstop:
## Summary

`SingleThreadedThreadPool.Dispose()` **cancels its worker rather than draining its queue**, so work that
was enqueued but not yet started is silently dropped. Because `DisposeAsync` *awaits* `_workerTask`, the
shape reads as "wait for outstanding work to finish" — and it does wait, but only for the item already
in flight.

This bit two of the three consumers in [DoxReloaded](https://github.com/Ambiguous-Interactive/DoxReloaded),
where the pool carries persistence writes: a settings change made immediately before teardown was dropped.

## Why the queue is not drained

```csharp
public async ValueTask DisposeAsync()
{
    _active = false;
    _disposed = true;
    _cancellationTokenSource.Cancel();   // <-- before the await
    Signal();
    try { await _workerTask.ConfigureAwait(false); }
    ...
}

private async Task DoWorkAsync()
{
    while (_active && !cancellationToken.IsCancellationRequested)   // <-- both already false
    {
        if (_work.TryDequeue(out WorkItem workItem)) { ... }
        ...
    }
}
```

By the time the worker re-evaluates its loop condition, `_active` is `false` **and** the token is
cancelled, so it exits with items still in `_work`.

## Measured

Against this file's own source, on .NET 9, 200 trials each:

| scenario | result |
| --- | --- |
| enqueue 25 items, `Dispose()` immediately | lost work in **200/200** trials, mean **20.9 of 25** dropped |
| enqueue 1 item, `Dispose()` immediately | dropped in **197/200** trials (**98.5 %**) |

And the window is narrow, which is what makes it hard to notice — the same probe with a delay between the
enqueue and the `Dispose`:

| gap | item dropped |
| ---: | ---: |
| 0 ms | 297/300 (99.0 %) |
| 1 ms | **0/300** |
| 2 ms+ | **0/300** |

So a consumer usually wins the race and the drop stays invisible until the one time it matters.

## Two suggestions, either of which would have prevented it

1. **Offer a drain.** A `CompleteAsync()` / `DrainAsync()` that stops accepting new work, lets the queue
   run down, and *then* cancels. Consumers that enqueue durable work (persistence) need this; consumers
   that enqueue recomputable work (we have one that enqueues level generation) correctly want today's
   cancel-on-dispose behaviour, so both should be expressible.
[...]

Issue #282, opened 2026-07-11
[Bug]: NativePCGRandom has code rot


wallstop:
### Description

NativePCGRandom has not kept up with correctness/performance changes in the AbstractRandom base and has problems such as NextBool bias, and NextFloat/NextDouble potentially returning 1.0 exactly, among others. We should bring it up to the latest and greatest speed.

### Steps to Reproduce

Look at NativePCGRandom code and compare to PCGRandom

### Expected Behavior

Equivalent, except for NativePCGRandom being a value type

### Actual Behavior

Different behavior

### Severity

Major (feature broken, no workaround)

### Unity Version

Other

### Package Version

3.5.0

### Platform

- [ ] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Regression

- [ ] Yes, this is a regression (worked in a previous version)

wallstop:
Fixed on `main`; closing with the list, because "code rot" turned out to cover six separate defects and it is worth recording which is which.

| Defect | Fix |
| --- | --- |
| Increment was not forced odd, so most seeds ran a shortened-period LCG | `NormalizeIncrement`, applied in both constructors |
| `NextFloat()` could return exactly `1.0f` | scale is `1f / (1 << 24)` |
| `NextLong()` could return a negative | masks the sign bit |
| `NextUint(0)` divided by zero | throws `ArgumentException`; `Next` guards `<= 0` |
| Bounded draws used modulo, so low bounds were biased | Lemire, with a power-of-two fast path |
| `NextBool()` spent a full step per flip and was slightly unfair | buffers bits |
| No `NextDouble()` | added |

Two of the tests are worth naming, because the others are the ones you would expect:

- The increment test asserts `_increment & 1 == 1` **directly**. It was originally written as a chi-square proxy, which a mutation walked straight through — a statistical test for a structural property is not a test of that property.
- The `NextFloat()` bound is checked **exhaustively over all 2^24 mantissas**, multiplied by the production constant rather than a re-derived copy, so the test cannot pass by making the same mistake twice.

Two things deliberately not done, both compat calls: the type still does not implement `IRandom` and still has no `RandomState`. Say so if either was part of what you meant by rot and I will open a follow-up.

**Note for anyone depending on this:** sequences for a given seed have changed. That is unavoidable — the old increment was the bug — but it means saved seeds do not reproduce their old streams.

[...]

Issue #275, opened 2026-07-05
Unity 6000.5 compile errors: 3.4.0 references Sirenix/Odin types without dependency


wallstop:
## Description

`com.wallstop-studios.unity-helpers@3.4.0` fails to compile in a Unity `6000.5.2f1` project because runtime sources reference Sirenix/Odin types that are not present in the project and are not resolved by the package manifest.

This was observed after resolving the package from the npm scoped registry, not from an embedded package folder.

## Environment

- Unity Editor: `6000.5.2f1`
- Package: `com.wallstop-studios.unity-helpers@3.4.0`
- Resolver/source: UPM registry, `https://registry.npmjs.org`
- Platform observed: Windows editor through Unity MCP

Manifest snippet from the consuming project:

```json
{
  "dependencies": {
    "com.unity.ai.assistant": "2.13.0-pre.2",
    "com.wallstop-studios.data-visualizer": "0.0.35-rc05.6",
    "com.wallstop-studios.dxmessaging": "3.2.0",
    "com.wallstop-studios.dxstate": "0.0.2",
    "com.wallstop-studios.unity-helpers": "3.4.0"
  },
  "scopedRegistries": [
    {
      "name": "NPM",
      "url": "https://registry.npmjs.org",
      "scopes": ["com.wallstop-studios"]
    }
  ]
}
```

Unity MCP package-state probe confirms registry resolution:

```text
UnityVersion=6000.5.2f1
com.wallstop-studios.unity-helpers version=3.4.0 source=Registry resolvedPath=D:\Code\DoxReloaded\Library\PackageCache\com.wallstop-studios.unity-helpers@ed8fe2701d2f
EmbeddedWallstopFolders=<none>
```

## Compiler log excerpt

```text
Library\PackageCache\com.wallstop-studios.unity-helpers@ed8fe2701d2f\Runtime\Tags\AttributeEffect.cs(17,11): error CS0246: The type or namespace name 'Sirenix' could not be found (are you missing a using directive or an assembly reference?)
Library\PackageCache\com.wallstop-studios.unity-helpers@ed8fe2701d2f\Runtime\Tags\AttributeEffect.cs(111,9): error CS0246: The type or namespace name 'SerializedScriptableObject' could not be found (are you missing a using directive or an assembly reference?)
[...]

PR #344, opened 2026-08-05
Ship the nested-collection remedy, fix signed-enum drawers, recover ghost-cancelled Unity runs

### LINKED ISSUE #314: SerializableDictionary<TKey, List<TValue>> silently serializes keys and drops every value (nested collection); the documented cache workaround has no property drawer
**Package**: `com.wallstop-studios.unity-helpers` **3.5.1** (PackageCache `ec77ef9677b2`)
**Unity**: 6000.5.2f1 (also present since at least 3.x — see "how long this survived")

## Summary

A `SerializableDictionary<TKey, TValue>` whose `TValue` is itself a collection is **write-only under
Unity serialization**: the asset records `_keys` and no `_values` at all. There is no error, no
warning, and the Inspector keeps accepting entries, so the dictionary looks authored and every
runtime lookup comes back empty.

Unity's serializer refuses a nested collection — a field typed `List<T>[]`, `List<List<T>>` or
`T[][]` is not serialized. `SerializableDictionaryBase<TKey, TValue, TValueCache>` declares

```csharp
[SerializeField]
protected internal TValueCache[] _values;
```

and the two-argument convenience type sets `TValueCache = TValue`:

```csharp
public class SerializableDictionary<TKey, TValue> : SerializableDictionaryBase<TKey, TValue, TValue>
```

so `SerializableDictionary<string, List<float>>` makes that field a `List<float>[]` and Unity drops
it. `_keys` is a `TKey[]`, so it survives — which is what makes the failure look like data rather
than a type error.

## Evidence (measured, on real assets)

Swept every `_keys:` block in a production project's `Assets/**/*.{asset,prefab,unity}`:

```text
total _keys blocks:  53
missing _values:      7   <- all seven are the List<>-valued shape
                          the other 46 (simple value types) round-trip fine
```

The decisive contrast is **one file, one serializer, two dictionaries**:

| Field | Declared type | `_keys` | `_values` |
| --- | --- | --- | --- |
| `_waitTimesStore` | `SerializableDictionary<Vector2, float>` | present | **present** |
| `_eventsStore` | `SerializableDictionary<Vector2, List<KEventData>>` | present | **absent** |

The only difference is the value type being a `List<>`.

Note the shape of the absence: a **correctly** typed empty dictionary writes both halves —

```yaml
  _waitTimesStore:
    _keys: []
    _values: []      # correct, empty
  _eventsStore:
    _keys: []        # mis-shaped: no _values line at all
```

so "no `_values` sibling" is a reliable detector of the declaration being wrong, independent of
whether anything has been authored yet.

**The positive direction is measured too.** Before changing the value type,
`new SerializedObject(component).FindProperty("_decals._values")` returned **null** — Unity has no
such field. After wrapping the list in a `[Serializable]` class it returned non-null and the YAML
became:

```yaml
  _decals:
    _keys:
    - {fileID: 11400000, guid: 8938d774870e045479e820d3dbfded72, type: 2}
    _values:
    - _items:
      - {fileID: 21300000, guid: 46a73363348532645be9383d370c420b, type: 3}
```

Also worth knowing: **this is Unity-specific**. `_keys`/`_values` carry `[ProtoMember]` and
`[JsonInclude]`, so the same dictionary can round-trip through the JSON/ProtoBuf paths and still lose
every value in a `.prefab` or `.asset`. That asymmetry makes it materially harder to notice.

## Why nothing notices — the drawer swallows it

`SerializableDictionaryPropertyDrawer` resolves both halves and tolerates a missing one:

```csharp
// GetOrCreateCachedPropertyPair
valuesProperty = dictionaryProperty.FindPropertyRelative(
    SerializableDictionarySerializedPropertyNames.Values);   // -> null for a nested value type

// EnsureParallelArraySizes
if (keysProperty == null || valuesProperty == null)
{
    return;                                                  // silently
}
```

and then continues drawing. So the user gets a working-looking key column and a value column that
persists nothing. The drawer is the one component that *can* detect this — `SerializedProperty` is
exactly the thing that knows the field was not serialized — and it is the one place currently
choosing not to.

## Second, separable bug: the documented workaround has no drawer

`SerializableDictionary<TKey, TValue, TValueCache>` is documented as the answer to this class of
problem:

> Serializable dictionary that stores value data inside cache objects so complex or
> **non-serializable** runtime types can participate in Unity serialization.

And it does work — `[Serializable] sealed class FloatListCache : SerializableDictionary.Cache<List<float>> {}`
makes `_values` a `FloatListCache[]`, which Unity serializes. But the drawer is not registered for it:

```csharp
[CustomPropertyDrawer(typeof(SerializableDictionary<,>), true)]
[CustomPropertyDrawer(typeof(SerializableSortedDictionary<,>), true)]
[CustomPropertyDrawer(typeof(SerializableSortedDictionary<,,>), true)]   // <- sorted cache form: yes
//         SerializableDictionary<,,>                                    <- unsorted cache form: missing
```

`useForChildren: true` does not cover it, because the three-argument type derives from
`SerializableDictionaryBase<TKey, TValue, TValueCache>` and **not** from `SerializableDictionary<TKey, TValue>`.
The sorted variant is registered, so this reads as an oversight rather than a decision. Net effect:
following the documentation to fix the data costs you the Inspector.

## Scope — the same shape, three more types

| Type | Backing field |
| --- | --- |
| `SerializableDictionaryBase<TKey, TValue, TValueCache>` | `TValueCache[] _values` |
| `SerializableSortedDictionaryBase<TKey, TValue, TValueCache>` | `TValueCache[] _values` |
| `SerializableSetBase<T, TSet>` (`SerializableHashSet<T>`, `SerializableSortedSet<T>`) | `T[] _items` |

A set of lists is a rarer thing to author than a dictionary of lists, but the mechanism is identical,
so a fix aimed at the whole family is worth more than one aimed at `SerializableDictionary`.

## Repro

```csharp
using System.Collections.Generic;
using UnityEngine;
using WallstopStudios.UnityHelpers.Core.DataStructure.Adapters;

public sealed class NestedValueRepro : MonoBehaviour
{
    [SerializeField]
    private SerializableDictionary<string, float> _control = new();          // simple value

    [SerializeField]
    private SerializableDictionary<string, List<float>> _subject = new();    // collection value
}
```

1. Put it on a prefab.
2. In the Inspector, add one entry to **each** dictionary and give both a value.
3. Save, then open the `.prefab` in a text editor.

`_control` has `_keys` and `_values`. `_subject` has `_keys` and **no `_values` key at all**.

4. Enter Play mode (or reimport) and read both back: `_control` has its entry, `_subject` is empty.
5. Optional, and the cleanest single check:
   `new SerializedObject(component).FindProperty("_subject._values")` is `null`, while
   `"_control._values"` is not.

*Honesty note*: steps 1–5 are the reduction of what was measured on real project assets (the sweep
and the before/after `FindProperty` results above). The standalone script itself has not been run —
the editor that produced the measurements became unavailable — so if you would like it confirmed in
this exact form before acting, that is a fair ask.

## Suggested fix

Ordered by value-per-risk. (1) alone would have saved the whole investigation.

**1. Make it loud in the Inspector — smallest change, biggest win.**
In `SerializableDictionaryPropertyDrawer`, treat "keys resolved, values did not" as a first-class
error rather than a shrug. In both `OnGUI` and `GetPropertyHeight`, right after the pair is resolved:

```csharp
if (keysProperty != null && valuesProperty == null)
{
    EditorGUI.HelpBox(
        position,
        $"'{property.displayName}' cannot be serialized: Unity does not serialize a nested "
            + "collection, so this dictionary's values are dropped and every lookup returns "
            + "nothing. Wrap the value type in a [Serializable] class.",
        MessageType.Error);
    return;   // and return the box's height from GetPropertyHeight
}
```

The drawer already has an `Unsupported type (X)` vocabulary (`GetUnsupportedTypeContent`), so this
fits what is there. Refusing to draw the rows is deliberate: letting someone keep typing into a
column that persists nothing is the actual harm.

**2. Register the missing drawer** — one attribute, restores the documented escape hatch:

```csharp
[CustomPropertyDrawer(typeof(SerializableDictionary<,,>), true)]
```

**3. Ship the wrapper, so the fix is a type change and not a new class per element type.**
Today the cache route needs a concrete `Cache<T>` subclass for every value type. A single
general-purpose box removes that:

```csharp
[Serializable]
public sealed class SerializableList<T>
{
    [SerializeField]
    private List<T> _items = new();

    public IReadOnlyList<T> Items => _items ?? (IReadOnlyList<T>)Array.Empty<T>();
    public int Count => _items == null ? 0 : _items.Count;

    public SerializableList() { }

    public SerializableList(IEnumerable<T> items) =>
        _items = items == null ? new List<T>() : new List<T>(items);
}
```

`SerializableDictionary<Prop, SerializableList<Sprite>>` then serializes, and keeps the existing
two-argument drawer. **Measured**: the generic instantiation serializes fine — no per-element-type
subclass needed — which was the one thing worth checking rather than assuming. (This is what we
shipped downstream; the library is the better home for it, since the constraint is the library's.)

**4. Say it in the XML docs.** `SerializableDictionary<TKey, TValue>`'s summary should state that
`TValue` must be Unity-serializable and that a collection is not, with a pointer to (2)/(3).

**5. Stretch: catch it at compile time.** A C# generic constraint cannot express "not a collection",
but a small Roslyn analyzer over `SerializableDictionary<,>` / `SerializableSet*<>` type arguments
could, and would move the failure from asset-authoring time to build time.

## How long this survived, for calibration

The project this was found in previously used Odin, which **could** serialize the nesting. The assets
still carry the correct mapping in their now-dead `serializationData` blobs, so migrating off Odin
silently changed what the data model could express — four separate declarations went inert at once
and one of them (a prop → decal-sprite mapping) was dead for months with the only symptom being a
visual effect that never appeared. The consumer guarded with `values.Count <= 0`, which is exactly
the defensive check that turns this into silence.

Filed from downstream work: Ambiguous-Interactive/IshoBoy#263 (analysis) and
Ambiguous-Interactive/IshoBoy#265 (the downstream fix, including the wrapper in (3) and a build gate
asserting no serialized dictionary has keys without values).

### LINKED ISSUE #333: [Bug]: MCP bridge resolves wrong repo
### Description

We need to migrate to the MCP like https://github.com/Ambiguous-Interactive/DxMessaging now uses - an npm command, with an optional config. Right now the MCP resolves the wrong repo root and is super confused with custom powershell scripts.

### Steps to Reproduce

Do MCP stuff locally

### Expected Behavior

It works!

### Actual Behavior

Confusion

### Severity

Critical (crash, data loss)

### Unity Version

Unity 6.2 (6000.2)

### Package Version

3.5.1

### Platform

- [x] Editor
- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Regression

### Additional Context

_No response_

### LINKED ISSUE #339: [Bug]: Custom Enum Operations Do Not Work On Signed Enums
### Description

I think our enum comparisons and "has flags no alloc" and similar are broken for enums that implement weird things like "sbyte" and have explicit negative ordinals.

I think we can easily fix this by doing a Typof(TEnum).TypeKind or similar (or finding its underlying type in a reliable way), switching on all possible numeric values, and just binding explicit comparisons and the like

### Steps to Reproduce

Have signed enum values with negative ordinals

### Expected Behavior

All Unity Helpers custom enum operations, hash codes, equality, Objects.<whatever> to work flawlessly

### Actual Behavior

Dragons!

### Severity

Critical (crash, data loss)

### Unity Version

Unity 6.2 (6000.2)

### Package Version

3.5.1

### Platform

- [ ] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Regression

### Additional Context

_No response_

### LINKED ISSUE #342: ci: Unity matrix job reported cancelled with every step successful; re-run went green
## Summary

Unity Tests run [`31030368120`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31030368120) (attempt 1, PR #341, head `c9b539bb`) reported `cancelled` for `Resolve Unity test matrix` **even though every one of that job's eight steps succeeded, including `Complete job`.** The three licensed tiers and the smoke job then skipped on the unmet `needs:`, and `Unity CI Success` reported failure.

A plain re-run of the same run id, with no code change, went green: 18/18 Unity jobs.

This is filed to **count occurrences**, not to fix anything yet. One instance is not a pattern, and the gate behaved correctly given its input.

## Evidence

Attempt 1 job conclusions:

```text
cancelled  Resolve Unity test matrix         (17:31:01 -> 17:33:04)
success    Self-hosted runner access preflight (17:31:00 -> 17:31:04)
skipped    Unity <version> <mode>
skipped    Unity standalone (gated)
skipped    Unity <version> <mode> (SINGLE_THREADED)
skipped    Unity package export smoke
failure    Unity CI Success                  (17:33:07 -> 17:33:12)
```

The cancelled job's own steps:

```text
success  Set up job
success  Checkout
success  Detect superseded pull request head
success  Validate Unity test-project module manifest
success  Resolve dispatch overrides
success  Check for required licensed workflow secrets
success  Post Checkout
success  Complete job
```

Its log confirms the work finished — all six outputs were set (`unity-versions`, `test-modes`, `matrix-include`, `matrix-include-standalone`, `has-required-secrets`, `superseded`) followed by `Cleaning up orphan processes`.

Attempt 2 (`rerun`, no changes): `success`, started 17:35:19, finished 18:26:10.

## What was ruled out

| Candidate | Ruled out by |
| --- | --- |
| A newer run cancelling it via concurrency | None existed. `c9b539bb` was and remained the PR head; only one Unity Tests run was created in the window. |
| The concurrency group generally | `cancel-in-progress: false`, and the previous occupant (run `31025096491`, head `4002db7f`) finished at **17:25:43** — five minutes before this run was created at 17:30:51. The group was empty. |
| A group-name collision with another workflow | `unity-tests.yml` is the only workflow using `unity-tests-<pr>`; `unity-file-naming-lint.yml` uses `${{ github.workflow }}-…`, which cannot collide. |
| The stuck-job watchdog | Its cycles bracket the event without touching it: `17:23:27 -> 17:23:41` and `17:59:28 -> 17:59:40`. The cancellation was at 17:33:04. |

For contrast, the other cancellation on that branch that day — run `31024804075`, head `c2556596` — **is** explainable and is not part of this report: it was cancelled at `16:24:13`, one second after run `31025096491` entered the group at `16:24:12`. That is GitHub's documented behavior for `cancel-in-progress: false` (one run may be in progress and one pending; a third arrival cancels the pending one). Worth knowing, because it means rapid pushes routinely discard a queued Unity run by design.

## Why this is not obviously fixable in this repo

`Unity CI Success` did the right thing. Supersession deliberately does **not** waive the hosted gates (see the comment in that job: `matrix-config` goes on to lint the test-project module manifest, and a failed job still publishes its outputs, so waiving it would report green for a bogus UPM module id or an offline runner fleet). A run whose matrix job did not succeed has produced no verdict, and red is the honest report.

Making the gate treat `cancelled` as "no verdict → pass" would be worse: a deliberately cancelled run would then satisfy a required check.

So the defect looks upstream — GitHub marking a job whose every step succeeded as `cancelled`.

## Suggested handling

1. **Count it.** If this recurs, link the run here. The signature is specific and cheap to recognise: a job with `conclusion == "cancelled"` whose steps are all `success`.
2. **Only if it recurs**, consider teaching `stuck-job-watchdog.yml` to detect that signature and re-run the failed jobs automatically — the remedy is entirely mechanical (`POST /actions/runs/{id}/rerun`), and the watchdog already holds the App credentials and already audits runs on a schedule.
3. Do **not** relax `Unity CI Success` for cancelled hosted gates.

## Cost while it stands

A re-run. The whole Unity matrix is ~50 min of wall clock and it occupies the single organization Unity seat for that time, so it is not free — but it is not a correctness risk either, and it is self-announcing (a red required check that goes green on re-run with no diff).

Context: #332 (superseded runs), #326 (dispatch transitions), and the dispatcher-stuck behaviour the watchdog exists for (community discussion 186811) — this may or may not be the same family.

wallstop:
Three things, each closing out something a previous session left half-done.

## 1. `[WEnumToggleButtons]` threw on any enum with a negative member (#339)

Session 163 fixed the runtime half of #339 and the CHANGELOG entry is accurate about what it was: enums with negative members were **slow** to name. That is a performance fix.

The editor drawers were never touched, and they were the crash:

- `WEnumToggleButtonsDrawer.ConvertToUInt64` is called from `BuildEnumOptions`' plain `for` loop with **no handler**. `Convert.ToUInt64` throws `OverflowException` on every member below zero, so one negative member took the whole Inspector down.
- `EnumToggleButtonsShared.ConvertToUInt64` makes the same call inside `catch { return 0UL; }`. Worse in its own way — the member silently became mask `0`, so the Odin drawer rendered it as "None" and wrote that back.

`enum Facing : sbyte { Left = -1, ... }` with that attribute throws on `main` today. The issue's severity field says Critical and its ask is "**all** Unity Helpers custom enum operations", so this was in scope and open.

**The fix is one method, not three.** Neither `Convert.ToUInt64` nor `Convert.ToInt64` is total over enums — the first throws on negatives, the second on `ulong`-backed members above `long.MaxValue`. Dispatching on the underlying `TypeCode` is the only conversion that covers all nine shapes, and it now lives once in `EnumExtensions`, beside the generic `EnumNumericHelper<T>` path it mirrors. `ApplyEnumOption`'s `Convert.ToInt64` was the mirror defect and got the same treatment.

**Swept, and deliberately not changed:** `DropDownShared` and `ShowIfConditionEvaluator` also convert boxed enums, but with `Convert.ToInt64`, which handles negatives correctly. They only break on the exotic `ulong`-above-`long.MaxValue` case, where they already degrade to a `false` comparison inside an existing `try`. Two duplicate private helpers would cost more than that buys.

## 2. `SerializableList<T>` — the remedy half of #314

#328 shipped the *detection*: the drawer reports "keys resolved, values did not" instead of drawing a column that persists nothing. What it did not ship is a remedy a consumer can reach for cheaply — the documented cache form needs a concrete `Cache<T>` subclass **per value type**.

`SerializableList<T>` is the layer of indirection Unity wants. `SerializableDictionary<TKey, SerializableList<TValue>>` serializes; `SerializableDictionary<TKey, List<TValue>>` does not. The consumer's fix becomes a type change and nothing else.

It implements `IList<T>`/`IReadOnlyList<T>`, converts implicitly both ways with `List<T>`, exposes `AsList()` for `Sort`/`BinarySearch`, draws as the wrapped list with no extra foldout, and carries a `JsonConverter` so the JSON shape stays a plain array. The inspector error now names it ahead of the cache form.

Out-of-range handling is split on purpose: every member that can express "did nothing" does (`RemoveAt`/`CopyTo` no-op, `Insert` clamps, null sequences ignored), and the indexer — which must produce a `T` and so has no way to signal failure — throws exactly as `List<T>` does. Returning `default` for a bad index would hand the caller a value that looks authored.

Coverage went into the existing `NestedCollectionSerializationTests` rather than a new fixture: that file already holds the control and dropped shapes inside one `SerializedObject`, so the wrapper is a third column in the same comparison.

## 3. The watchdog now recovers a run cancelled with every step green (#342)

#342 asked to *count* occurrences and automate only on recurrence. Implemented now instead, because the detection is exact and the remedy is one API call — and the alternative is a human noticing a red required check with nothing wrong with it.

Signature: a job with `conclusion: cancelled` whose step list is **non-empty** and whose every step is `success`. The non-empty requirement is what makes it safe — a deliberate cancel always leaves an in-flight step that is itself cancelled, so it can never match. Recovery is `POST actions/runs/{id}/rerun` on `GITHUB_TOKEN`'s existing `actions: write`; the read-only build-lock App is not used and must not be widened for it.

**Two placement details that would otherwise have been bugs:**

- The phase runs **before** the queued-run scan. That scan calls `flush_summary_and_exit` the moment the queue is clean, which is the normal state — a phase appended after it would have been dead code that read correctly in review.
- The counter lives in `rerun-<id>.json`, not the cancel state file, whose reader is `jq -r '.cancels // .reruns // 0'`. Writing a `reruns` key into that file would have been read back as a *cancel* count and silently spent the cancel budget.

Bounded twice — `run_attempt <= MAX_RERUN_ATTEMPT` (a re-run landing in the same state escalates to a human instead of looping every five minutes) and `MAX_RERUNS_PER_DAY` per run id. The window is one hour because each extra hour costs a jobs API call per cancelled run, and rapid pushes leave queued runs cancelled by design.

## How this was verified

**The Unity MCP compile gate was unavailable.** It resolved `D:/Code/IshoBoy`, not the development project, on two consecutive calls. `Packages/com.wallstop-studios.unity-helpers` existed there — which is the trap — but with no `.git`, no file created seconds earlier, and 3 package assemblies instead of ~39. That is a consumer copy. Data and three discriminating assertions added to #333.

So the behavioral signal came from a `dotnet` harness compiling the **real** sources and the **real** NUnit tests against stubs for the few Unity types: **81 passed, 0 failed.**

**Red proven by mutating the fix, not by asserting it:**

| Mutation | Result |
| --- | --- |
| `Convert.ToUInt64` alone (the drawers' old behavior) | 7 signed shapes fail |
| `Convert.ToInt64` alone (the mirror error) | `UnsignedLongEnum` fails |
| the shipped dispatch | 81/81 |

The boxed-vs-generic agreement assertion went **inside the existing 16-shape `TestCaseSource`**, so a new enum shape costs one source line and picks up the conversion automatically.

The harness also caught a defect the compiler could not: `[JsonConverter(typeof(Factory))]` requires a **public** parameterless constructor, and the factory had copied the private-singleton pattern from `CyclicBufferConverterFactory`, which is only ever reached through `Instance`. That was a runtime `InvalidOperationException` on first serialize.

The watchdog change was checked with `bash -n` and `shellcheck` (clean) on the extracted `run:` body, `actionlint`, `yamllint`, and four new contract assertions — **each proven to fail** by deleting the guard it pins, then restored. `npm run validate:prepush` passes.

**Not verified locally:** the NUnit suites inside Unity, and everything touching `SerializedObject`. CI is the gate for those.

### One risk taken off the table deliberately

The drawer test initially covered a `ulong`-backed enum with members above `long.MaxValue`. Unity's handling of 64-bit-backed enum *fields* is not something this environment can verify, and betting a CI leg on it would have been guessing. That extreme is covered in the pure-C# runtime tests instead (`UnsignedLongEnum.VeryLarge = 1UL << 63`); the drawer test covers two signed widths Unity certainly serializes.

## Also in this session, not in this diff

#282, #318, and #327 were verified fixed on `main` by earlier sessions and closed with evidence. #339 and #314 close when this merges, since their remaining halves ship here.

Closes #314
Closes #339
Closes #342

---

> [!NOTE]
> **Medium Risk**
> Touches editor serialization drawers, protobuf/JSON paths, and automated GitHub Actions cancel/rerun behavior; MCP and workflow changes affect CI recovery and local agent setup rather than runtime game code.
> 
> **Overview**
> Adds **`SerializableList<T>`** as the recommended fix for nested collections in serialized dictionaries/sets (inspector drawer, JSON as a plain array, protobuf empty-list handling), and updates dropped-value diagnostics and docs to point at it ahead of cache subclasses.
> 
> **Enum/editor fixes (#339):** New **`TryConvertToUInt64` / `TryConvertToInt64`** (generic + boxed) replace `Convert.To*` in **`[WEnumToggleButtons]`** drawers and shared flag logic (underlying-type mask for signed top-bit flags). **`EditorCacheHelper.GetEnumDisplayName`** uses a per-type value→name cache instead of **`Enum.GetValues`** on every call.
> 
> **CI watchdog (#342):** Before the queued-run scan, detects completed runs **cancelled with all steps successful**, **`POST` reruns** them with caps (`MAX_RERUNS_PER_DAY`, attempt/age limits), and ages stuck **queued** runs on **`run_started_at`** so fresh reruns are not auto-cancelled.
> 
> **Unity MCP (#333):** Replaces bash/PowerShell helpers with **`scripts/mcp/unity-mcp.mjs`** and **`npm run unity:mcp:{bridge,configure,probe}`** (bearer auth, default port **9007**, **`GetProjectRoot`** pinning in **`.env.local`**). **`.env.local`** is gitignored; MCP docs/skills updated.
> 
> **Tests:** Enum/list coverage, **`LogAssert.ignoreFailingMessages`** in flaky singleton tests plus unconditional teardown reset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af175696e603f900824a11dd9f97ace78de5aeb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

# Added after the original description

## 4. The MCP bridge resolved the wrong Unity project (#333)

An MCP endpoint is a **host and port**; a bridge is bound to **one editor**, and that editor is whichever one claimed the port. Nothing ever asked which project was on the other end, so agents compile-gated a completely unrelated project while every check reported success. This is not hypothetical — it is why two compile errors in this very branch reached CI instead of being caught in seconds.

The identity was one POST away the whole time. Against the endpoint `.mcp.json` actually named:

```json
{ "success": true, "data": { "projectRoot": "D:/Code/IshoBoy" } }
```

Three things made it silent: the default endpoint was **one developer's LAN address hard-coded in two shell scripts**; `.env.local` (their documented source) was **not gitignored** although all four generated configs were; and the obvious check does not discriminate, because a consumer project contains `Packages/com.wallstop-studios.unity-helpers` too. Port `9003` was also DxMessaging's, so the two repos were competing for one endpoint by construction.

**Fix:** ported the sibling projects' Node entry point — `npm run unity:mcp:{bridge,configure,probe}`, optional `.env.local`, bearer auth, transactional writes — and retired all three bash/PowerShell scripts. Then added the layer **none of DxMessaging, IshoBoy, DoxReloaded or qora-redux has**: `probe` and `configure` call `GetProjectRoot` after the handshake. `probe` always prints the project; `configure` pins it into `.env.local` and refuses to write on a mismatch.

**Two traps in the reference implementation**, both of which would have made things worse: its default port was **9020 (qora-redux)** with a **9003 (DxMessaging)** fallback, and its server key was `unity-mcp` rather than `unity-mcp-remote` — which would have **renamed every `Unity_*` tool** an agent sees. This repo takes **9007**, keeps `unity-mcp-remote`, and a new `UNH-MCP-PORT` check pins all of it (verified to fail on a mutation adopting a sibling's port).

## 5. Reviewer feedback

Cursor Bugbot found **six real defects** across two passes; all six are fixed with a test each and every thread is resolved. Copilot returned *"unable to review — quota limit"* on all four attempts.

| Finding | Severity |
| --- | --- |
| Watchdog could cancel the run it had just re-run (`created_at` is not reset by a re-run) | Medium |
| Empty `SerializableList<T>` could not survive protobuf | Medium |
| Signed top-bit flag lost its button (`sbyte -128` sign-extends) | Medium |
| `--project-root` / `--any-project` rejected by `parseArgs` | Medium |
| Probe skipped `notifications/initialized` | **High** |
| Codex table key drift (`unity-mcp-remote` vs `unity_mcp_remote`) | **High** |

The two High ones would have failed **only against correct implementations** — a conforming MCP relay, and an existing Codex config — which is exactly the class local testing against one lenient server never finds.

## 6. Allocation work, from review

`this Enum value` boxes. Measured over 10,000 calls with `GC.GetAllocatedBytesForCurrentThread`:

| call shape | before | after |
| --- | ---: | ---: |
| `Direction.Left.TryConvertToUInt64(out _)` | 48 B | **0 B** |
| `((Enum)boxed).TryConvertToUInt64(out _)` | 24 B | 24 B |

Overload resolution picks the new generic overload with **no call-site change**. The boxed one stays because the drawers hold values as `object`; at those sites the box already exists, which is why that row is 24 and not 48 — the residual is `IConvertible` *inside* the method, not the signature.

The same sweep found `EditorCacheHelper.GetEnumDisplayName` rebuilding `Enum.GetValues` per call:

| enum size | before | after |
| --- | ---: | ---: |
| 4 members | 40 B/call | 24 B/call |
| 30 members | 144 B/call | 24 B/call |

Old tracks `24 + 4N` — the array itself, so one label cost more as the enum grew. New is constant. The method had **no test coverage at all**; it does now, including negative members of a signed enum (the case that breaks if the key is computed naively).

## 7. One Unity leg was red, and it was not a flake to re-run

`Unity 6000.5.2f1 editmode` failed on `HandlesCreateAssetExceptionWithPartialFile` — 7 of 8 legs passed identical code, 4182 of 4183 tests passed on the failing one.

`TestRoot` is `Assets/Resources/CreatorTests`, **inside the directory the singleton creator scans**, and two tests deliberately write invalid asset YAML there and let frames pass. Any refresh landing in that window logs a load error — including the deferred one `EnsureSingletonAssets` queues on `EditorApplication.delayCall` (`ScriptableObjectSingletonCreator.cs:436`), which fires on a later frame and is attributed to whichever test is running then.

The fixture **already had the guard on its four other garbage-writing tests**, with a comment describing this exact mechanism. The two later tests never got it. Both have it now, plus a backstop: `CommonTestBase.TearDown` clears `LogAssert.ignoreFailingMessages` unconditionally, since Unity does not reset it between tests and a leaked suppression silently swallows real errors.

Closes #333

## CI status at time of writing

**GitHub Actions is not scheduling.** The last two pushes produced zero workflow runs; a manual `workflow_dispatch` was accepted, sat `queued` 15 minutes on plain `ubuntu-latest`, then cancelled without executing, and cancelling it returned `HTTP 500`. Actions is enabled (`allowed_actions: all`). Evidence recorded on #342.

Note for anyone reading the red on `47f638e4`: those runs show `conclusion: failure` but every **job** under them is `cancelled` — supersession from rapid pushes, not breakage.

Local signal is green: `validate:prepush` exit 0, a dotnet harness compiling the real sources and NUnit tests 82/82, `test:unity-mcp` 12/12, `validate:mcp-config`, `test:validate-mcp-config`, and the watchdog contract tests.

**Not claimed:** the flake fix, the boxing overloads and the display-name cache have no completed Unity run behind them yet.


wallstop:
@cursor review

@copilot review

wallstop:
@cursor review

@copilot review

Since the last review this adds the #333 fix: the MCP tooling now asks the bridge which Unity project it has open instead of trusting a host:port. Worth a look at `scripts/mcp/unity-mcp.mjs` — particularly `probeEndpoint`'s new classification and `pinProjectRoot`.

wallstop:
@cursor review

Three findings from the last pass, all real and all fixed in `8c52e9fd`:

- **`--project-root` / `--any-project` were rejected by `parseArgs`.** Documented in `--help`, consumed by `resolveOptions`, never added to `OPTION_NAMES`/`FLAG_NAMES`. Every check I ran used the environment variables, which is precisely why I missed it. Good catch.
- **The probe skipped `notifications/initialized`.** Unity's server tolerates it; a conforming relay need not — and a refused identity query reads as "would not identify itself", so the new verification would have failed hardest against the strictest servers.
- **Codex table key drift.** `unity-mcp-remote` vs the `unity_mcp_remote` the validator and every on-disk config use. Verified against the real `.codex/config.toml`: the merge now produces exactly one table on the new endpoint, where before it would have left the old enabled one beside it and failed `validate:mcp-config`.

One test per finding, each verified red with its fix reverted (3 failed / 12) and green with it (12/12).

wallstop:
@cursor review

`47f638e4` fixes the one red Unity leg. It was not a scheduling accident, so re-running would have been the wrong call.

`ScriptableObjectSingletonCreatorTests` uses `TestRoot = Assets/Resources/CreatorTests` — **inside the directory the singleton creator scans** — and two of its tests deliberately write invalid asset YAML there and then let frames pass. Any `AssetDatabase` refresh landing in that window logs `Unknown error occurred while loading`, including the deferred one `EnsureSingletonAssets` queues on `EditorApplication.delayCall`. The stack named it: `ScriptableObjectSingletonCreator.cs:436`. Because it fires on a later frame, NUnit attributes it to whichever test is running then.

Evidence it is a window and not a bad assertion: 7 of 8 Unity legs passed the identical commit, and 4182 of 4183 executed tests passed on the failing one.

The fixture **already had this guard on its four other garbage-writing tests**, and the comment on one of them describes this precise failure mode. The two later tests never got it. Both now suppress across the window — it cannot be `try`/`finally`, since it has to span `yield`s in a coroutine.

Also added the backstop that was missing: `CommonTestBase.TearDown` now clears `LogAssert.ignoreFailingMessages` unconditionally. Unity does not reset it between tests, so a test failing inside a suppression window leaks it into everything after and silently swallows real errors — strictly worse than the flake. That covers every fixture in the repo.

Runtime/Core/Extension/EnumExtensions.cs
```diff
@@ -463,6 +464,99 @@ public static IEnumerable<string> ToCachedNames<T>(this IEnumerable<T> enumerabl
...
+        /// </remarks>
+        /// <example>
+        /// <code><![CDATA[
+        /// if (Direction.Left.TryConvertToUInt64(out ulong pattern))
+        /// {
+        ///     Debug.Log(pattern);
+        /// }
+        /// ]]></code>
+        /// </example>
+        public static bool TryConvertToUInt64(this Enum value, out ulong result)
```
wallstop, line 564: Is it possible to do like "this T value" where T : Enum, to get no boxing? Here and other helpers, across the repo?

wallstop: Yes — and it is worth the full allocation at every statically-typed call site. Done in `ffe2cbff`.

Measured over 10,000 calls on net9.0 with `GC.GetAllocatedBytesForCurrentThread`, against the real source:

| call shape | before | after |
| --- | ---: | ---: |
| `Direction.Left.TryConvertToUInt64(out _)` | 48 B/call | **0 B/call** |
| `((Enum)boxed).TryConvertToUInt64(out _)` | 24 B/call | 24 B/call |

The generic overload delegates to the existing `EnumNumericHelper<T>`, which already reads the value through `Unsafe.As` with no boxing at all. **Overload resolution picks it with no call-site change**, so existing callers get it for free.

**Why the boxed overload stays.** It is not redundant — it is the one the editor drawers need. They get values from `Array.GetValue` and `ToggleOption.Value` as `object`, so the enum type is only known at runtime and there is no type argument to supply. `Enum` also fails the `unmanaged` constraint, so the generic overload is not even a candidate there.

**Why that row was 24 and not 48, which is the interesting part.** At those sites the box already exists before the call, so the receiver costs nothing extra. The 24 B is *inside* the method — `IConvertible.ToInt64/ToUInt64` boxing the underlying value — not on its signature. Removing it needs `Unsafe.Unbox`-style tricks that do not belong in a defensive public API, and that path runs once per enum type behind `WEnumToggleButtonsUtility`'s `EnumOptionsCache`, so it is a handful of allocations per domain reload. I left it.

The 16-shape contract test now asserts the public generic overload and the boxed one agree for every member of every shape, instead of comparing against the internal helper the generic one delegates to.

**On "other helpers, across the repo":** I swept for `Enum`-typed parameters. There is exactly one other, and it is a bigger fish than boxing:

```csharp
// Editor/Core/Helper/EditorCacheHelper.cs:238
public static string GetEnumDisplayName(Enum value)
{
    ...
    int index = Array.IndexOf(Enum.GetValues(enumType), value);
```

`Enum.GetValues` allocates a fresh array **on every call** — for an N-member enum that dwarfs the 24 B box — and `Array.IndexOf` then boxes during comparison. The display names beside it are already cached; the values are not. It is editor-only and pre-existing, so I did not fold it into this PR. Say the word and I will either fix it here or file it with these numbers.

Everything else enum-shaped in the repo already takes `T : unmanaged, Enum` (`ToCachedName`, `ToDisplayName`, `HasFlagNoAlloc`, `EnumNumericHelper<T>`), so there is no third instance.

wallstop: Folded in as `7ed84964`.

`GetEnumDisplayName` no longer rebuilds anything per call. Measured over 2,000 calls on net9.0, same method:

| enum size | before | after |
| --- | ---: | ---: |
| 4 members | 40 B/call | **24 B/call** |
| 30 members | 144 B/call | **24 B/call** |

The old figure tracks `24 + 4N` almost exactly — that is the `Enum.GetValues` array itself, so the cost of asking for one label grew with the enum. The new figure is constant; the residual 24 B is the `IConvertible` box inside the boxed overload and does not scale.

It is now a `Dictionary<ulong, string>` per enum type, built once from the **same names array** so display-name resolution keeps a single source of truth, and cleared alongside it in `ClearAllCaches` — letting those two diverge would be worse than the allocation.

Behavior is unchanged, and there was **no coverage for this method at all** before, so the new fixture pins it: `InspectorName` wins, undecorated members are nicified, aliases collapse to one stable declared name, negative members of a signed enum resolve (the key sign-extends — this is the one that would break if the bit pattern were computed naively), null returns empty, an undefined value falls back to `ToString`, and a cache clear does not strand the map.

One deliberate non-assertion: the alias test does **not** pin which of the two names wins. `Enum.GetValues` gives no ordering guarantee among members sharing a value, so pinning it would test the runtime rather than this method. Old and new read the same array in the same order, so whatever that order is, they agree.

Also added a CHANGELOG entry, since `EditorCacheHelper` is shipped public API.

wallstop:
Disclosure on one file in this PR, found while waiting out the Actions outage.

**`Editor/CustomDrawers/Odin/WEnumToggleButtonsOdinDrawer.cs` is read-verified only — it compiles nowhere.**

`WALLSTOP_UNITY_HELPERS_ODIN_INSPECTOR` comes from a `versionDefines` entry keyed on the `odininspector` package. CI's ephemeral test project is generated from `.github/unity-test-project-modules.json`, which declares built-in modules and `com.unity.ugui` only — `rg 'Sirenix|odininspector' scripts/unity/ .github/` returns zero matches. So the define is never set and **16 Odin-gated files are compiled by no CI leg**. My dotnet harness cannot cover them either, and the Unity MCP compile gate was pointed at the wrong project all session (#333).

What I could verify by hand, and did: `enumType` is the enclosing method's parameter so it is in scope; `underlyingMask` is declared before the loop that uses it; the `EnumShared` alias is present at line 17; and `GetUnderlyingTypeMask` / `IsPowerOfTwo` are both `public static` on the aliased type. The change is four lines and mirrors the non-Odin drawer exactly.

That is genuinely all the assurance available today, and I would rather say so than let it read as covered.

This is a pre-existing structural gap, not something this PR introduces — [#275](https://github.com/Ambiguous-Interactive/unity-helpers/issues/275) is the same blind spot shipping a compile break to consumers once already. Filed as [#347](https://github.com/Ambiguous-Interactive/unity-helpers/issues/347) with four remediation options; my pick is compiling against a minimal Sirenix stub, which catches the #275 class without needing a license.

It also bites this PR specifically: I changed the same flag filter in both `WEnumToggleButtonsDrawer` (four Unity legs cover it) and `WEnumToggleButtonsOdinDrawer` (nothing does). Divergence between that pair is invisible until a consumer with Odin installed hits it.

Runtime/Core/DataStructure/Adapters/SerializableList.cs
```diff
@@ -0,0 +1,435 @@
...
+                List<T> items = _items;
+                if (items == null || index < 0 || items.Count <= index)
+                {
+                    throw new ArgumentOutOfRangeException(
+                        nameof(index),
+                        $"{index} is outside of bounds [0, {Count})"
+                    );
+                }
+
+                return items[index];
```
wallstop: Why do this? Isn't this a double-bounds-check? Similar in set

Runtime/Core/DataStructure/Adapters/SerializableList.cs
```diff
@@ -0,0 +1,435 @@
...
+    /// and a null sequence is ignored. The indexer is the one member with no way to signal failure
+    /// -- it must produce a <typeparamref name="T"/> -- so it throws exactly as
+    /// <see cref="List{T}"/> does, because returning <c>default</c> for a bad index would hand the
+    /// caller a value that looks authored.
+    /// </para>
+    /// </remarks>
+    [Serializable]
+    [ProtoContract]
+    [JsonConverter(typeof(SerializableListConverterFactory))]
+    public sealed class SerializableList<T> : IList<T>, IReadOnlyList<T>
```
wallstop, line 63: This is nice, but does it mean that users have to explicitly use this? Why can't we just have consumers that decide to use our sets/dictionaries with this type have it "just work" automagically? I want, ideally, simplest-possible consumer/user code. Can we get there?

Runtime/Core/DataStructure/Adapters/SerializableList.cs
wallstop: You are right, it was a double check. Fixed in `e5dd74a1`.

`List<T>`'s own indexer performs the identical bounds test and throws the identical `ArgumentOutOfRangeException`; all mine added was a custom message and a null guard, and `EnsureItems()` already covers null. Both accessors are now one expression:

```csharp
public T this[int index]
{
    get => EnsureItems()[index];
    set => EnsureItems()[index] = value;
}
```

Behavior is unchanged — the existing `Assert.Throws<ArgumentOutOfRangeException>` test still passes, because that is exactly what `List<T>` throws. 82/82 in the harness.

On "similar in set": the set adapters have no indexer and no equivalent bounds check (`rg` for `this[int` / `Count <= index` across `SerializableHashSet`, `SerializableSortedSet`, `SerializableDictionary`, `SerializableSortedDictionary` finds nothing), so I read that as the setter, which the same change covers. If you meant something else there, point me at it.

The other members are deliberately *not* delegating, and I want to be explicit that it is a choice rather than an oversight: `RemoveAt` and `CopyTo` no-op on bad input and `Insert` clamps, per the repo's defensive-programming rule. The indexer is the one member that cannot express "did nothing" — it has to produce a `T` — so it is the one that throws.

Runtime/Core/DataStructure/Adapters/SerializableList.cs
wallstop: Yes, we can get there — backwards-compatibly. Design written up as [#348](https://github.com/Ambiguous-Interactive/unity-helpers/issues/348); I have deliberately not shipped it in this PR, reasons at the bottom.

**First, a correction I owe you.** #314 says the cache route *"needs a concrete `Cache<T>` subclass for every value type"*, and I repeated that in this PR's docs and commit message. **It is false.** The open generic satisfies the constraint directly — I compiled the hierarchy's generic shape to check:

```csharp
new Dict<string, List<float>, SD.Cache<List<float>>>();   // compiles, no subclass
```

So `SerializableDictionary<string, List<float>, SerializableDictionary.Cache<List<float>>>` already works today. Verbose, but zero boilerplate. I will fix the docs regardless of what we do with #348.

**The automagic version.** The blocker is that `_values` is `TValueCache[]`, so the two-arg form makes it `List<float>[]` — array-of-collection, which Unity drops. Add a second field to the base and choose between them per value type:

```csharp
[SerializeField] protected internal TValueCache[] _values;                     // existing
[SerializeField] private SerializableDictionary.Cache<TValueCache>[] _boxed;   // new

private static readonly bool RequiresBoxing =
    typeof(TValueCache).IsArray
    || (typeof(TValueCache).IsGenericType
        && typeof(TValueCache).GetGenericTypeDefinition() == typeof(List<>));
```

`Cache<List<float>>[]` is an array of a serializable **class**, which is the nesting Unity accepts.

**Why it is not a breaking change**, which is the part that makes it viable: ordinary value types keep using `_values` byte-identically — no migration, no YAML churn, no downgrade hazard — and collection value types previously stored *nothing*, so there is no old data to migrate. "Always box" would have rewritten every dictionary in every consumer project on next save; per-type selection avoids that entirely.

That rule is also exact rather than heuristic. It covers precisely the cases where `TValueCache[]` is dropped **and** boxing repairs it. Other unserializable value types (`Dictionary<,>`, interfaces, non-`[Serializable]` classes) are not helped by boxing — `Cache<Dictionary<,>>.Data` is equally unserializable — so those must keep showing the Inspector error rather than silently switching fields.

**Why not in this PR.** It touches the serialization core, a ~10k-line drawer that keys off `_values` throughout, and three adapter families — and right now it can be validated **nowhere**: Unity licensing is not configured in this devcontainer, the MCP compile gate resolves the wrong project (#333), the Odin path compiles in no CI leg (#347), and GitHub Actions has not scheduled a job in over an hour. Shipping an unvalidated change to how user data serializes is the exact failure mode #314 is about. I would rather hand you a design that is right than a diff that is unreviewable.

Happy to take it as the next PR the moment there is a Unity gate.

wallstop:
## Real Unity validation, at last — 107 passed / 0 failed

GitHub Actions has not run a Unity leg for this branch all session (see [#342](https://github.com/Ambiguous-Interactive/unity-helpers/issues/342) — jobs assigned, never picked up, reaped at exactly 15 minutes). So I drove the host editor directly over the MCP bridge instead.

**Compile gate, Unity 6000.4.6f1:**

```text
FRESH=39  STALE=0  MISSING=0
GIT_HEAD=refs/heads/dev/wallstop/nested-collection-serialization
```

Every package assembly is newer than its own newest source, and the editor is confirmed on this branch.

**EditMode run, filtered to the fixtures this PR touches:**

```text
pass=107  fail=0  skip=0  seconds=155.1
```

Covering `SerializableListTests`, `EnumExtensionTests`, `NestedCollectionSerializationTests`, `WEnumToggleButtonsDrawerTests`, `EditorCacheHelperEnumDisplayNameTests`, and `ScriptableObjectSingletonCreatorTests`.

### What that actually settles

Several claims in this PR were previously only argued, and are now measured **inside Unity**:

- **`SerializableList<T>` genuinely serializes.** `UnitySerializesValuesWrappedInASerializableList` passes against a real `SerializedObject` — the central claim of #314, which no amount of local C# testing could establish.
- **The `LogAssert` flake fix holds.** `ScriptableObjectSingletonCreatorTests` passes in full, including `HandlesCreateAssetExceptionWithPartialFile`.
- **Signed enum members round-trip through `SerializedProperty`**, not just through the dotnet harness.
- **The `Is.AnyOf` compile fix is correct** — `EditorCacheHelperEnumDisplayNameTests` compiled and ran under Unity's bundled NUnit.

### Three methodological traps worth recording

Getting to that number required correcting myself three times, and each wrong version produced a confident, wrong answer:

1. **"0 assemblies not built"** — checked only that the DLLs *exist*. A stale DLL from an earlier compile satisfies that. False green.
2. **"38 of 39 stale"** — compared every assembly against the newest source *anywhere in the package*. Unity builds incrementally, so an assembly whose own sources did not change is legitimately older. False red. The only meaningful test is each assembly against its own `sourceFiles`.
3. **The test run silently never started.** A `TestRunnerApi` created and not stored in a static is garbage collected; the status file sat at `starting` forever with **no error**. Holding the api and callbacks statically fixed it. (Also: never `Thread.Sleep` inside a probe — EditMode tests need the editor main loop, so sleeping there blocks the very thing being waited on.)

### Caveats, stated plainly

- **PlayMode has not been run**, only EditMode.
- **Odin is still uncovered.** `D:/Code/Packages` does not have Odin installed — `WEnumToggleButtonsOdinDrawer` compiles into no assembly there and `Sirenix.OdinInspector.Editor` is not loaded. That file remains read-verified only, per [#347](https://github.com/Ambiguous-Interactive/unity-helpers/issues/347).
- This ran on **6000.4.6f1**; CI's matrix spans 2021.3 / 2022.3 / 6000.3 / 6000.5, and the one leg that failed earlier in this PR was 6000.5.
- The runner bridge lives in the host project at `Assets/Editor/UnityHelpersMcpTestRunner.cs`, writes only to `Temp/`, and is labelled as safe to delete.

PR #352, opened 2026-08-07
Make an interrupted save survivable, unblock Dependabot CI, and stop paying for a parallelism cap

### LINKED ISSUE #319: feature: upstream DurableFile — FileHelper has no durable write, and Serializer's four File.WriteAllText sites can lose player data
## Summary

`FileHelper` currently offers `InitializePath` and `CopyFileAsync`. Neither is *durable*: a write that is
interrupted part-way can leave a truncated file where a valid one used to be, and a write that returns has
only reached the OS page cache, not the disk. For "project asset file management" — `FileHelper`'s stated
focus — that is fine. For **player-owned data** (saves, settings, ledgers) it is a data-loss bug, and it is
one every consumer currently has to re-solve.

We built `DurableFile` in another Ambiguous-Interactive project to close exactly this. It has been in
production since, is covered by 22 test cases, and is entirely
Unity-agnostic — pure `System.IO`, no `UnityEngine` reference. **We'd like to upstream it** so the
durability rules live in one place instead of being re-derived per project, and delete our copy.

Happy to open the PR ourselves if the API shape below is agreeable.

## Why this belongs here rather than in each consumer

Three arguments, in increasing order of strength.

### 1. The package already writes user data non-durably

`Runtime/Core/Serialization/Serializer.cs` has four sites:

```
3223:  File.WriteAllText(path, jsonAsText);
3236:  await File.WriteAllTextAsync(path, jsonAsText);
3272:  File.WriteAllText(path, jsonAsText);
3289:  await File.WriteAllTextAsync(path, jsonAsText);
```

`File.WriteAllText` **truncates the destination before writing a single byte**. If the process dies, the
device loses power, or the disk fills between the truncate and the last byte, the previous contents are
gone and the new ones are incomplete. These are the package's JSON-to-disk helpers, so whatever a consumer
persists through them — a save file, a settings document — is exposed to that window on *every* write.
This is not hypothetical: it is the failure that prompted our implementation.

### 2. `DurableFile.CopyAsync` already has a `FileHelper`-shaped hole in it

Our signature takes the copy as a delegate specifically so the type stays free of any particular async file
helper:

```csharp
public static async ValueTask<Exception> CopyAsync(
    string sourcePath,
    string destinationPath,
    Func<string, string, ValueTask<bool>> copyAsync)
```

and every one of our call sites passes `FileHelper.CopyFileAsync`. Upstreamed, that parameter collapses
into a direct call and the indirection disappears. The two types were converging on each other already.

### 3. The rules are subtle enough that per-project re-derivation gets them wrong

The interesting part of this code is not the staging; it is knowing *which* guarantee you actually have.
Ours carries an explicit non-guarantee (see [Scope](#scope-what-this-does-and-does-not-promise)) precisely
because an earlier draft of ours overclaimed it. A shared implementation is where that nuance can be
written down once.

## Proposed API

Namespace `WallstopStudios.UnityHelpers.Core.Helper`, either as a new `DurableFile` static class beside
`FileHelper` or folded into `FileHelper` itself — **your call**; I'd lean to a separate type, because the
name is the documentation ("durable" is the whole point) and `FileHelper`'s existing docs promise something
weaker.

```csharp
public static class DurableFile
{
    // Sentinel suffix for the staged sibling. Public so consumers can recognise/ignore leftovers.
    public const string TemporarySuffix = ".tmp";

    // Whole-document rewrite: stage -> fsync -> atomic swap.
    public static bool TryWriteAllText(string path, string contents, out Exception error);
    public static ValueTask<Exception> WriteAllTextAsync(string path, string contents);

    // Record append: O_APPEND + fsync. Preferred where the format allows it.
    public static bool TryAppendAllText(string path, string contents, out Exception error);
    public static ValueTask<Exception> AppendAllTextAsync(string path, string contents);

    // Replace a file with another file's contents, same staging discipline.
    public static bool TryCopy(string sourcePath, string destinationPath, out Exception error);
    public static ValueTask<Exception> CopyAsync(string sourcePath, string destinationPath);

    public static bool TryDelete(string path);
}
```

Conventions chosen to match the package's existing style — `TryXxx` + `out Exception` for sync, and
`ValueTask<Exception>` (null on success) for async, because these are I/O paths where throwing at the
caller is the thing we are trying to avoid. If you'd rather they threw, or returned a `OneOf`-style result
(the package already has `Runtime/Core/OneOf/`), say so and I'll adapt.

### `TryWriteAllText` — the core of it

1. Reject a blank path up front, returning `false` with an `ArgumentException`. Never throw.
2. `Directory.CreateDirectory` for the parent, so callers do not each have to.
3. Write the bytes to `path + ".tmp"` with `FileMode.Create, FileAccess.Write, FileShare.None`.
4. **`stream.Flush(flushToDisk: true)`** before closing — this is the fsync; without it you have staged
   nothing, you have only moved the same page-cache problem to a different filename.
5. Swap. `File.Replace(temporary, path, destinationBackupFileName: null)` when the destination exists,
   `File.Move` when it does not — **`File.Replace` throws if the destination is absent**, which is the one
   sharp edge in this whole design.
6. On any failure, delete the temporary and return the exception. A leftover temporary would be read as an
   interrupted write by the next attempt and, if the cause was a full disk, would keep failing identically.

Encoding: **UTF-8 with no BOM** (`new UTF8Encoding(encoderShouldEmitUTF8Identifier: false)`). A BOM breaks
strict JSON parsers and quietly changes file bytes on every round-trip, so it is worth pinning explicitly
rather than inheriting a default.

### `TryAppendAllText`

`FileMode.Append` maps to `O_APPEND` / `FILE_APPEND_DATA`, so a write cannot damage an earlier one and
concurrent writers interleave instead of clobbering. No staging needed — just append and
`Flush(flushToDisk: true)`. Use `FileShare.Read` so readers are not locked out. Empty contents is a
**no-op success**, not an error, and must not create the file.

Worth stating in the docs: where the format allows a log of records, append is strictly stronger than
rewrite, and consumers should prefer it.

### `TryCopy` / `CopyAsync`

Same staging discipline. One trap: **`File.Copy` does not fsync**, so the staged copy has to be reopened
purely to `Flush(flushToDisk: true)` before the swap makes it live. `CopyAsync` can delegate the bulk copy
to the existing `FileHelper.CopyFileAsync`; the fsync and the rename stay synchronous, because splitting a
flush and a rename across an `await` buys nothing.

### Async specifics

`ConfigureAwait(false)` on every await. These are I/O paths with no need for Unity's main thread, and a
caller that blocks on the result would otherwise be able to deadlock itself against the Unity
synchronization context. Open the stream with `useAsync: true`.

### Platform / profile constraints to check

I have **not** verified these against the package's full support matrix, and they need to be before merge:

- **`File.Replace` is not universally supported.** It is the atomic-swap primitive here, but it can throw
  `PlatformNotSupportedException` on some targets, and WebGL has no real filesystem at all. A documented
  fallback (delete-then-move, which is *not* atomic and should say so) or an explicit platform carve-out is
  needed. Our own usage is desktop-first, so we have not exercised this.
- **`FileHelper` already notes a .NET Standard 2.0 / Unity 2021.3 floor** (the comment on `CopyFileAsync`
  about avoiding `IAsyncDisposable`). Note that `File.WriteAllTextAsync` — used in `Serializer` today —
  does not exist in .NET Standard 2.0; the implementation above deliberately uses `FileStream.WriteAsync`
  instead, which does.

## Scope: what this does and does not promise

**Please keep this in the XML docs.** An earlier draft of ours overclaimed it and it took a review to catch.

- It **does** eliminate the torn-file window that `File.WriteAllText` opens on every write. A reader sees
  either the complete previous contents or the complete new ones.
- It **does** force the data out of the page cache before the swap.
- It is **not** full crash-safety. .NET cannot fsync a *directory*, so the rename itself can still be
  reordered behind the data write by the filesystem. Consumers of this type must not be described as
  crash-safe.

That third bullet is the reason to upstream rather than let each project reinvent it: the implementation is
easy, and knowing exactly what it is worth is not.

## Tests

Our fixture is **16 test methods / 22 executed cases** and ports directly (NUnit, EditMode, no Unity API — a temp directory in `SetUp`, a
best-effort recursive delete in `TearDown` that never fails the run):

| Test | What it pins |
| --- | --- |
| `BlankPathsFailWithoutThrowing` | null / `""` / `"   "` across write, append and copy — returns `false` with an exception, never throws |
| `WriteCreatesMissingDirectories` | nested path is created |
| `WriteReplacesExistingContentEntirely` | **a shorter replacement leaves no tail of the previous document** — the test that actually catches a naive in-place rewrite |
| `WriteLeavesNoTemporaryBehind` | no `.tmp` after success |
| `FailedWriteLeavesThePreviousContentsIntact` | the whole point of the feature |
| `WriteRoundTripsUnicodeWithoutAByteOrderMark` | pins the encoding decision |
| `WriteTreatsNullContentsAsEmpty` | null is empty, not a crash |
| `AppendCreatesThenAccumulates` | create-on-first-append, then accumulate |
| `AppendingNothingSucceedsWithoutCreatingAFile` | empty append is a no-op success and creates nothing |
| `CopyReplacesTheDestination` / `AFailedCopyLeavesTheDestinationIntact` | sync copy, both directions |
| `CopyAsyncReplacesTheDestination` / `CopyAsyncLeavesTheDestinationIntactWhenTheStagedCopyFails` | async copy, both directions |
| `CopyAsyncRejectsBlankPathsWithoutTouchingTheDisk` | validation happens before any I/O |
| `CopyAsyncRejectsAMissingCopyDelegate` | (drops out if the delegate parameter goes away upstream) |
| `CopyFromAMissingSourceLeavesTheDestinationIntact` | a missing source is not allowed to destroy the destination |

The failure-injection tests work by making the staged path un-writable (e.g. pre-creating `path + ".tmp"`
as a *directory*), which forces the staging step to fail without needing to kill the process.

Two additions worth having upstream that we do not currently cover, because they are package-level
concerns rather than ours:

- **`File.Replace` fallback behaviour** on any target where it is unsupported, per the constraint above.
- **A concurrent-append test** — two writers appending simultaneously must interleave whole records rather
  than clobber, which is the property `O_APPEND` is being relied on for.

## What we'd do on our side

Delete `Assets/Scripts/Data/Persistence/DurableFile.cs` and its fixture, and repoint our four consumers
(`SettingsAccess`, `PersistentAchievementBackend`, `LocalFileAdapter`, and the save/settings copy paths) at
`WallstopStudios.UnityHelpers.Core.Helper.DurableFile`. That is a pure delete on our side — the API above is
already exactly what we call.

## References

The originating project is a **private** repo, so rather than link files you cannot open, the full
implementation is inlined above and below — this issue is self-contained.

Originating bug: changing a setting and quitting could lose it outright. `File.WriteAllText` was running on
a thread pool that abandons its queue on dispose, so the truncate had happened and the write had not. The
fix was this type plus a synchronous flush on `Dispose` / `OnApplicationPause`.

<details>
<summary>Full current implementation (MIT, happy to relicense to match the package)</summary>

```csharp
public static class DurableFile
{
    public const string TemporarySuffix = ".tmp";

    private const int DefaultBufferSize = 4096;

    private static readonly UTF8Encoding Utf8NoByteOrderMark = new(
        encoderShouldEmitUTF8Identifier: false);

    public static bool TryWriteAllText(string path, string contents, out Exception error)
    {
        if (string.IsNullOrWhiteSpace(path))
        {
            error = new ArgumentException("A destination path is required.", nameof(path));
            return false;
        }

        string temporaryPath = path + TemporarySuffix;
        try
        {
            EnsureDirectory(path);
            WriteAndFlush(temporaryPath, Utf8NoByteOrderMark.GetBytes(contents ?? string.Empty));
            Swap(temporaryPath, path);
            error = null;
            return true;
        }
        catch (Exception e)
        {
            error = e;
            // A leftover temporary reads as a half-finished write to the next attempt, and would
            // keep failing identically if the cause was a full disk.
            TryDelete(temporaryPath);
            return false;
        }
    }

    public static bool TryAppendAllText(string path, string contents, out Exception error)
    {
        if (string.IsNullOrWhiteSpace(path))
        {
            error = new ArgumentException("A destination path is required.", nameof(path));
            return false;
        }

        if (string.IsNullOrEmpty(contents))
        {
            error = null;
            return true;
        }

        try
        {
            EnsureDirectory(path);
            byte[] bytes = Utf8NoByteOrderMark.GetBytes(contents);
            using FileStream stream = new(path, FileMode.Append, FileAccess.Write, FileShare.Read);
            stream.Write(bytes, 0, bytes.Length);
            stream.Flush(flushToDisk: true);
            error = null;
            return true;
        }
        catch (Exception e)
        {
            error = e;
            return false;
        }
    }

    public static async ValueTask<Exception> WriteAllTextAsync(string path, string contents)
    {
        if (string.IsNullOrWhiteSpace(path))
        {
            return new ArgumentException("A destination path is required.", nameof(path));
        }

        string temporaryPath = path + TemporarySuffix;
        try
        {
            EnsureDirectory(path);
            byte[] bytes = Utf8NoByteOrderMark.GetBytes(contents ?? string.Empty);
            using (FileStream stream = new(temporaryPath, FileMode.Create, FileAccess.Write,
                       FileShare.None, DefaultBufferSize, useAsync: true))
            {
                await stream.WriteAsync(bytes, 0, bytes.Length).ConfigureAwait(false);
                stream.Flush(flushToDisk: true);
            }

            Swap(temporaryPath, path);
            return null;
        }
        catch (Exception e)
        {
            _ = TryDelete(temporaryPath);
            return e;
        }
    }

    public static bool TryCopy(string sourcePath, string destinationPath, out Exception error)
    {
        // ... blank-path validation elided for brevity, same shape as above ...
        string temporaryPath = destinationPath + TemporarySuffix;
        try
        {
            EnsureDirectory(destinationPath);
            File.Copy(sourcePath, temporaryPath, overwrite: true);
            FlushToDisk(temporaryPath);
            Swap(temporaryPath, destinationPath);
            error = null;
            return true;
        }
        catch (Exception e)
        {
            error = e;
            TryDelete(temporaryPath);
            return false;
        }
    }

    private static void EnsureDirectory(string path)
    {
        string directory = Path.GetDirectoryName(path);
        if (!string.IsNullOrEmpty(directory))
        {
            _ = Directory.CreateDirectory(directory);
        }
    }

    private static void WriteAndFlush(string path, byte[] bytes)
    {
        using FileStream stream = new(path, FileMode.Create, FileAccess.Write, FileShare.None);
        stream.Write(bytes, 0, bytes.Length);
        stream.Flush(flushToDisk: true);   // the fsync; without it nothing has been staged
    }

    // File.Copy does not fsync, so the staged copy is reopened purely to force it out of the
    // page cache before the swap makes it the live file.
    private static void FlushToDisk(string path)
    {
        using FileStream stream = new(path, FileMode.Open, FileAccess.Write, FileShare.None);
        stream.Flush(flushToDisk: true);
    }

    private static void Swap(string temporaryPath, string path)
    {
        if (File.Exists(path))
        {
            // File.Replace is the atomic swap, but requires the destination to already exist.
            File.Replace(temporaryPath, path, destinationBackupFileName: null);
            return;
        }

        File.Move(temporaryPath, path);
    }

    public static bool TryDelete(string path)
    {
        if (string.IsNullOrWhiteSpace(path))
        {
            return false;
        }

        try
        {
            if (File.Exists(path))
            {
                File.Delete(path);
            }

            return true;
        }
        catch (Exception)
        {
            return false;
        }
    }
}
```

</details>

wallstop:
Three things, in descending order of how much they can hurt a player.

## 1. An interrupted save destroyed the previous one (#319)

`File.WriteAllText` empties the destination before it writes a byte. A crash, a power loss, or a full disk in that window replaces a valid save with a truncated one. All five of `Serializer`'s JSON write paths had it, so anything a consumer persists through them was exposed on **every** write.

`DurableFile` stages the document in a sibling file, flushes it to disk, and swaps it over the destination. `WriteToJsonFile` / `WriteToJsonFileAsync` route through it and keep their throwing contract (`ExceptionDispatchInfo`, so the original I/O stack survives). Existing bytes are unchanged: UTF-8, no BOM, same as before.

The type is the one #319 offered to upstream, with two corrections found by **running** the fixture rather than reading it:

- **.NET's `FileMode.Append` is not `O_APPEND`.** It is a seek-to-end at open time, on Linux and Windows alike. Four threads appending 50 records each kept **155 of 200** — the rest were silently overwritten. That is the exact property #319 said consumers were relying on. Every operation now takes a striped per-path gate; the same test is 200/200.
- **A failed staging attempt deleted the staged file unconditionally**, including one a concurrent writer owned. It is now discarded only when this call opened it.

The XML docs carry the non-guarantee #319 asked to keep: this is not crash safety, because .NET cannot flush a directory.

**Verification.** Red→green under `dotnet` against the real sources and the real fixture: a fully naive `File.WriteAllText` fails `FailedWriteLeavesThePreviousContentsIntact`, the gate takes the concurrent-append test from 155/200 to 200/200, and the fixture is 22/22 green. Unity compile gate via the MCP bridge: `FRESH=39 STALE=0 MISSING=0`.

## 2. Every Dependabot pull request was permanently red, and none of it was the dependency's fault

Dependabot draws from a separate secret store, so `BUILD_LOCK_READER_APP_ID` arrives empty. `runner-preflight` was the one job in the workflow that reached for organization secrets **without first asking whether this run has any** — the availability action failed closed on "Missing required input", `Unity CI Success` required that job to have succeeded, and the verdict was a guaranteed failure. Meanwhile `matrix-config` had already, and correctly, skipped every licensed tier those credentials exist to serve. Preflighting a runner fleet nothing will be dispatched to is not a gate.

`runner-preflight` now detects the credentials in a step and skips the probe when they are absent. `Unity CI Success` gains an explicit branch for that case, beside the fork and unprotected-push branches it already had.

It is an assertion, not a waiver: both hosted gates must have passed and all four licensed jobs must report `skipped`, or it is still red. The step summary says plainly that **no Unity test executed** and that coverage comes from the protected-branch push after merge. Three contract assertions pin it, verified red against the pre-fix workflow.

This unblocks #334.

## 3. CI wall clock: the parallelism cap was costing more than it bought (#326)

`max-parallel: 2` reads like it matches the two Unity runners. It does not gate execution — a self-hosted runner already takes one job at a time — it gates *eligibility*, so every completed leg needs a fresh dispatch transition, and dispatch to a self-hosted runner is the slow part.

Measured on run [`31130259492`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31130259492) (main, green, 88 min wall clock):

| | |
| --- | --- |
| Licensed legs | 16 |
| Sum of job durations | 70.9 min |
| Runners | 2 |
| Packing floor | **35.5 min** |
| Actual wall clock | **88 min** |

`ELI-MACHINE` finished a leg at +9.4 min and started its next at +22.8, with four legs queued throughout. The standalone tier was created at +32.0 and its first leg started at +42.1, both runners idle for those ten minutes.

Two alternative explanations, both checked and ruled out:

- **Sibling contention.** No Ambiguous-Interactive repo touched the self-hosted runners in that window (checked DxMessaging, DoxReloaded, IshoBoy, qora-redux).
- **The organization Unity lock.** `Acquire organization Unity lock` took **2–4 s on all 15 licensed legs**. It never serialized anything in this run.

That second point is worth flagging on its own: [`PLAN.md`](../blob/main/PLAN.md) lists "the org build-lock serializes ALL Unity CI legs" as a verified hard constraint, "do not re-litigate". As of this run it is not one, and several conclusions downstream of it deserve re-examination.

Concurrency is still bounded by the runner pool and the organization lock, so no correctness property changes. What is traded is politeness — a sibling repo's Unity job can now queue behind more of ours at once. The contract that pinned `max-parallel: 2` is inverted to forbid a cap, carrying the measurement so it cannot drift back without new evidence.

**The confirmation is this pull request's own Unity run.** With an 88 min wall clock over a 35.5 min floor, a real improvement should show up as a shrunken gap between one leg finishing and the next starting. If it does not, the cap was not the cause and this commit should be reverted rather than kept on theory.

---

Closes #319
Refs #326, #334

---

> [!NOTE]
> **Medium Risk**
> Changes affect player save durability and all JSON file persistence paths; CI gating and self-hosted job scheduling behavior also shift, though tests and contract scripts cover the main cases.
> 
> **Overview**
> **Player saves and JSON file writes** no longer truncate the destination on failure. New **`DurableFile`** stages content in a sibling `.tmp` file, flushes, then swaps; **`Serializer.WriteToJsonFile`** / async overloads delegate to it while keeping the same throwing behavior. Per-path striped gates fix concurrent append corruption. **`SemaphoreLease`** plus **`Acquire` / `TryAcquire` / `AcquireAsync`** extensions support `using`-based critical sections (used by `DurableFile`).
> 
> **Unity CI** stops failing Dependabot PRs: **`runner-preflight`** probes build-lock reader secrets in a step and skips the self-hosted fleet check when they are missing; **`Unity CI Success`** treats credential-less runs like fork PRs—pass when hosted gates succeed and all licensed jobs are **`skipped`**, with an explicit summary that no Unity tests ran.
> 
> **Matrix jobs** drop **`max-parallel: 2`** on Unity self-hosted tiers (and benchmarks), with measured rationale that the cap limited eligibility and slowed dispatch; the workflow contract script now **forbids** a parallelism cap. Docs and **`.llm`** references add filesystem/dispose/semaphore patterns; integration tests cover **`DurableFile`** and **`SemaphoreLease`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ff2a2d29cc61515966180ef2b87627185d677c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Helper/DurableFile.cs
```diff
@@ -0,0 +1,576 @@
...
+                    FlushToDisk(temporaryPath);
+                    Swap(temporaryPath, destinationPath);
+                    error = null;
+                    return true;
+                }
+                catch (Exception e)
+                {
+                    error = e;
+                    DiscardStagedFile(temporaryPath);
+                    return false;
```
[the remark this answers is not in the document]
wallstop: Correct, and already fixed in `2fef678d`, which landed after the commit reviewed here (`c9e22d7f`). Independently found in the same pass, which is a good sign for the finding.

Both copy paths now track whether *this* call created the staged file and only clean up what is theirs:

```csharp
bool staged = false;
try
{
    EnsureDirectory(destinationPath);
    File.Copy(sourcePath, temporaryPath, overwrite: true);
    staged = true;
    ...
}
catch (Exception e)
{
    error = e;
    if (staged) { DiscardStagedFile(temporaryPath); }
    return false;
}
```

`ValidateCopyPaths` also now probes the source up front, so the most common pre-staging failure (a missing source) never reaches the staging step at all.

The same commit closes a second hole in the same family that this review did not flag: encoding ran *inside* the try that owned the open staging handle, so a throw from `GetBytes` left the `FileStream` undisposed and the subsequent `TryDelete` then failed on Windows against a file this process still held. Encoding now happens before the handle is opened.

Runtime/Core/Helper/DurableFile.cs
```diff
@@ -0,0 +1,598 @@
...
+        /// <returns>True when the destination holds the new contents.</returns>
+        public static bool TryWriteAllText(string path, string contents, out Exception error)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                error = new ArgumentException("A destination path is required.", nameof(path));
+                return false;
+            }
+
+            SemaphoreSlim gate = GateFor(path);
```
wallstop: Can we expose using/IDisposable zero-cost concepts here and elsewhere (that we can expose to callers) so we can just do `using` instead of try/finally?

wallstop:
## The parallelism confirmation the description promised

Run [`31145238181`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31145238181), the first with the cap removed. Relative minutes from run creation:

```text
  Unity 6000.3.16f1 playmode   ELI-MACHINE   start +0.4   end +2.9
  Unity 6000.5.2f1  playmode   ELI-MACHINE   start +2.9   end +5.5     <- 0.0 min handoff
  Unity 2021.3.45f1 playmode   ELI-MACHINE   start +5.5   end +8.2     <- 0.0 min handoff
  Unity 2022.3.45f1 playmode   ELI-MACHINE   start +8.2
  Unity 2021.3.45f1 editmode   DAD-MACHINE   start +1.3   end +8.6
  Unity 6000.3.16f1 editmode   DAD-MACHINE   start +9.1                <- 0.5 min handoff
```

Against the baseline run [`31130259492`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31130259492), same runner:

```text
  Unity 2021.3.45f1 editmode   ELI-MACHINE   start +0.3   end +5.2
  Unity 6000.3.16f1 editmode   ELI-MACHINE   start +5.2   end +9.4
  Unity 2022.3.45f1 playmode   ELI-MACHINE   start +22.8               <- 13.4 min handoff
```

All eight fast-tier legs became eligible at **+0.3 min** instead of two at a time. Six legs completed in 9.1 min of wall clock carrying 15.5 min of work across two runners — a 7.75 min floor, so the packing is now close to optimal where it was 40% efficient before.

That is the measurement the description said would decide whether this commit is kept. It is kept.

## Verification that the new tests actually ran

`pass=0 fail=0` is the false green worth fearing with a new fixture, so from the 2021.3 PlayMode artifact of that run: **`total=8723 passed=8703 failed=0 skipped=4`**, with all 22 `DurableFileTests` cases present and passing — including `ConcurrentAppendsInterleaveWholeRecords` at 0.741 s, which is the case that fails without the per-path gate. Notably it passes on Windows too, confirming `FileMode.Append` is non-atomic there as well and not just on the Linux harness where it was found.

## Two review notes

- **`copilot-pull-request-reviewer` is failing** with *"the user who requested the review has reached their quota limit."* That is an account quota, not anything in this branch.
- **Cursor Bugbot flagged the staged-file ownership bug in the copy paths** on `c9e22d7f`. It was already fixed in `2fef678d`, which also closes a second hole in the same family that the review did not flag — encoding running inside the try that owned the open staging handle, stranding a `FileStream` on a throw from `GetBytes`. Replied in the thread.

Runtime/Core/Helper/DurableFile.cs
```diff
@@ -0,0 +1,598 @@
...
+        public static bool TryDelete(string path)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                return false;
+            }
+
+            try
+            {
+                if (File.Exists(path))
```
wallstop, line 459: Is there a TOCTOU bug here?

Runtime/Core/Helper/DurableFile.cs
```diff
@@ -0,0 +1,598 @@
...
+            {
+                key = Path.GetFullPath(path);
+            }
+            catch (Exception)
+            {
+                // A path that cannot be normalized still gets a gate, just not one shared with its
+                // aliases.
+            }
+
+            int hash = StringComparer.OrdinalIgnoreCase.GetHashCode(key);
```
wallstop, line 527: hash can be negative, is bitwise and with max value really the safest, guaranteed way to enforce positve? Why not use our positive mod helper?

Runtime/Core/Helper/DurableFile.cs
wallstop: Done in `247d6d1c` — `readonly struct GateScope : IDisposable`, so every gated block is now `using (EnterGate(path)) { … }` instead of five lines of `try`/`finally`. Zero allocation: `using` over a struct calls `Dispose` directly with no boxing.

```csharp
private readonly struct GateScope : IDisposable
{
    private readonly SemaphoreSlim _gate;

    internal GateScope(SemaphoreSlim gate) => _gate = gate;

    public void Dispose() => _gate.Release();
}
```

One asymmetry worth flagging: the `async` paths acquire **before** the `using`, not inside it —

```csharp
GateScope gate;
try { gate = await EnterGateAsync(path, cancellationToken).ConfigureAwait(false); }
catch (Exception e) { return e; }

using (gate) { … }
```

because `using (await EnterGateAsync(...))` would let a cancelled wait unwind through a scope whose gate was never taken, releasing a semaphore nobody acquired and permanently raising its count.

On "and elsewhere": I swept `Runtime/` for `try`/`finally` around a lock release and this was the only one, so there is nothing else to convert right now. The idiom already exists in the repo (`AssetDatabaseBatchScope`, `GroupGUIWidthUtility.PushContentPadding`, `LabelWidthScope`), so I wrote it down as the expected pattern rather than leaving it as folklore — [forbidden-patterns § Filesystem, Scope, and Index Patterns](https://github.com/Ambiguous-Interactive/unity-helpers/blob/dev/wallstop/durable-file-and-ci-throughput/.llm/references/forbidden-patterns.md#filesystem-scope-and-index-patterns), with a pointer from `defensive-programming`.

Nothing here is public, so there is no caller-facing scope to expose from `DurableFile` itself.

Runtime/Core/Helper/DurableFile.cs
wallstop: Not at that line, but yes one line down — thanks for making me look.

**`TryDelete` (the line asked about) is benign.** `File.Delete` is already a no-op on a missing file, so losing the race just means somebody else deleted it first, and the postcondition this method reports is "no file remains", not "we removed it". The probe is a fast path to avoid exception cost, not a guard. It now says so:

```csharp
// File.Delete is already a no-op on a missing file, so the probe is a fast path
// rather than a correctness guard, and losing the race with another deleter is
// harmless: the postcondition this reports is "no file remains", not "we removed it".
```

**`Swap` was the real one.** There the probe chose *which API is legal*, and only one direction of the race recovered:

```csharp
if (File.Exists(path)) { File.Replace(temp, path, null); }   // throws FileNotFoundException
                                                             // if the destination just vanished
else { File.Move(temp, path); }                              // this side already recovered
```

Fixed in `247d6d1c` — `Replace` now falls back to `File.Move` on `FileNotFoundException`, so both directions are handled and the probe is back to being a hint.

Swept the rest: `Runtime/` has 40 `File.Exists` call sites; the only other probe-then-act is `FileHelper.InitializePath`, which is already correctly backed by `FileMode.CreateNew` (a racing creator loses the open, and the catch handles it). Written up as a rule so the next one does not have to be found by review.

Runtime/Core/Helper/DurableFile.cs
wallstop: Switched to the helper in `247d6d1c`: `Gates[hash.PositiveMod(GateCount)]`.

To answer the question directly, since "safest" and "correct" come apart here:

- `hash & int.MaxValue` **is** guaranteed non-negative — it clears the sign bit, and unlike `Math.Abs` it has no `int.MinValue` trap (`Math.Abs(int.MinValue)` throws `OverflowException`). So it was not a latent crash.
- It is worse in a way that matters less but is still real: discarding the sign bit maps `h` and `h | int.MinValue` to the same value, so two distinct hashes always collide into the same gate. `PositiveMod` keeps that bit.
- And it is a magic idiom where the project already owns the named operation, which is the actual argument.

`WallMath.PositiveMod(int, int)` is safe at the boundary too — `value %= max` first, so `value + max` cannot overflow.

Swept `Runtime/` for the class: the only other `& int.MaxValue` is `RandomPerformanceTests.DeriveSeed`, which is a deliberate `unchecked` seed derivation rather than a bucket index, and there are no `Math.Abs(...) % n` sites at all. Both forms are now listed as forbidden with `PositiveMod` as the replacement in [forbidden-patterns](https://github.com/Ambiguous-Interactive/unity-helpers/blob/dev/wallstop/durable-file-and-ci-throughput/.llm/references/forbidden-patterns.md#filesystem-scope-and-index-patterns).

Runtime/Core/Helper/DurableFile.cs
```diff
@@ -0,0 +1,598 @@
...
+                    Swap(temporaryPath, destinationPath);
+                    error = null;
+                    return true;
+                }
+                catch (Exception e)
+                {
+                    error = e;
+                    if (staged)
+                    {
+                        DiscardStagedFile(temporaryPath);
```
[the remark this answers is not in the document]
wallstop: Correct, and this one was a regression from my own previous fix — the ownership guard I added for the earlier finding created this leak. Fixed properly in `247d6d1c`.

Moving the flag earlier would just have re-broken the ownership case, because `File.Copy` folds **"could not take the staging path"** and **"failed halfway through writing it"** into one exception, and the cleanup decision differs between them. So copy no longer uses `File.Copy`. It opens the staging stream itself with the same `FileMode.Create` + `FileShare.None` the write paths use, and streams the source in:

```csharp
try { staging = OpenStagingStream(temporaryPath, useAsync: false); }
catch (Exception e) { source.Dispose(); error = e; return false; }   // not ours

try
{
    using (source)
    using (staging)
    {
        source.CopyTo(staging, DefaultBufferSize);
        staging.Flush(flushToDisk: true);
    }
    Swap(temporaryPath, destinationPath);
}
catch (Exception e) { error = e; DiscardStagedFile(temporaryPath); return false; }   // ours
```

That leaves the file with exactly one ownership rule, shared by write and copy: **the staged file is ours if and only if our exclusive open returned.** Failures before it touch nothing; failures after it always clean up, partial file included. The separate reopen-to-fsync also disappears, since we now hold the handle we need to flush.

Pinned by three tests, each verified to fail against a mutation of the thing it pins — `AFailureBeforeStagingLeavesAnExistingStagedFileAlone` (mutation: discard unconditionally) and `AFailureAfterStagingRemovesTheStagedFile` / `AFailedWriteAfterStagingRemovesTheStagedFile` (mutation: never discard).

Runtime/Core/Helper/DurableFile.cs
```diff
@@ -0,0 +1,652 @@
...
+                // complete and flushed, but the destination is briefly absent.
+                File.Delete(path);
+                File.Move(temporaryPath, path);
+            }
+        }
+
+        /// <summary>
+        /// Holds a path's gate until disposed. A struct so <c>using</c> costs no allocation.
+        /// </summary>
+        private readonly struct GateScope : IDisposable
```
wallstop: We should promote this to a publicly accessible utility that wraps semaphoreslim on both the sync and async paths, along with tests.

It should also track disposal, ideally

Runtime/Core/Helper/DurableFile.cs
wallstop: **Correction to my reply above, before the new work: the reason I gave was wrong.**

I said `using (await EnterGateAsync(...))` "would let a cancelled wait unwind through a scope whose gate was never taken." That is not how `using` works. The operand is evaluated *before* the scope is established — `using (X) { … }` is `var tmp = X; try { … } finally { tmp.Dispose(); }` — so if `X` throws, the block is never entered and `Dispose` never runs. `using (await sem.AcquireAsync(token))` is safe.

The acquire-outside-the-`using` shape in `DurableFile` is still correct, but for a different reason: those methods must **return** the exception rather than throw it (never-throw-from-public-APIs), so the `await` needs a `catch` regardless. I have documented the real semantics on `AcquireAsync` rather than the wrong ones.

**On the promotion request — done in the commit that follows this.**

`Runtime/Core/Threading/SemaphoreLease.cs`, public, both paths, disposal tracked:

```csharp
public struct SemaphoreLease : IDisposable
{
    private SemaphoreSlim _semaphore;

    public bool IsHeld => _semaphore != null;

    public void Dispose()
    {
        SemaphoreSlim semaphore = _semaphore;
        if (semaphore == null) { return; }   // already released
        _semaphore = null;
        semaphore.Release();
    }
}

public static class SemaphoreSlimExtensions
{
    public static SemaphoreLease Acquire(this SemaphoreSlim semaphore);
    public static bool TryAcquire(this SemaphoreSlim semaphore, out SemaphoreLease lease);
    public static bool TryAcquire(this SemaphoreSlim semaphore, TimeSpan timeout, out SemaphoreLease lease);
    public static ValueTask<SemaphoreLease> AcquireAsync(this SemaphoreSlim semaphore, CancellationToken cancellationToken = default);
}
```

Three design notes, since each is a judgement call:

1. **Disposal tracking needs a mutable struct, not a `readonly` one.** `Dispose` nulls the field, so a second `Dispose` on the same local is a no-op. That is the property worth having — a stray `Release()` raises the count *above* the semaphore maximum and quietly admits a second caller to a section built for one. The limit is honest and documented: a *copy* of a lease carries a live reference, so copying is called out as forbidden in the XML docs.
2. **`Acquire` throws `ArgumentNullException` on a null semaphore** rather than returning an unheld lease. This is the rare case where the never-throw rule points the wrong way: a lease that silently is not held leaves the critical section unguarded, and that corruption surfaces nowhere near its cause. `TryAcquire` is the non-throwing door.
3. **`TryAcquire` returns a lease that is not held on failure**, and disposing it is a no-op — so `if (TryAcquire(...)) { using (lease) { … } }` and a stray `lease.Dispose()` on the failure path are both safe.

`DurableFile` now uses it; its private `GateScope` is gone.

**11 tests** in `Tests/Runtime/Threading/SemaphoreLeaseTests.cs`, each verified to fail against a mutation of the thing it pins:

| Mutation | Tests that go red |
| --- | --- |
| `Dispose` releases unconditionally (no tracking) | `DisposingTwiceReturnsOnlyOnePermit`, `ADefaultLeaseIsNotHeldAndDisposesHarmlessly`, `TryAcquireReportsFailureWhenNoPermitIsFree` |
| `TryAcquire` ignores the `Wait` result | `TryAcquireReportsFailureWhenNoPermitIsFree`, `TryAcquireReportsFailureWhenTheTimeoutElapses` |
| `Acquire` returns without waiting | `ConcurrentLeasesSerializeTheCriticalSection` plus 20 others, including every `DurableFile` concurrency test |

That last row is the useful one: `ConcurrentLeasesSerializeTheCriticalSection` runs 8 workers × 500 deliberately non-atomic increments and only passes if the lease actually provides mutual exclusion.

38/38 locally under the dotnet harness (`DurableFile` + `SemaphoreLease` together).

wallstop:
## Green, and the parallelism claim settled with a full run

Unity Tests [`31147776574`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31147776574) on `ad2c5b47`: **all 18 jobs success**, `Unity CI Success` included.

### Wall clock

| | Baseline `31130259492` | This run |
| --- | ---: | ---: |
| Licensed legs | 16 | 14 |
| Sum of leg durations | 70.9 min | 68.5 min |
| Packing floor (2 runners) | 35.5 min | 34.3 min |
| **Wall clock** | **88.0 min** | **56.1 min** |
| Packing efficiency | 40% | 61% |

**36% off the wall clock for comparable work.** The mechanism shows up directly in the leg-to-leg handoffs on a single runner:

```text
ELI-MACHINE   2.3 ->  7.0   Unity 2021.3.45f1 editmode
ELI-MACHINE   7.1 ->  9.6   Unity 2021.3.45f1 playmode              gap 0.03 min
ELI-MACHINE   9.7 -> 13.8   Unity 6000.3.16f1 editmode              gap 0.03 min
ELI-MACHINE  13.8 -> 18.3   Unity 6000.5.2f1 editmode               gap 0.02 min
ELI-MACHINE  18.3 -> 22.3   Unity 2022.3.45f1 editmode              gap 0.02 min
ELI-MACHINE  40.6 -> 46.5   standalone 2022.3.45f1                  gap 0.03 min
ELI-MACHINE  49.2 -> 53.3   6000.3.16f1 editmode (SINGLE_THREADED)  gap 0.03 min
```

Seconds, where the baseline's equivalent handoff was **13.4 minutes**. The cap was the cause, and the commit stays.

### Two gaps that remain, stated rather than glossed

- **12.6 min** on ELI between the fast tier ending and standalone starting. That is the `needs:` barrier — the remaining half of #326, now the single largest line item.
- **10.9 min** on DAD inside the fast tier (finished at +6.7, next leg at +17.6). I have not explained this one. The likeliest candidate is the superseded run for `247d6d1c` still draining its head-guarded legs through the same runner group during that window, but I did not confirm it, so I am not claiming it.

Both are worth a follow-up read of the next few runs rather than a fix guessed at now.

## Verification that the new tests ran

`pass=0 fail=0` is the false green to fear with new fixtures. From the 2021.3 PlayMode artifact — **`total=8738 passed=8718 failed=0 skipped=4`**:

- `SemaphoreLeaseTests`: **12 of 12 passed**, including `ConcurrentLeasesSerializeTheCriticalSection` and `DisposingTwiceReturnsOnlyOnePermit`.
- `DurableFileTests`: **25 of 25 passed**, including `ConcurrentAppendsInterleaveWholeRecords` — which is the case that fails without the per-path gate, and it passes on Windows, confirming `FileMode.Append` is non-atomic there too and not only on the Linux harness where it was found.

## One check outside this branch

`copilot-pull-request-reviewer` reports failure with *"the user who requested the review has reached their quota limit."* That is an account quota, not anything here; `mergeStateStatus` is `CLEAN`. Everything else: 51 success, 7 skipped, 1 neutral.

Runtime/Core/Threading/SemaphoreLease.cs
```diff
@@ -0,0 +1,190 @@
...
+        public void Dispose()
+        {
+            SemaphoreSlim semaphore = _semaphore;
+            if (semaphore == null)
+            {
+                return;
+            }
+
+            _semaphore = null;
+            semaphore.Release();
```
wallstop: I think, here and elsewhere, we should try/catch/swallow things like ObjectDisposed exceptions

wallstop: Agreed, and it is worse than it first looks. Fixed in `0bfecb61`.

`Dispose` runs from a `finally`, so a throw there does not *add* information — it **replaces** the exception the caller was already unwinding with. Disposing a semaphore while a lease is outstanding turned the caller's real failure into a confusing `ObjectDisposedException` about teardown.

```csharp
_semaphore = null;
try
{
    semaphore.Release();
}
catch
{
    // ObjectDisposedException: the semaphore is gone, so the permit is moot.
    // SemaphoreFullException: the count is already back at its maximum.
}
```

Both reachable exceptions clear the bar for swallowing: nothing is actionable, **and** the permit is accounted for either way. Where that second half does not hold — a resource genuinely leaked — the rule I wrote down says log it rather than hide it, but still do not throw.

Three tests, each verified red against a mutation restoring the throwing `Dispose`. The one that captures the actual defect asserts on the message of the exception that escapes:

```csharp
InvalidOperationException thrown = Assert.Throws<InvalidOperationException>(() =>
{
    using (semaphore.Acquire())
    {
        semaphore.Dispose();
        throw new InvalidOperationException("the real failure");
    }
});
Assert.AreEqual("the real failure", thrown.Message);   // was ObjectDisposedException before
```

The third pins the other branch: copying a lease is documented as forbidden, and doing it anyway now degrades to a no-op instead of raising the count *above* the semaphore maximum.

## "and elsewhere" — swept, and the convention already existed

Every other place in `Runtime/` that touches a possibly-disposed object already does this. Mine was the only gap:

| Site | Status |
| --- | --- |
| `SingleThreadedThreadPool.Signal()` → `_workAvailable.Release()` | already `try`/`catch`/swallow |
| `SingleThreadedThreadPool.DoWorkAsync` → `_workAvailable.WaitAsync` | already covered by an outer `catch (ObjectDisposedException)` |
| `SingleThreadedThreadPool.DisposeAsync` | already swallows |
| `DurableFile.TryDelete` | already swallows |
| `SemaphoreLease.Dispose` | **was the gap** |

So this was folklore that the new code did not inherit. It is now a rule in [forbidden-patterns § Dispose Never Throws](https://github.com/Ambiguous-Interactive/unity-helpers/blob/dev/wallstop/durable-file-and-ci-throughput/.llm/references/forbidden-patterns.md#dispose-never-throws) with a checklist item in `defensive-programming`.

## One related finding I did not change

`PooledArray.Dispose` and `PooledResource.Dispose` in `Buffers.cs` invoke a **consumer-supplied** `_onDispose` delegate with no `try`/`catch`. Same masking hazard — a throwing `onRelease` callback replaces the caller's exception — across ~113 files of call sites.

I stopped short of changing it, because it is a different judgement call: swallowing a known-benign infrastructure exception is not the same as silently swallowing an arbitrary consumer callback, and doing that quietly on a hot pooling path would hide their bugs. Today the pool leaks the item **and** masks the exception; swallowing would fix the masking and keep the leak; swallow-and-log would fix both but adds logging to a hot path.

Happy to take any of the three — just say which. Not doing it blind in a PR about durable writes and CI plumbing.

41/41 locally; Unity compile gate `FRESH=39 STALE=0 MISSING=0`.

Runtime/Core/Threading/SemaphoreLease.cs
```diff
@@ -0,0 +1,210 @@
...
+    /// sounds: an extra <see cref="SemaphoreSlim.Release()"/> raises the permit count above what the
+    /// semaphore was constructed with and quietly lets two callers into a section built for one.
+    /// <see cref="IsHeld"/> reports whether this lease still owns a permit.
+    /// </para>
+    /// <para>
+    /// <b>Do not copy a lease.</b> Assigning one to another variable, capturing it, or passing it by
+    /// value produces a second lease pointing at the same permit. Disposing both is harmless — the
+    /// second release is swallowed rather than allowed to corrupt the count — but the section is
+    /// unguarded from the first disposal onward. Acquire it directly into a <c>using</c> and let it
+    /// die there.
```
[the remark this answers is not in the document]
wallstop: **Correct, and the guarantee I wrote was simply false.** Fixed in `4ff2a2d2`. Measured before believing it:

```text
new SemaphoreSlim(1, 1): CurrentCount=1   <- extra Release throws, Dispose swallows
new SemaphoreSlim(1)   : CurrentCount=2   <- extra Release SUCCEEDS, silently
```

`DisposingACopiedLeaseCannotCorruptThePermitCount` used the bounded constructor, so it certified the claim by exercising the one configuration where it happened to hold. Replaced with a pair that pins both halves, the second one documenting the corruption rather than a guarantee:

```csharp
[Test]
public void DisposingACopiedLeaseInflatesACountWithNoExplicitMaximum()
{
    using SemaphoreSlim semaphore = new(1);   // maximum is int.MaxValue
    SemaphoreLease lease = semaphore.Acquire();
    SemaphoreLease copy = lease;

    lease.Dispose();
    copy.Dispose();

    Assert.AreEqual(2, semaphore.CurrentCount);
}
```

**The deeper point, now stated in the docs rather than implied away:** per-lease disposal tracking *cannot* make copying safe. A copy carries its own flags and cannot observe that the original already released — so no amount of state on the struct closes this. Only shared state would, which means an allocation per acquire. The docs now say that, and recommend the two-argument constructor as damage control: it does not make copying safe, it decides whether the mistake is a swallowed no-op or silent corruption.

**Also taking your earlier finding** ("Missing sources create destination directories"), which was still open. `TryCopy`/`CopyAsync` called `EnsureDirectory` before opening the source, so a copy from a missing source created the destination tree and then failed — a no-op operation leaving a mutation behind. Source opens first now, and `CopyFromAMissingSourceCreatesNothing` is verified red against the old ordering.

43/43 locally; Unity compile gate `FRESH=39 STALE=0 MISSING=0`.

Issue #354, opened 2026-08-07
serialization: sets of collections now diverge from dictionaries — fix the guidance and pin why


wallstop:
Follow-up from #348. Dictionaries with collection values now serialize automatically; sets do not,
and that divergence is deliberate but currently under-explained in the product.

## Why sets were not given the same treatment

The same boxing would work mechanically — `SerializableSetBase<T, TSet>` declares `T[] _items` and
has the identical defect. It was left alone because a set of collections is almost always a modelling
mistake rather than a serialization one:

`List<T>` has **reference** equality. A `SerializableHashSet<List<float>>` treats two lists with
identical contents as two distinct elements, and a set restored from disk is a set of *new* list
instances — so `Contains(someList)` is false for every list the consumer holds. Making it serialize
would turn a shape that fails loudly into one that fails silently and much later.

So the set drawer keeps reporting the shape as unsupported, and that is the right guidance.

## What is actually left to do

1. **The set diagnostic still recommends `SerializableList<T>`**, which has the same reference
   equality and therefore the same problem. The message should say that the element type is the thing
   to reconsider, and mention identity semantics explicitly for the case where the consumer really
   does want them.
2. **Nothing tests the equality consequence.** A test that saves a `SerializableHashSet<SerializableList<float>>`,
   restores it, and asserts `Contains` fails for an equal-content list would pin *why* this is refused,
   so a future session does not "fix" it by adding boxing.
3. **`SerializableSortedSet<List<T>>` has a second problem** the dictionary work did not touch:
   `List<T>` is not `IComparable<List<T>>`, so the sorted set cannot order its elements at all. Worth
   confirming what it currently does — a runtime throw would be better reported at author time.

## Not in scope

Adding boxing to `SerializableSetBase`. If that is ever revisited, read the equality argument above
first — the missing serialization is a symptom, not the defect.

wallstop:
All three items are done.

**1. The diagnostic no longer recommends `SerializableList<T>` as the fix.** Shipped in [#362](https://github.com/Ambiguous-Interactive/unity-helpers/pull/362):

> The element type is the thing to reconsider: a list compares by reference, so a set of lists already treats equal contents as distinct elements and a restored set contains none of the instances your code holds. Prefer a `[Serializable]` class element with value equality; use `SerializableList<T>` only when you genuinely want identity semantics.

[...]

Issue #353, opened 2026-08-07
ci: a self-hosted runner sits idle 13.9 min between legs with legs queued (measured)


wallstop:
Now that `max-parallel` is gone (#352) and the tier barrier is measured at **1 second**, the only
remaining CI wall-clock cost is a self-hosted runner sitting idle *between* legs while legs are
queued. This issue records the measurement so the next person does not re-derive it.

## Measurement

Unity Tests run [`31221618477`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31221618477),
`main` at `766b7c26` — the first protected-branch run with no eligibility cap.

Fast tier: 8 legs, 2 runners.

| Leg | Start | End | Duration |
| --- | --- | --- | ---: |
| 2022.3 editmode | 21:51:17 | 21:56:07 | 4.8 min |
| 6000.5 editmode | 21:56:09 | 22:00:52 | 4.7 min |
| 6000.3 playmode | 21:57:26 | 22:01:40 | 4.2 min |
| 2022.3 playmode | 22:07:39 | 22:10:45 | 3.1 min |
| 2021.3 editmode | 22:10:46 | 22:16:11 | 5.4 min |
| 6000.5 playmode | 22:14:49 | 22:18:49 | 4.0 min |
| 2021.3 playmode | 22:16:13 | 22:19:02 | 2.8 min |
| 6000.3 editmode | 22:19:04 | 22:23:20 | 4.3 min |

- **Sum of leg durations: 33.3 min.** Two runners → a **16.7 min** packing floor.
- **Actual tier wall clock: 32.1 min** (21:51:17 → 22:23:20).
- Six of the seven leg-to-leg handoffs were **0.0 min**, which is `max-parallel` removal working.

Two were not:

| Runner | Idle from | To | Cost |
| --- | --- | --- | ---: |
| A | 22:00:52 | 22:14:49 | **13.9 min** |
| B | 22:01:40 | 22:07:39 | **6.0 min** |

Both runners were idle **simultaneously** from 22:01:40 to 22:07:39 — six minutes with four legs
still queued and nothing running.

## What it is not

- **Not the eligibility cap.** It is gone, and six handoffs at 0.0 min prove dispatch can be instant.
- **Not the organization Unity lock.** `Acquire organization Unity lock` took **2-4 s on all eight
  legs** (3, 2, 3, 4, 2, 3, 4, 3). This is the second independent run to show that, and it is also
  why #337's premise does not hold.
[...]

Issue #337, opened 2026-08-05
Optimization: Parallel Unity CI


wallstop:
Research if it is possible, or worth it, to run the various Unity test suites/CI that are already serialized on one runner into parallel operations. Since they are all on one host, we shouldn't have licensing issues, and since Unity is single threaded, theoretically this should be fine. The dangers I can think about are them sharing a checkout/source/asset directory, as tests may dirty it.

wallstop:
**This issue's premise needs re-checking before any work goes into it.**

The issue opens on "The org build lock serializes every leg onto one Unity seat", which `PLAN.md` also records as a verified hard constraint. Measured on Unity Tests run [`31130259492`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31130259492) (main, green, 2026-08-06), the `Acquire organization Unity lock` step took:

```text
   4s  Unity 2022.3.45f1 playmode          2s  Unity 2021.3.45f1 editmode
   2s  Unity 6000.5.2f1 editmode           2s  Unity 2022.3.45f1 editmode
   4s  Unity 2021.3.45f1 playmode          2s  Unity 6000.3.16f1 playmode
   3s  Unity 6000.3.16f1 editmode          4s  Unity 6000.5.2f1 playmode
   3s  standalone 6000.5.2f1               4s  standalone 2022.3.45f1
   2s  standalone 2021.3.45f1              3s  standalone 6000.3.16f1
   3s  6000.3.16f1 editmode (ST)           4s  6000.3.16f1 playmode (ST)
   2s  Unity package export smoke
```

Fifteen of fifteen licensed legs, 2–4 seconds each. Nothing waited on the lock in that run; the two runners really were working legs concurrently.

So "the lock serializes everything" was not the reason CI was slow. The actual cost was `max-parallel: 2`, which gates matrix *eligibility* rather than execution — 16 legs, 70.9 min of work, 2 runners, a 35.5 min packing floor against an 88 min wall clock, with a runner idle 13.4 min while four legs sat queued. That is removed in [#352](https://github.com/Ambiguous-Interactive/unity-helpers/pull/352), and the first legs of its own run hand off with a **0.0 min** gap.

What that means here: if legs already run concurrently, the win this issue is reaching for may largely not exist. Worth re-measuring the acquire step before investing — the command is one API call:

```bash
gh api "repos/$R/actions/runs/$RID/jobs?per_page=100" \
  -q '.jobs[] | . as $j | .steps[]
[...]

PR #355, opened 2026-08-07
Make collection-valued dictionaries serialize, and stop a release contract from lying

### LINKED ISSUE #348: serialization: make SerializableDictionary<TKey, List<TValue>> just work, with no consumer type change
Follow-up from review on [#344](https://github.com/Ambiguous-Interactive/unity-helpers/pull/344): *"Why can't consumers that use our sets/dictionaries with this type have it just work automagically? I want simplest-possible consumer code."*

It is achievable **without breaking a single existing asset**. Writing the design down rather than shipping it blind, because it changes how user data is serialized and there is currently no Unity gate available to validate it (#333, #347, plus a GitHub Actions outage).

## Why it fails today

`SerializableDictionaryBase<TKey, TValue, TValueCache>` declares `[SerializeField] TValueCache[] _values`. In the two-argument form `TValueCache = TValue`, so `SerializableDictionary<string, List<float>>` makes that field a `List<float>[]` — an array whose element type is itself a collection, which Unity refuses. The parallel `TKey[]` survives, which is what makes it look like a data problem instead of a type problem.

## The fix: choose the backing field per value type

Add a second serialized field to the base:

```csharp
[SerializeField] protected internal TValueCache[] _values;                     // existing
[SerializeField] private SerializableDictionary.Cache<TValueCache>[] _boxed;   // new
```

`Cache<T>` is a `[Serializable]` class with a `T Data` field, so `Cache<List<float>>[]` is an array of a serializable **class** — the nesting Unity accepts. It serializes for every value type, collections included.

Pick between them with a rule that is exact rather than heuristic:

```csharp
private static readonly bool RequiresBoxing =
    typeof(TValueCache).IsArray
    || (typeof(TValueCache).IsGenericType
        && typeof(TValueCache).GetGenericTypeDefinition() == typeof(List<>));
```

That is the complete set of cases where `TValueCache[]` is dropped **and** boxing repairs it. Other unserializable value types (interfaces, `Dictionary<,>`, non-`[Serializable]` classes) are not fixed by boxing either — `Cache<Dictionary<,>>.Data` is just as unserializable — so they must keep reporting the existing Inspector error, not silently switch fields.

`OnBeforeSerialize` writes whichever field the flag selects; `OnAfterDeserialize` reads whichever is populated.

## Why this is backwards compatible

- **Ordinary value types keep using `_values`.** Byte-identical YAML. No migration, no diff churn, no downgrade hazard.
- **Collection value types previously stored nothing at all.** There is no old data to migrate — the whole point of #314 is that those values were being dropped.

That property is the reason to prefer per-type selection over "always box", which would rewrite every dictionary in every consumer project on next save and make new assets unreadable by older package versions.

## Cost, honestly

1. `SerializableDictionaryPropertyDrawer` (~10k lines) keys off `_values` throughout — rows, pagination, duplicate/null-key detection, height caching. It needs to resolve either field.
2. `HasDroppedValuesArray` must stop firing for the now-working shape while still firing for genuinely unserializable value types.
3. The same treatment for `SerializableSortedDictionary` and `SerializableSetBase` (`T[] _items`).
4. Every dictionary gains one extra serialized field, an empty array when unused. Small, but it is a format addition across every consumer asset.
5. It cannot be merged without a real Unity run. This is exactly the class of change that ships data loss when unvalidated.

## Correction worth recording

#314 states the cache escape hatch *"needs a concrete `Cache<T>` subclass for every value type"*, and I repeated that in #344's docs and commit message. **It is not true.** The open generic satisfies the constraint directly — verified by compiling the hierarchy's generic shape:

```csharp
new Dict<string, List<float>, SD.Cache<List<float>>>();   // compiles; no subclass needed
```

So `SerializableDictionary<string, List<float>, SerializableDictionary.Cache<List<float>>>` already works today. Verbose, but no per-type boilerplate. The docs should say so regardless of whether this issue is implemented.

## Ergonomic ladder

| Form | Consumer cost |
| --- | --- |
| `SerializableDictionary<string, List<float>, SerializableDictionary.Cache<List<float>>>` | works today, verbose |
| `SerializableDictionary<string, SerializableList<float>>` | shipped in #344; short, and the value is a usable list |
| `SerializableDictionary<string, List<float>>` | this issue |

I would take this on once a Unity gate exists again.

wallstop:
## Summary

- `SerializableDictionary<TKey, List<TValue>>` and its sorted counterpart now serialize, with no consumer type change (#348)
- Dependency bumps from #334 folded in here so real Unity legs validate them — which immediately caught that the build-lock v1.11.0 bump reddens every Unity leg, so those six pins are held and the rest ship (#356)
- A release contract that restates a pinned SHA is fixed to derive it, five ordering assertions that passed vacuously are hardened, and the suite that caught all of this now runs in CI
- #326 closed by measurement rather than by a change; the remaining CI cost is measured and filed as #353

Closes #348. Supersedes #334 (partially — see below). Blocks the build-lock bump behind #356.

## `SerializableDictionary<TKey, List<TValue>>` just works (#348)

#344 shipped the detection; this ships the remedy.

Unity refuses an array whose element type is itself a collection, so the values array of a
list-valued dictionary was a `List<TValue>[]` that never reached the asset. The keys array survived,
because it is a plain `TKey[]` — which is what made an asset look authored while every lookup came
back empty.

Values are now mirrored into a parallel array of one-field boxes at `OnBeforeSerialize` and unpacked
at `OnAfterDeserialize`, so the collection is a field of a class rather than an array element. Nothing
between those two boundaries changed: duplicate detection, order-preserving sync and `GetValue` still
operate on the original array. Both dictionary bases are fixed, because both declare the same defect.

**The boxed array is populated only for the two shapes Unity drops and boxing repairs** — `T[]` and
`List<T>`. Every other value type keeps storing values exactly where it did, so no existing asset is
rewritten and new assets stay readable by older package versions.

**One honest cost.** The design intended byte-identical YAML for ordinary dictionaries. Measured in a
real editor, that does not hold: a `[SerializeField]` cannot be declared conditionally, so Unity
declares the boxed array on *every* dictionary. `RequiresBoxedValues` controls whether it is
populated, not whether it exists. The result is one empty array per dictionary in the serialized form
— a line added on first save, no data moved. The docs and a test now state that instead of the
stronger claim.

**Sets are deliberately left reporting the shape as unsupported.** `List<T>` has reference equality,
so a `SerializableHashSet<List<T>>` treats two equal-content lists as distinct elements and a restored
set never matches the lists a consumer holds. Making it serialize would convert a loud failure into a
silent one. Follow-up in #354.

**Recursive nesting composes, because the two mechanisms are disjoint.** Review asked what happens to
`SerializableDictionary<string, SerializableDictionary<int, ...>>`. The boxing selects on a *raw*
collection (`IsArray || List<>`); a serializable dictionary is a `[Serializable]` **class**, which is
the shape Unity has always accepted as an array element. So dictionary-in-dictionary never needed
boxing and still does not, while `List<Dictionary<..>>` boxes the outer list and nests its elements
normally inside the box. Measured, then pinned with four round-trip tests including
three-dictionaries-deep:

```text
nestedDictionaryValues  keys=True values=True  boxed=empty
listOfDictionaryValues  keys=True values=False boxed=True
deeplyNestedValues      keys=True values=True  boxed=empty
dictionaryItems         items=True     (a set OF dictionaries works -- the element is a class)
```

The remaining bound is **Unity's own serialization depth limit**, which no wrapper lifts; the docs say
so and point at flattening to a composite key rather than implying arbitrary recursion is supported.

**The diagnostic keeps a job.** Boxing does not repair `Cache<List<ISomething>>`, whose `Data` field
Unity still refuses, so the drawer inspects a real element rather than trusting that the array exists.

## Dependencies (#334)

#352's credential-detection fix was already correct — #334 was simply stale, and after a rebase it is
42 green / 0 failed. It is **not** merged directly, because every Unity leg on a Dependabot run is
skipped by design, and this bump moves the organization build lock from v1.9.1 to v1.11.0 — the one
dependency whose behavior only a real Unity leg exercises. Cherry-picked here (authorship preserved)
so the full matrix validates it before it reaches `main`.

**That decision immediately paid for itself.** `Unity 2021.3.45f1 editmode` and
`Unity 2022.3.45f1 playmode` both failed `Require confirmed Unity cleanup` while their **tests
passed** (`4199 passed / 0 failed` and `8723 passed / 0 failed`; `Run Unity Test Runner` and
`Verify tests actually ran` green on both). Unity batchmode crashes on shutdown *after* writing a
valid results file — a condition the workflow already absorbs — and that truncates the license return
log. On `main` at v1.9.1 the identical warnings appear on the identical leg and it passes; at v1.11.0
the same `cleanup-reason=return-log-truncated` quarantines the seat and fails closed.

So the six `ambiguous-organization-build-lock` pins are held at the SHAs `main` is green on, and every
third-party bump in the group ships: `actions/checkout`, `actions/setup-node`, `actions/setup-dotnet`,
`actions/setup-python`, `docker/login-action`, `lycheeverse/lychee-action`,
`release-drafter/release-drafter`. Filed as #356 with the evidence and three ways to unblock it.

Merging #334 on its 42 green hosted checks would have reddened every Unity leg on `main`.

## The contract the bump broke, and the two holes that hid it

`release publish tag preparation sets up Node before npm publication checks` **restates** the
`actions/setup-node` SHA, so bumping the pin desynchronizes the copy. This is the same defect #351
removed from the build-lock pins, still present here. The assertion now derives the pin from
`release.yml` and requires every `actions/setup-node` in the file to agree, which also catches a
partial bump the old form could not see.

Checking that the fix held surfaced two more problems:

**Five ordering assertions never worked.** `IndexOf(a) -lt IndexOf(b)` passes vacuously when `a` is
absent, because `IndexOf` returns `-1` and `-1` is below every real index — so each contract stopped
checking the thing it was written for at exactly the moment its named step was renamed or deleted.
Proven by renaming prepare-tag's `Setup Node.js` step: the suite stayed **132/132 green**. Replaced
with a `Test-AppearsBefore` helper requiring both to be present; the same mutation is now caught. The
equivalent assertion in `test-unity-workflow-matrix-contract.ps1` already guards its index and is
unchanged.

**Nothing in CI ran this suite.** `test-sync-script-contracts.ps1` is reachable only through
`validate:tests` → `validate:prepush`, a local pre-push hook. No workflow invoked it. That is why
#334 reported 42 green checks while carrying the break. Added to `test-lint.yml` beside its siblings.

## CI throughput: #326 closed by measurement, not by a change

PLAN.md asked for the next run to be read before touching the tier barrier. Read — Unity Tests run
`31221618477`, `main` at `766b7c26`, the first protected-branch run with no eligibility cap:

| Transition | Before (`31130259492`) | Now |
| --- | ---: | ---: |
| fast → standalone | 10.1 min | **1 second** |

Removing `max-parallel` in #352 absorbed it, because a fresh self-hosted dispatch was the whole
expense. Retiring the `unity-tests-single-threaded` → `unity-tests-standalone` edge would therefore
buy at most the 2.3 min that same mechanism already collapsed, while giving up a contract that keeps
two same-workflow jobs off the one Unity seat. **The contracts stay, now for a measured reason
instead of a disproved one.**

The org lock is confirmed not to be a constraint for a second independent run: `Acquire organization
Unity lock` took 2–4 s on all eight legs, which also settles #337's premise.

What the measurement found instead: 33.3 min of leg work on two runners is a 16.7 min floor, and the
fast tier took 32.1 min. Six of seven handoffs were 0.0 min; two were **13.9 min** and **6.0 min**,
with both runners idle together for six of those minutes and four legs queued. Not eligibility, not
the lock, not the barrier. Filed with the full timeline as #353.

## Verification

Real Unity editor (6000.4.6f1), `TestRunnerApi`, targeted EditMode runs:

| Filter | Result |
| --- | --- |
| `NestedCollectionSerializationTests` | **29 / 29** |
| `SerializableDictionaryPropertyDrawerTests` | **156 / 156** |
| `.*Serializable(Dictionary\|SortedDictionary).*` | **322 / 322** |

Mutation tests, because a test that cannot fail is not evidence:

- Disabling `RequiresBoxedValues` fails exactly the two tests that pin the behavior
  (`ACollectionValuedDictionaryRoundTripsThroughUnitySerialization`, expected 2 got 0, and
  `ACollectionValueTypeKeepsItsValuesInTheBoxedArray`, expected 1 got 0) and nothing else. The sorted
  test stays green under that mutation, which is correct — the sorted base carries its own copy.
- Desynchronizing one `actions/setup-node` pin fails the release contract; renaming prepare-tag's
  `Setup Node.js` step fails it only after the `Test-AppearsBefore` fix.
- Restoring rehydration in `EditorAfterDeserialize` fails the new
  `EditorAfterDeserializeDoesNotOverwriteAFreshlyWrittenValuesArray` with `Expected: 42.0f, But was:
  1.0f` — the stale boxed value winning over the Inspector's write — and nothing else.

`npm run validate:prepush` — 132/132 contract tests, spelling, EOL, formatting, all green.

## Note for reviewers

The Unity MCP gate was itself broken for most of this work and reported green while lying: a stale
`Assets/Editor/UnityHelpersMcpTestRunner.cs` in the host project failed to compile, so the editor
never reloaded its domain and every probe read the previous domain's types — with zero console
errors, `isCompiling=False`, and a correctly-updated DLL on disk. An early probe under that condition
reported that the documented open-generic cache box does **not** serialize. That was wrong. Re-measured
after the reload: it serializes and round-trips, and it now has the test it shipped without.

---

> [!NOTE]
> **Medium Risk**
> Serialization and editor drawer paths change how dictionary assets round-trip (mitigated by extensive tests); broad workflow pin updates affect every CI job; release contract changes catch regressions but depend on workflow structure staying aligned.
> 
> **Overview**
> **`SerializableDictionary` / `SerializableSortedDictionary`** now persist **`List<T>` and `T[]` values** without changing consumer types: values mirror into a **`_boxedValues`** array Unity accepts, rehydrate on load, and the property drawer reads either `_values` or boxed `Data`. **`EditorAfterDeserializeFromManagedArrays`** avoids overwriting Inspector writes; ordinary value types still use **`_values`** only (one empty boxed array line per dictionary in YAML). Sets of raw lists still error; nested **`List<List<T>>`** is not boxed. Docs, CHANGELOG, diagnostics text, and **`NestedCollectionSerializationTests`** reflect the new behavior.
> 
> **CI / tooling:** Workflow **`uses:`** pins move forward (notably **`actions/checkout`**, **`actions/setup-node`**, **`actions/setup-dotnet`**, **`actions/setup-python`**, Docker login, lychee, release-drafter). **`test-sync-script-contracts.ps1`** runs in **`test-lint.yml`**; ordering checks use **`Test-AppearsBefore`** so missing steps cannot pass vacuously; the release workflow’s **`setup-node`** SHA is **derived** from the file instead of duplicated.
> 
> **Self-hosted runners:** **`print-self-hosted-runner-diagnostics`** logs **free physical/commit memory** and **resident Unity-related processes** before jobs start, to surface OOM risk and stray editors from prior runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bd10c5ff6b4cc415dc8c6c359e77a16a966ef21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wallstop:
## CI found the thing this PR existed to find

`Unity 2021.3.45f1 editmode` and `Unity 2022.3.45f1 playmode` both failed `Require confirmed Unity cleanup` — while **their tests passed**:

| Leg | Test result | Failing step |
| --- | --- | --- |
| 2021.3 editmode | `total=4661 passed=4199 failed=0 skipped=443` | `Require confirmed Unity cleanup` |
| 2022.3 playmode | `total=8743 passed=8723 failed=0 skipped=4` | `Require confirmed Unity cleanup` |

`Run Unity Test Runner` and `Verify tests actually ran` succeeded on both. Nothing in the serialization change is implicated.

**Cause: the build-lock bump, not this branch.** Unity batchmode crashes on shutdown *after* writing a valid results file — a condition the workflow already recognizes ("honoring the artifact as the source of truth … benign post-work shutdown crash") — and that crash truncates the license return log. On `main` at v1.9.1 the identical pair of warnings appears on the same leg (run `31221618477`, green) and the job passes. At v1.11.0 the same `cleanup-reason=return-log-truncated` quarantines the seat and fails closed.

**Resolution:** the six `ambiguous-organization-build-lock` pins go back to the SHAs `main` is green on; every third-party bump in the group is kept (checkout, setup-node, setup-dotnet, setup-python, docker/login-action, lychee, release-drafter). Filed as #356 with the full evidence and the three ways it could be unblocked.

This is exactly why the bumps were brought here instead of merged from #334, whose Unity legs are all skipped for want of organization secrets — it reported 42 green checks while carrying a change that reddens every Unity leg on `main`.

## Also in this push

`EditorAfterDeserialize` no longer rehydrates from the boxed array. Its callers have just written the managed values array from the Inspector's SerializedProperties; rehydrating overwrote that with a copy stale until the next serialize, so an edit would silently revert. The drawer already guards this for ScriptableSingleton targets ("would read stale data and overwrite our changes") and the boxed array reopened it. Rehydration is now explicit per call site: on for Unity's callback, off for the editor path and ProtoBuf. Mutation-verified — restoring it fails the new test with `Expected: 42.0f, But was: 1.0f` and nothing else.

wallstop:
## Bugbot review addressed — all three findings were real

**1. Inspector edits lost for list values (High).** Correct, and it was the *inverse* of the bug the previous commit fixed. The editor rebuild is two different operations that shared one method:

- Most drawer callers reach `EditorAfterDeserialize` **without** having written the managed values array — for a collection value Unity only restored the boxed one — so that path **must** refill it, or the runtime map rebuilds from a stale array and the next serialize writes that staleness back over good data.
- Exactly one caller (the managed-array sync near the end of the drawer) has just filled that array itself and needs the refill **skipped**, or its write is overwritten.

Turning the refill off for both fixed the second and broke the first. Now split into `EditorAfterDeserialize` (refills) and `EditorAfterDeserializeFromManagedArrays` (does not), with a test pinning each direction.

**2. Nested collections fail silently (Medium).** Correct. Boxing repairs exactly **one** level: Unity refuses `List<List<T>>`, `List<T[]>` and jagged arrays as a class field just as firmly as as an array element, so the box's own `Data` is dropped and boxing stores nothing. Measured — a `Cache<List<List<float>>>` element's `Data` does not resolve. The predicate now requires the collection's **element** to be a shape Unity accepts, so those types are not boxed at all.

The **diagnostic** half of that finding is *not* fixed here, and I filed #357 rather than paper over it. Unity declares the boxed field on every dictionary, so for an unrepairable shape the drawer resolves an empty boxed array and reports nothing. Every cheap discriminator fails: an empty *supported* dictionary looks identical, and the keys array is not the tell either — measured, `OnAfterDeserialize` nulls `_keys` whenever keys and values disagree, so after one round trip there is no asymmetry left. The real fix is to expose `RequiresBoxedValues` to the drawer; #357 carries the analysis and the decision table.

**3. Stale boxes overwrite valid values (Medium).** `OnProtoAfterDeserialization` now passes `rehydrate=false` — protobuf populates the values array directly and never writes the boxed one (`[ProtoIgnore]`). The JSON converters were already safe: both construct a fresh instance, so the boxed array is null there. Verified by reading both converter read paths rather than assuming.

Verification: fixture **31/31**, whole dictionary surface **322/322** on a real editor (6000.4.6f1).

wallstop:
Right on both counts, and the first one is the more serious of the two — thanks.

**The test could not fail.** It never populated the dictionary, so `MirrorValuesToBoxedCache` never ran, and an empty boxed array was the expected outcome whether `RequiresBoxedValues` was correct or not. It now sets a value and calls `OnBeforeSerialize` — the call that *would* fill the boxed array if the predicate regressed — and asserts on the **managed** arrays rather than SerializedProperties, because the Inspector's view of an unserializable shape is precisely what does not survive a round trip.

Mutation-verified, which is the only thing that makes the claim worth anything. Widening the predicate back to every `List<>`/array fails it with:

```text
Expected: null
But was:  <SerializableDictionary+Cache`1[List`1[List`1[Single]]]>
```

A box really was created, and nothing else in the fixture moves. The previous version of the test passed under that same mutation — exactly as you said it would.

**The name lied.** `…AndStillReportsTheFailure` kept a claim whose assertion I removed when the drawer half was split out to #357. Renamed to `AMultiLevelCollectionValueIsNotBoxed`, which is what it actually checks, and the comment no longer says the drawer reports anything.

Fixture 31/31 on a real editor.

PR #364, opened 2026-08-08
Dispose every struct-based scope exactly once, at zero allocation

### LINKED ISSUE #358: [Feature]: Correct Struct-Based IDisposable
### Feature Description

All of our struct-based IDisposables can be copied and then do bad things if double-disposed.

To fix, we should research solutions. My proposed solutions is to have a static (concurrent)set (keyed off of SINGLE_THREADED) that, when each struct is created, gets assigned an auto-incremented (u)long, and it is atomically added to the set. Then, in dispose, the disposal logic only runs if it is able to atomically remove the id from the set.

There may be more optimal patterns.

### Use Case / Motivation

struct based disposals can be scary

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Unity Version

None

### Package Version

None

### LINKED ISSUE #365: PooledArray<T> can still be returned twice, and the PooledResource remedy does not transfer as-is
Follow-up from #358 / PR #364, which fixed the same defect for `PooledResource<T>` and deliberately left this one open rather than half-doing it.

## The defect

`PooledArray<T>` tracks disposal in a `bool` field, so a **copy** of the lease — passing it to a method by value, assigning it, capturing it — carries its own copy of that flag and cannot see that the original already returned:

```csharp
public struct PooledArray<T> : IDisposable
{
    private bool _disposed;   // per copy, not per rent
    public void Dispose()
    {
        if (_disposed) { return; }
        _disposed = true;
        ...
    }
}
```

Dispose both copies and the array is returned twice. A pool holding one array twice hands it to two live callers, and from then on every write through one is visible through the other. `WallstopArrayPool<T>` clears on return, so the second return also wipes whatever the current renter had put there.

This is the same failure mode proven for `PooledResource<T>` in #358, where a test showed a second renter's write landing in the first renter's list.

## Why the `PooledResource<T>` remedy does not transfer as-is

`PooledResource<T>` got a per-instance `PoolLease<T>` carrying a generation that advances on every rent. It works there because the pool owns its instances for their whole life, so the lease can be stored in the free-list entry and travel pool → renter → pool at no allocation cost.

Two things block the same move here:

1. **`SystemArrayPool<T>` does not own its arrays.** It rents from `System.Buffers.ArrayPool<T>.Shared`, which hands back arbitrary arrays from its own buckets. There is nowhere to park a per-array lease. The options are a `ConditionalWeakTable<T[], PoolLease<T[]>>` (weak keys, no leak, but the lookup roughly doubles the cost of the cheapest array pool) or a recycled per-rent lease free list (the generation already defeats the ABA problem recycling would otherwise create). Neither is obviously right and both want measuring. This is also the most-used array pool — 62 call sites in-package against 8 for `WallstopArrayPool` and 4 for `WallstopFastArrayPool`.

2. **`Array.Empty<T>()` is a shared singleton.** Every pool returns it for a zero-length request:

   ```csharp
   case 0:
   {
       array = Array.Empty<T>();
       return new PooledArray<T>(array, 0, null);
   }
   ```

   A lease keyed per array instance would give *all* zero-length rents across the whole process one lease and one generation, so two concurrent zero-length rents would fight over it. Harmless today only because `Dispose` short-circuits on `array.Length == 0`. Any fix must keep zero-length leases lease-free, or give them a per-rent lease.

## The four storage shapes to convert

Each needs its stack/list to hold the lease rather than the bare array:

| Pool | `SINGLE_THREADED` | threaded |
| --- | --- | --- |
| `WallstopArrayPool<T>` | `Dictionary<int, List<T[]>>` | `ConcurrentDictionary<int, ConcurrentStack<T[]>>` |
| `WallstopFastArrayPool<T>` | `List<Stack<T[]>>` | `List<ConcurrentStack<T[]>>` |

## Suggested order

1. Reproduce with a failing test first, the way #358 did — two live rentals aliasing after a copy is disposed twice. Do it for each of the three pools, so the zero-length and BCL cases are visible up front.
2. Convert the two `Wallstop*ArrayPool` families (persistent lease in the bucket, zero allocation).
3. Decide `SystemArrayPool<T>` on measurement, not on preference: benchmark `ConditionalWeakTable` against a recycled lease free list against leaving it documented.
4. Only then remove the "do not copy this lease" warning from `PooledArray<T>`'s XML docs, which PR #364 added as the honest interim.

## Not in scope

`PooledResource<T>` — already fixed. Read `Runtime/Utils/PoolLease.cs` first; the reasoning about why a "is it already in the free list?" check is insufficient, and why a *rented*-instance set would trade the bug for a permanent leak, applies here unchanged.

### LINKED ISSUE #366: AssetDatabaseBatchScope: disposing one scope twice closes another live scope's batch, and the tests assert it
Found while sweeping struct-based `IDisposable`s for #358. Split out of PR #364 because the honest fix is a redesign of the helper's accounting, not the one-line flag I first tried — and because the repository's **own test suite already asserts the defect as if it were the contract**.

## The defect

`AssetDatabaseBatchHelper` counts live scopes in a single depth counter. Each `AssetDatabaseBatchScope` increments once on construction, and `Dispose` decrements once — every time it is called. So disposing one scope twice removes **two** scopes from the count while only one has ended, and the batch belonging to some other, still-live scope is closed early.

That is not hypothetical. It is what `Tests/Core/Editor/AssetDatabaseBatchScopeTests.cs` asserts today:

```csharp
[Test]
public void MixedDoubleDisposalAndNormalDisposal()
{
    AssetDatabaseBatchScope scope1 = AssetDatabaseBatchHelper.BeginBatch();
    AssetDatabaseBatchScope scope2 = AssetDatabaseBatchHelper.BeginBatch();
    AssetDatabaseBatchScope scope3 = AssetDatabaseBatchHelper.BeginBatch();
    // depth 3

    scope2.Dispose();
    scope2.Dispose();

    Assert.That(AssetDatabaseBatchHelper.CurrentBatchDepth, Is.EqualTo(1),
        "Depth should be 1 after double-disposing scope2");   // <-- two scopes' worth, one scope ended

    scope1.Dispose();   // depth 0 -> AllowAutoRefresh + StopAssetEditing fire here,
                        //            while scope3 is still live and still expects batching
    ...
}
```

`DoubleDisposeWithNestedScopesHandledCorrectly` does the same with a nested pair: double-disposing the **inner** scope drops the depth to 0 and runs the outermost cleanup while `outer` is still live. Its comment reads "protection against negative", which is true — the counter is clamped — but clamping at zero is not the same as keeping the count equal to the number of live scopes, and the assertion quietly accepts the second, wrong thing.

Consequence for a consumer: `AssetDatabase.StopAssetEditing()` and `AllowAutoRefresh()` run in the middle of an outer batch, so the rest of that batch's asset writes are unbatched — the exact performance loss the type exists to prevent — and an `AssetDatabase.Refresh()` requested by the still-live scope may never happen.

## Why the obvious fix is not enough

Adding a `private bool _disposed` guard to the struct fixes only the same-**variable** case. `AssetDatabaseBatchScope` is a struct, so a **copy** carries its own copy of that flag and disposing both still decrements twice — the same mechanism proven for `PooledResource<T>` in #358. It also forces the type to stop being a `readonly struct`, which `ScopeIsReadonlyStruct` pins deliberately.

Tried in PR #364 and reverted: it broke 7 tests, fixed half the problem, and changed a structural property of a public type — a poor trade to smuggle into an unrelated change.

## The fix that would work

Give each scope an identity the helper can account for, rather than an anonymous +1:

- `BeginBatch()` mints a small token (a class, so copies share it) and registers it.
- `Dispose` removes its token and decrements **only if the removal succeeded**, so every scope contributes exactly one decrement no matter how many copies are disposed or in what order.
- Out-of-order disposal keeps working unchanged: cleanup still runs for whichever disposal brings the live-token count to zero.

Editor-only and low frequency, so the per-scope allocation is irrelevant here — unlike the pooling case in #358, where it drove the design to a per-instance lease instead.

## Test work this requires

These four assert the current accounting and must be rewritten to the new contract, not deleted — the scenarios are worth keeping:

- `MixedDoubleDisposalAndNormalDisposal` — depth after double-disposing `scope2` becomes **2**, not 1.
- `DoubleDisposeWithNestedScopesHandledCorrectly` — depth after double-disposing `inner` stays **1**, not 0, and no mismatch warning fires.
- `DisposingSameScopeTwiceDoesNotCauseNegativeDepth` — keep the name, drop the expected warning; a second disposal becomes a silent no-op.
- `MultipleDispose.Count2` / `Count5` / `Count10` — same; extra disposals stop producing the mismatch warning.

And one asserts the implementation shape rather than a behavior:

- `ScopeIsReadonlyStruct` — decide deliberately. If the token approach is used the struct **can** stay `readonly` (the token is a readonly reference), in which case this test survives untouched, which is a point in that approach's favour.

Worth adding: a test that a **copy** of a scope disposed alongside the original leaves the depth correct. Nothing covers that today, which is why the gap survived.

wallstop:
## What

A pooled buffer could be handed to two callers at once. Every struct-based `IDisposable` in the package tracked disposal in one of its own fields, so a **copy** — passing it to a method by value, assigning it, capturing it — carried its own copy of that flag and could not see the original was already disposed. Disposing both ran the disposal twice.

Proven before it was fixed, against the real sources:

```text
Failed  TwoLiveRentalsDoNotAliasAfterADoubleReturn
  A second renter's write leaked into the first renter's list.
Failed  ACopiedLeaseReturnsTheSameInstanceToThePoolTwice
  Two live rentals resolved to the SAME list instance.
```

## How — a generational handle, held by value

`DisposalLease` is a `(slot, generation)` pair. The generation lives in a table **outside** the struct, so every copy reads the same one; it advances on each acquire, and only the copy still holding the current value may claim. This is the slotmap pattern — DOTS's `Entity{Index, Version}`, Rust's `slotmap`, Vulkan handles — which is the standard way to invalidate a stale handle without allocating one.

It catches the copy disposed immediately, the same lease disposed twice, **and** the case a "is it already back in the pool?" check cannot see: a stale copy disposed *after* the instance has been rented out again. At that moment the instance is not free, so a membership check says "fine" — and the release callback then clears a buffer its current renter is still using. That is the dangerous shape, and it has its own test.

## Zero allocation

This is a hot path, so the guarantee has to be free. Per cycle:

| path | `main` | this PR |
| --- | ---: | ---: |
| steady-state rent/return | 0 B | **0 B** |
| public `PooledResource(T, Action<T>)` | 0 B | **0 B** |
| `Get()` on a disposed pool | 32 B | **32 B** |

That last 32 B is not overhead — it is the object being handed out. A disposed pool has no pooled instance left, so it must produce one. Proven rather than asserted: the same path with a non-allocating producer measures **0 B**, and a bare `new List<int>()` measures exactly **32 B**.

Three choices make it free:

- **Slots recycle through a per-thread free list**, so a balanced acquire/release — what a `using` does — reuses the caller's own slot with no atomics and no shared head to contend on.
- **The generation table grows in fixed blocks that are never reallocated**, so a live generation is never moved out from under a concurrent claim.
- **Acquire advances the generation with a store-release, not an interlocked increment.** Sound because the acquirer is the slot's exclusive owner: a recycled slot reached the free list only through a winning claim that already advanced it past every outstanding lease, and it was pushed onto the free list of the thread now reading it. Worth ~21 ns/cycle — the difference between measurably slower and inside the noise.

Throughput, 2M cycles, 4 runs interleaved: `main` mean **210.8 ns**, this PR mean **206.4 ns**. Inside run-to-run noise.

## The sweep

`rg 'struct \w+.*: .*IDisposable'` over `Runtime/` and `Editor/`, all of it:

| Type | Before | Now |
| --- | --- | --- |
| `PooledResource<T>` | `bool _initialized` | lease |
| `PooledArray<T>` | `bool _disposed` | lease, all three array pools |
| `SemaphoreLease` | nulled field | lease, and now a `readonly struct` |
| `AssetDatabaseBatchScope` | nothing at all | lease, still a `readonly struct` |
| `AssetDatabasePauseScope` | `bool _disposed` | lease, and now a `readonly struct` |
| `Trie.Enumerator` ×2 | — | no change needed; holds `PooledResource`s |
| `SpatialHash2D`/`3D` `EntryBucket` | — | no change needed; holds a `PooledResource`, and is stored **by value** in a `Dictionary`, so it is copied by construction |
| `LabelWidthScope` ×2 | — | left alone; restoring the same saved label width twice is already idempotent |

Zero-length array rents deliberately take **no** lease: every pool hands back the same `Array.Empty<T>()` instance, so one shared slot would have every zero-length rent in the process fighting over one generation.

## Four tests asserted the defect as the contract

`MixedDoubleDisposalAndNormalDisposal` opened three scopes, disposed one twice, and asserted the depth fell 3 → **1** — two scopes off the count for one scope ending, which runs `StopAssetEditing()` while a third scope is still live and leaves the rest of its writes unbatched. Those four now assert the corrected accounting, and `DisposingACopyAndTheOriginalDecrementsOnce` covers the copy case nothing covered before.

## Evidence

- New fixtures `DisposalLeaseTests` (13 tests) and `PooledResourceDisposalTests` (14), green against the real sources compiled under `dotnet`.
- **Mutation:** making `TryClaim` always succeed — the pre-fix behavior — fails **9 of 17**. The survivors are the guard-rails and the allocation probes, which must pass either way.
- Zero-allocation pinned at a **zero-byte tolerance** in three tests.
- Unity compile gate green: 39/39 assemblies fresh, 0 console errors, with content assertions (`SemaphoreLease` is now `readonly`, `PoolLease` is gone) proving the domain actually reloaded rather than reporting a stale green.

Closes #358. Closes #365. Closes #366.

---

> [!NOTE]
> **Medium Risk**
> Touches hot-path pooling and editor asset batching with subtle concurrency semantics; behavior change is intentional but widespread across buffer pools and threading helpers.
> 
> **Overview**
> Fixes **double-dispose when struct `IDisposable` types are copied** by replacing per-copy `bool` guards with shared **`DisposalLease`** `(slot, generation)` handles backed by **`DisposalLeases`** (per-thread slot reuse, CAS on claim, zero alloc on hot paths).
> 
> **`PooledResource<T>`** and **`PooledArray<T>`** now return pooled buffers at most once per rent—including stale copies disposed after a re-rent. Zero-length array rents skip acquiring a lease.
> 
> **`SemaphoreLease`** is a **`readonly struct`**; copies share one permit release (fixes silent permit inflation when `SemaphoreSlim` has no explicit max). Docs now say copying is safe.
> 
> **`AssetDatabaseBatchScope`** / **`AssetDatabasePauseScope`** use leases so double dispose or copy+dispose cannot decrement batch depth twice or end another scope’s batch early.
> 
> Adds **`DisposalLeaseTests`** and **`PooledResourceDisposalTests`**; updates batch-scope and semaphore tests. **`SerializableSortedSet`** gains remarks on why `IComparable<T>` must stay.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f112fb585c06589a401ef8503b2ad7fb50a8da4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/Utils/AssetDatabaseBatchScope.cs
```diff
@@ -32,7 +32,7 @@ namespace WallstopStudios.UnityHelpers.Editor.Utils
     ///     Nested scopes are supported. The actual AssetDatabase calls are only made at the outermost scope.
     ///     </para>
     /// </remarks>
-    public readonly struct AssetDatabaseBatchScope : IDisposable
+    public struct AssetDatabaseBatchScope : IDisposable
```
wallstop: This approach still does not prevent copies from double disposing. If you copy this struct, you can trigger its dispose twice, once per copy. Here and potentially elsewhere. Please do a sweep and see if my "id" or "lease" concept needs to be applied EVERYWHERE that we do struct-based IDisposables.

Runtime/Utils/PoolLease.cs
```diff
@@ -0,0 +1,143 @@
...
+    /// </para>
+    /// <para>
+    /// One of these exists per pooled instance, not per rent, and it travels with its instance from
+    /// the pool to the renter and back. An instance whose lease is abandoned loses both together, so
+    /// nothing accumulates. Tracking rented instances in a set instead would turn a forgotten
+    /// <c>using</c> from "the garbage collector takes it" into a permanent leak, which is worse than
+    /// the defect being fixed.
+    /// </para>
+    /// </remarks>
+    internal sealed class PoolLease<T>
```
wallstop: Now we allocate - I want all struct-based IDIsposable to be zero allocation. But we're allocating - why?

Editor/Utils/AssetDatabaseBatchScope.cs
wallstop: Right on both counts, and the sweep found more than I had. Redone in `3fa3808`.

`AssetDatabaseBatchScope` keeps a `DisposalLease` — a `(slot, generation)` handle held **by value**, where the generation lives outside the struct so every copy reads the same one. A copy now decrements the batch depth once between them, not once each. It also stays a `readonly struct`, because a handle is a readonly field where the `bool` I first tried could not be, so `ScopeIsReadonlyStruct` passes untouched.

The full sweep, which is `rg 'struct \w+.*: .*IDisposable'` over `Runtime/` and `Editor/`:

| Type | Before | Now |
| --- | --- | --- |
| `PooledResource<T>` | `bool _initialized` | lease |
| `PooledArray<T>` | `bool _disposed` | lease (all three array pools) |
| `SemaphoreLease` | nulled field | lease, and now a `readonly struct` |
| `AssetDatabaseBatchScope` | nothing at all | lease |
| `AssetDatabasePauseScope` | `bool _disposed` | lease, and now a `readonly struct` |
| `Trie.Enumerator` ×2 | — | no change needed; holds `PooledResource`s |
| `SpatialHash2D`/`3D` `EntryBucket` | — | no change needed; holds a `PooledResource`, and it is stored **by value** in a `Dictionary`, so it is copied by construction |
| `LabelWidthScope` ×2 | — | left alone: restoring the same saved label width twice is already idempotent |

Two things the sweep turned up that were worth the trip:

**Four tests asserted the defect as the contract.** `MixedDoubleDisposalAndNormalDisposal` opened three scopes, disposed one of them twice, and asserted the depth fell from 3 to **1** — two scopes off the count for one scope ending, which runs `StopAssetEditing()` while a third scope is still live and leaves the rest of its writes unbatched. They now assert the corrected accounting, and there is a new `DisposingACopyAndTheOriginalDecrementsOnce` for the case nothing covered.

**`SemaphoreLease`'s "do not copy a lease" caveat is gone**, because copying is now safe. That paragraph existed precisely because a per-copy flag could not express this.

Runtime/Utils/PoolLease.cs
wallstop: Fair, and the measurement backs you up — I should have taken it before shipping the first version rather than after. Per cycle:

| path | `main` | heap lease (what you reviewed) | generational handle (now) |
| --- | ---: | ---: | ---: |
| steady-state rent/return | 0 B | 0 B | **0 B** |
| public `PooledResource(T, Action<T>)` | 0 B | **48 B** | **0 B** |
| `Get()` on a disposed pool | 32 B | **80 B** | **32 B** |

The pooled hot path was already zero — the lease object was per pooled *instance*, not per rent — but the two paths that build a lease per call were a straight regression, and a heap object per lease is the wrong shape regardless.

`PoolLease<T>` is deleted. In its place, `DisposalLease` is a `(slot, generation)` pair held **by value**; the generation lives in a segmented table outside the struct, so every copy reads the same one and only the copy still holding the current generation claims. This is the generational-handle (slotmap) pattern — DOTS's `Entity{Index, Version}`, Rust's `slotmap`, Vulkan handles — which is the standard way to invalidate a stale handle without allocating one.

Zero allocation comes from three choices:

- **Slots recycle through a per-thread free list.** A balanced acquire/release — what a `using` block does — reuses the caller's own slot with no atomics and no shared head for threads to contend on. `BalancedAcquireAndClaimReusesSlots` pins that 1,000 cycles create **zero** new slots.
- **The table grows in fixed blocks that are never reallocated**, so a live generation is never moved out from under a concurrent claim. `LeasesSurviveGenerationTableGrowth` holds 3,000 leases across that boundary.
- **Acquire advances the generation with a store-release, not an interlocked increment.** Sound because the acquirer is the slot's exclusive owner: a recycled slot reached the free list only through a winning claim that already advanced it past every outstanding lease, and it was pushed onto the free list of the thread now reading it. Worth ~21 ns/cycle — the difference between measurably slower and inside the noise.

Throughput, 2M cycles, 4 runs interleaved: `main` 241.3 / 208.4 / 189.3 / 204.2 (mean 210.8), lease 198.6 / 191.5 / 220.9 / 214.6 (mean 206.4). Inside the run-to-run noise on this box.

The remaining 32 B on the disposed-pool path is not overhead — it is the object being handed out. A disposed pool has no pooled instance left, so it must produce one. Proven rather than asserted: the same path with a non-allocating producer measures **0 B**, and a bare `new List<int>()` measures exactly **32 B**.

Pinned by `DisposalLeaseTests.AcquiringAndClaimingAllocatesNothing`, `PooledResourceDisposalTests.RentAndReturnAllocatesNothing` and `APubliclyConstructedLeaseAllocatesNothing`, all at a zero-byte tolerance.

One note on what is deliberately *not* asserted: the array pools allocate 32 B per rent on their own, because `ConcurrentStack` allocates a node per push. Measured identically with and without the lease, so it is pre-existing; I left a comment saying so rather than an assertion that would fail for a reason this change did not cause.

Runtime/Utils/DisposalLease.cs
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        /// <summary>
+        /// Puts a slot back on the calling thread's free list.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static void Release(int slot)
+        {
+            SetFreeNext(slot, _freeHead);
+            _freeHead = slot;
```
wallstop, line 215: Is this atomic? Can this, and similar scenarios race? Do we need special handling to ensure the right things happen in SINGLE_THREADED v non?

wallstop: Good question, and asking it caught a false green in my own test. Three parts.

## `Release` — not atomic, and does not need to be

`_freeHead` is `[ThreadStatic]`, so both lines touch only the calling thread's own free list; no other thread can observe or write it.

The `_freeNext` links do live in a shared array, so that is the part worth checking — and two threads can never write the same entry. A slot reaches `Release` only through a **winning** `TryClaim`, and the compare-and-swap elects exactly one winner, so at most one thread owns any given slot at any moment. Different threads therefore always write different indices, and distinct aligned `int` elements are independent. Nothing to tear, nothing to order.

Now documented on the method, rather than left to be re-derived by the next reader.

## The other unsynchronized write, which is the subtler one

`Acquire` advances the generation with a store-release rather than an interlocked increment. The obligation there is narrow and worth stating explicitly: **no outstanding lease may hold the generation being written.** None can —

- a recycled slot got to the free list only through a winning claim, which set the generation to `g + 1` while every lease outstanding on it held at most `g`, and nothing has since handed out a lease at `g + 1` because this call is the first;
- that winning claim ran on **this** thread (a slot is pushed onto the free list of the thread that claimed it), so the plain read sees its own write;
- a brand-new slot has never been leased at all.

A stale claim therefore fails its compare-and-swap rather than racing the write. That is also in the XML docs now, as the thing to re-check if anyone changes this.

## `SINGLE_THREADED`

Yes, and it was missing. `TryClaim`, `Acquire` and `CurrentGeneration` now strip the interlocked operations under the define, the same way the pools in `Buffers.cs` strip their locks — there is no second thread to race, so it is pure cost. Verified to compile with the define set, which nothing in CI does for this file today.

## The test that was lying

Before answering I mutated the compare-and-swap into a plain read/compare/write — exactly the mistake this question is about — and **all 13 tests passed**. Two independent reasons:

1. `ConcurrentAcquireAndClaimNeverHandsOneSlotToTwoOwners` claims both copies **on the same thread**, so it never raced the CAS at all. It is a good test of free-list integrity and table growth under load, and no test of atomicity.
2. `ConcurrentClaimsOfOneLeaseElectOneWinner` used a `Barrier`, whose own synchronization scatters the wake-ups far enough apart that a two-instruction window essentially never overlaps.

Rewritten: racers spin on a release flag rather than a barrier, and winners are summed across **3000 trials** instead of asserted per trial, so a single double-claim anywhere fails it. The mutation now fails it loudly — **3054 winners for 3000 leases**, 54 double-claims caught — and the real implementation passes.

The stress test stays, retargeted at what it genuinely covers, and gained an assertion that no slot is ever live on two threads at once.

wallstop:
## Green: 48 checks, 0 failures

Verified **by name from each leg's `results.xml`**, not by leg colour — a green leg does not prove a new test ran.

| Runtime | `DisposalLeaseTests` | `PooledResourceDisposalTests` | `SemaphoreLeaseTests` |
| --- | ---: | ---: | ---: |
| Mono, playmode 6000.5.2 | 13/13 | 15/15 | 20/20 |
| **IL2CPP standalone 2021.3** | 13/13 | 15/15 | 20/20 |
| **SINGLE_THREADED** 6000.3 | 10/10 | 14/14 | 20/20 |

The SINGLE_THREADED counts are lower by exactly the `#if !SINGLE_THREADED` tests, and `ConcurrentClaimsOfOneLeaseElectOneWinner` is correctly absent there.

**IL2CPP on 2021.3 was the real risk in this design**, and it is now proven rather than assumed: `[ThreadStatic]`, `Interlocked.CompareExchange(ref long)` and `ref long` returns all AOT-compile and behave, and the zero-allocation guarantee holds under IL2CPP and Mono, not just CoreCLR.

The six rewritten `AssetDatabaseBatchScopeTests` all ran and passed on editmode, `ScopeIsReadonlyStruct` included — the type is still a `readonly struct`.

`npm run validate:prepush` and `npm run validate:tests` both exit 0 locally, covering the 30 contract suites that never run in CI.

## One more test had pinned the defect

`SemaphoreLeaseTests.DisposingACopiedLeaseInflatesACountWithNoExplicitMaximum` asserted that copying a lease raises the permit count to **2**, with a comment reading "per-lease disposal tracking cannot prevent this: the copy carries its own flags" — an exact description of the defect, written as a contract. Rewritten to assert the count stays at 1, plus `ManyCopiesOfALeaseStillReleaseOnce` and `ACopyReportsNotHeldOnceAnotherCopyHasReleased`.

It reached CI rather than the local harness because `Tests/Runtime` is a PlayMode assembly. That gap is closed — `SemaphoreLease.cs` and its fixture now compile into the `dotnet` harness (53 tests), where mutating `TryClaim` so every copy wins fails 20+ of them.

Rather than wait for CI to find a fourth, I swept the remaining struct-disposal tests: `SystemArrayPoolDoubleDisposeIsNoOp`, `WallstopFastArrayPoolDoubleDisposeIsNoOp` and `WallstopArrayPool`'s equivalent all assert behaviour the lease preserves, confirmed by replicating each in the harness. **Three of eleven had encoded the bug.**

## Follow-ups filed

- **#367** — `WallstopArrayPool` allocates 32 B per rent on `main`, because `ConcurrentStack` allocates a node per push, in a pool whose purpose is avoiding allocation. Found while pinning the lease's cost; measured identically with and without this change, so deliberately out of scope here.
- **#365** and **#366** are closed by this PR rather than deferred — the by-value handle removed the reasons I had filed them as separate work.

Issue #347, opened 2026-08-06
ci: the Odin-gated source (16 files) compiles nowhere in CI, and that has already shipped a break once


wallstop:
Found while editing `WEnumToggleButtonsOdinDrawer.cs` for [#344](https://github.com/Ambiguous-Interactive/unity-helpers/pull/344) and looking for the job that would validate it. There is not one.

## Mechanism

`WALLSTOP_UNITY_HELPERS_ODIN_INSPECTOR` is produced by a `versionDefines` entry keyed on the `odininspector` package:

```jsonc
// Runtime/WallstopStudios.UnityHelpers.asmdef (and 7 more asmdefs)
"versionDefines": [
  { "name": "odininspector", "expression": "0.0.1", "define": "WALLSTOP_UNITY_HELPERS_ODIN_INSPECTOR" }
]
```

CI's ephemeral test project is generated from [`.github/unity-test-project-modules.json`](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/.github/unity-test-project-modules.json), which declares built-in modules plus `com.unity.ugui` and nothing else. `rg 'Sirenix|odininspector' scripts/unity/ .github/` returns **zero** matches.

So the define is never set, and **16 source files behind it are never compiled by any CI leg**:

```text
Editor/CustomDrawers/Odin/          9 drawers
Editor/CustomEditors/               WButtonOdin* inspectors and friends
Runtime/                            Odin serialized base classes
```

Not a syntax check, not a compile, nothing. They are validated only by whoever happens to have Odin installed locally.

## This is not theoretical

[#275](https://github.com/Ambiguous-Interactive/unity-helpers/issues/275) — *"Unity 6000.5 compile errors: 3.4.0 references Sirenix/Odin types without dependency"* — is this blind spot shipping a compile break to consumers. It was closed by fixing that instance. The gap that let it through is still open.

## Why it bites harder than usual

The Odin files are the ones most likely to break silently, because they duplicate logic from the non-Odin drawers. In #344 I changed a flag filter in **both** `WEnumToggleButtonsDrawer` and `WEnumToggleButtonsOdinDrawer`; the first is covered by four Unity legs, the second by nothing. Any divergence between the pair is invisible until a consumer with Odin hits it.

## Options

1. **Add an Odin compile leg.** Cleanest, but Odin is paid and license-bound, so it likely needs a self-hosted runner with a seat, or a vendored stub. Probably the real fix, and probably the expensive one.
[...]

Issue #284, opened 2026-07-11
[Bug]: SerializedDictionary Key/Value Visual Offset


wallstop:
### Description

In some scenerios, when inside a WGroup, the `Key` field appears to be offset

<img width="325" height="314" alt="Image" src="https://github.com/user-attachments/assets/21e79a63-93db-408a-a274-09b52e51e4f5" />

### Steps to Reproduce

```csharp
namespace Data.Levels
{
    using System.Collections.Generic;
    using System.Threading.Tasks;
    using Environment.Levels;
    using UnityEngine;
    using WallstopStudios.UnityHelpers.Core.Attributes;
    using WallstopStudios.UnityHelpers.Core.DataStructure.Adapters;
    using WallstopStudios.UnityHelpers.Core.Extension;
    using WallstopStudios.UnityHelpers.Core.Helper;
    using WallstopStudios.UnityHelpers.Utils;
#if UNITY_EDITOR
    using UnityEditor;
#endif

    public sealed class LevelDatabase : ScriptableObjectSingleton<LevelDatabase>
    {
        public IReadOnlyDictionary<string, string> ScenesByLevelName => _scenesByLevelName;

        public string InitialLevel => _initialLevel;

        [WGroup("Data")]
        [SerializeField]
        private string _initialLevel = string.Empty;

        [SerializeField]
        private SerializableDictionary<string, string> _scenesByLevelName = new();

#if UNITY_EDITOR
        [WButton("Load All Levels")]
        private async ValueTask LoadAllAsync()
        {
            _scenesByLevelName.Clear();
            List<string> nonLevelScenePaths = new();
            foreach (string scenePath in SceneHelper.GetScenesInBuild())
            {
                if (scenePath.Contains("MainMenu"))
                {
                    continue;
                }
                DeferredDisposalResult<LevelChecker> disposal =
                    await SceneHelper.GetObjectOfTypeInScene<LevelChecker>(scenePath);
                LevelChecker levelChecker = disposal.result;

                try
                {
                    if (levelChecker == null)
                    {
                        this.Log($"Failed to find relevant LevelChecker in scene {scenePath}.");
                        nonLevelScenePaths.Add(scenePath);
                        continue;
                    }

                    LevelConfiguration levelConfiguration = levelChecker.Configuration;
                    if (levelConfiguration == null)
                    {
                        this.LogWarn($"LevelChecker in scene {scenePath} has no configuration.");
                        nonLevelScenePaths.Add(scenePath);
[...]

PR #368, opened 2026-08-08
Stop the array pools allocating on every rent, and make Inspector indent survive an exception

### LINKED ISSUE #367: WallstopArrayPool allocates 32 bytes per rent, because ConcurrentStack allocates a node per push
Found while pinning the allocation cost of the disposal lease in #364, and deliberately left out of that PR because it is pre-existing and unrelated to it.

## Measured

`WallstopArrayPool<T>.Get(16)` followed by disposal, steady state, 100,000 cycles:

```text
main                       32.00 bytes/cycle
main + the disposal lease  32.00 bytes/cycle
```

Identical with and without the change, so it is not new. For comparison, `WallstopGenericPool<T>` measures **0.00** bytes/cycle over the same window.

## Mechanism

The threaded builds store free arrays in a `ConcurrentStack<T[]>`:

```csharp
// Runtime/Utils/Buffers.cs -- WallstopArrayPool<T>
private static readonly ConcurrentDictionary<int, ConcurrentStack<T[]>> _pool = new();

private static void Release(T[] resource)
{
    ...
    result.Push(resource);
}
```

`ConcurrentStack<T>` is a lock-free linked list: every `Push` allocates a `Node` to hold the item, and `TryPop` drops it for the GC. That is 32 bytes per return, in a pool that exists to avoid allocation. `WallstopFastArrayPool<T>` uses `ConcurrentStack` too and has the same cost; the `SINGLE_THREADED` builds use `Stack<T[]>`/`List<T[]>` and do not.

This is a documented property of `ConcurrentStack`, not a bug in it — `ConcurrentBag` and `ConcurrentQueue` have their own allocation shapes, and only an array-backed structure avoids per-item nodes entirely.

## Why it matters more than 32 bytes suggests

`SystemArrayPool<T>` is used 62 times in-package against 8 for `WallstopArrayPool` and 4 for `WallstopFastArrayPool`, so the blast radius is small today. The reason to fix it anyway is the guidance: `WallstopArrayPool` is documented as the pool to prefer "when array sizes are fixed or highly predictable (e.g., internal PRNG state buffers)" — exactly the tight-loop, per-frame case where a consumer reaches for a pool *because* they are counting bytes, and where 32 bytes per rent is the whole cost they were trying to remove.

## Candidate remedies, in rough order of preference

1. **An array-backed lock-free stack**: a `T[][] items` plus an index moved with `Interlocked`, growing under a lock. No per-push node, and it matches the `SINGLE_THREADED` shape so the two builds stop diverging in allocation behaviour.
2. **A lock plus `Stack<T[]>`**, matching the `SINGLE_THREADED` implementation. `WallstopGenericPool` already takes a lock per rent and measures 0 B/cycle, so the precedent exists and the contention profile is known.
3. **Leave it and document it**, so a consumer choosing between the array pools can see the trade rather than discover it with the profiler.

Whichever is chosen, pin it the way #364 pinned the lease — `GCAssert.DoesNotAllocate` at a zero-byte tolerance, over a window wide enough to exclude one-off warm-up costs (ten iterations is not; the pool's internal lists and usage tracker have not settled by then).

## Not in scope

`SystemArrayPool<T>`, which delegates to `System.Buffers.ArrayPool<T>.Shared` and whose allocation behaviour is the BCL's to define.

wallstop:
Two changes, both about a guarantee the code claimed but did not hold.

## 1. The array pools allocated 32 bytes on every rent (#367)

`WallstopArrayPool<T>` and `WallstopFastArrayPool<T>` kept their idle arrays in a `ConcurrentStack`, which is a lock-free linked list: every `Push` allocates a node and every `TryPop` drops it for the GC. Measured against the real `Buffers.cs` under a `dotnet` harness, 10,000 warm-up and 100,000 measured cycles:

| Pool | bytes/cycle |
| --- | ---: |
| `WallstopArrayPool<int>.Get(16)` | **32.00** |
| `WallstopFastArrayPool<int>.Get(16)` | **32.00** |
| `WallstopGenericPool` list rent | 0.00 |

Not a rounding error in a cold corner: `FlurryBurstRandom`, `PhotonSpinRandom`, `StormDropRandom` and `IListExtensions` rent on hot paths, and the pool's own docs recommend it for "PRNG internal state buffers" — exactly the per-frame case where a caller reaches for a pool *because* they are counting bytes.

Both pools now share `PoolBucket<T>`, an array-backed LIFO free list behind a monitor. **0 bytes.**

### The lock is measured, not assumed

| | `ConcurrentStack` | lock + array-backed `Stack` |
| --- | ---: | ---: |
| 1 thread | 40.0 Mops/s | 32.9 Mops/s |
| 4 threads on one bucket | 13.6 Mops/s | **25.4 Mops/s** |
| allocation | 32 B/op | **0 B/op** |

18% slower uncontended, **1.9x faster contended**, and ~1.3 GB/s of garbage removed at that rate. A correct lock-free array stack needs per-slot claiming and a bounded capacity — the naive `Interlocked` index has a window where a popper reads a slot the pusher has not written yet. `WallstopGenericPool` already reaches 0 bytes with a monitor, so both the precedent and the contention profile were known.

### Two things fell out of sharing the storage

**The `SINGLE_THREADED` and threaded bodies of both pools collapsed into one implementation each** — the synchronization is compiled out inside `PoolBucket` instead of duplicating ~300 lines of pool (and identical XML docs) around it. Net **−71 lines** despite adding a file.

**`WallstopFastArrayPool.Release` stopped indexing an unsynchronized `List<T>`.** Every other access took a `ReaderWriterLockSlim`; the release path took nothing, while a concurrent `Get` could be growing that same list. It is safe on x86 by accident of `List<T>`'s layout, but nothing ordered the publication of a resized backing array against its contents — the guarantee a weakly-ordered target (every ARM player) does not give. `PoolBucketTable` publishes a whole new array with `Volatile.Write`, so a reader sees the old table or the new one and never a half-resized one.

### Red-green, mutation-verified

`ArrayPoolSteadyStateRentAndReturnDoesNotAllocate` is data-driven over both pools, zero tolerance, 100,000-iteration window. Against `main`'s implementation both cases fail at exactly **3,200,000 bytes** — 32 bytes times the window — and pass at 0 with the fix.

### The sweep

Every pooled rent path in the package now measures **0 bytes/cycle**: `WallstopArrayPool`, `WallstopFastArrayPool`, `SystemArrayPool`, `Buffers<T>` List/HashSet/Queue/Stack, `Buffers.StringBuilder`. The two other concurrent collections `Runtime/` uses are segment-based rather than node-per-item and measure 0 too. `ConcurrentStack` was the only member of this class, and no use of it remains in `Runtime/`.

## 2. Every Inspector indent change now survives an exception

`WGroupGUI` raised `EditorGUI.indentLevel`, drew every property in the group through arbitrary user drawers, and lowered it again — with no `try`/`finally`. Nineteen sites across thirteen files had that shape.

A body throwing here is **ordinary, not exceptional**: Unity unwinds a drawer with `GUIUtility.ExitGUI()` whenever a control opens an object picker. The leaked increment then indents every property drawn afterwards for the rest of the pass, including the ungrouped properties `WButtonInspector` draws after a `[WGroup]` in the same loop.

Rather than nineteen `try`/`finally` blocks, this adds one scope — `Editor/Utils/IndentLevelScope.cs` — and routes every site through it. Three properties are deliberate:

- **It restores the level saved on entry rather than decrementing**, so it also heals a nested drawer that leaked an increment of its own.
- **It carries a `DisposalLease`**, so one scope produces exactly one restore however many copies of the struct exist — a copy disposed early would otherwise un-indent the rest of a still-open body, which is the #358 failure mode.
- **It is a `readonly struct`.** Unity's own `EditorGUI.IndentLevelScope` is a class and allocates on every repaint; this one allocates nothing.

Using `using` *declarations* wherever the span is a whole method body is what let a nineteen-site sweep land **without re-indenting a single existing block**, so the diff is the state handling and nothing else. After the sweep exactly one raw `EditorGUI.indentLevel` write remains outside the scope type: a per-child level set inside a loop whose enclosing scope owns the restore.

Seven new tests pin the contract, including the two properties a plain save/restore does not give: the level survives a throwing body, and a copy disposed first neither restores early nor lets the original restore twice over a level someone else has since set.

This is a **candidate** cause of #284 (a `SerializableDictionary` key column offset right when a `[WGroup]` is on the same object — the dictionary drawer indents the key column only when `indentLevel > 0`, which is exactly that geometry). Not claimed as fixed: reproducing it needs the reporter's asset, and a leak only persists across repaints if something throws every pass. #284 stays open.

## Also investigated: #347, the Odin-gated source compiles nowhere

Not landed, and the [analysis is recorded on the issue](https://github.com/Ambiguous-Interactive/unity-helpers/issues/347#issuecomment-5227466640). Short version: the gap is wider than reported (the local Unity gate has no Sirenix either, so the 16 files compile nowhere in the toolchain); an asmdef-based stub **cannot** work because both package asmdefs carry `references: []` with `overrideReferences: false` and Odin resolves only as auto-referenced precompiled DLLs; so the stub must be a DLL built on a runner that has an editor. The stub surface is ~150 lines, the plumbing is not, and none of it can be exercised locally — each iteration costs a licensed Unity leg on the single organization seat. It wants its own PR.

### Unity EditMode, and both mutations killed

| Run | Result |
| --- | --- |
| Fixtures this branch touches | **482 pass / 0 fail** (30.4 s) |
| `IndentLevelScopeTests` alone | **7 pass / 0 fail** |
| Mutant: decrement instead of absolute restore | **4 fail** — `ALeakInsideTheBodyIsHealedOnExit` (1 → 4), `AtLevelSetsAnAbsoluteLevel…` (5 → −1), `NestedScopesUnwind…`, `TheLevelNeverGoesBelowZero` |
| Mutant: remove the `DisposalLease` guard | **1 fail** — `ACopyDisposedFirstDoesNotRestoreTheLevelEarly` (7 → 2) |
| After reverting both mutants | **482 pass / 0 fail** |

Both properties a plain save/restore does not give are therefore covered by tests that can
actually fail: absolute restore by four of them, single-dispose-across-copies by exactly the one
written for it.

### CI, confirmed by name rather than by leg colour

- `unity-2022.3.45f1-editmode` `results.xml`: **4218 passed / 0 failed**, all seven
  `IndentLevelScopeTests` present and `Passed`.
- `unity-2021.3.45f1-playmode`: **8759 passed / 0 failed** — the leg that matters for #367, because
  the allocation assertions run under Unity's **Mono**, not the harness's CoreCLR. Zero tolerance
  over a 100,000-iteration window holds on both runtimes.

That check also caught a naming defect in my own test: `TestCaseData.SetName` replaces the whole
name, so `.SetName(nameof(WallstopArrayPool<int>))` produced `BuffersTests.WallstopArrayPool` — the
subject rather than the assertion, indistinguishable from twenty sibling tests, and invisible to a
search by method name on a leg where it had run and passed. Renamed to
`ArrayPoolSteadyStateRentAndReturnDoesNotAllocate(WallstopArrayPool)`.

### Everything else

- `dotnet` harness compiling the **real** `Buffers.cs` plus the real `BuffersTests`, `PooledResourceDisposalTests` and `DisposalLeaseTests`: **143 passed / 0 failed**.
- Mutation: both new allocation assertions fail against `main`'s pool at 3,200,000 bytes.
- Unity compile gate (6000.4, this working tree), content-asserted by loading the new types out of the compiled DLLs rather than trusting mtimes.
- `npm run validate:prepush` green.

Closes #367

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path array pooling and many Inspector drawers; behavior is heavily tested (allocation + concurrent pool scenarios + indent scope tests) but regressions could affect GC-sensitive runtime code or Inspector layout.
> 
> **Overview**
> **`WallstopArrayPool<T>` and `WallstopFastArrayPool<T>`** no longer keep idle arrays in `ConcurrentStack` (32 bytes allocated per push/pop). They share new **`PoolBucket<T>`** free lists (`Stack` + monitor; sync stripped when `SINGLE_THREADED`), with **`PoolBucketMap`** / **`PoolBucketTable`** for size-indexed buckets. The duplicate single-threaded vs threaded pool bodies collapse to one implementation each; **`WallstopFastArrayPool`** release no longer indexes a `List` without the same locking as growth. **`ArrayPoolSteadyStateRentAndReturnDoesNotAllocate`** asserts 0 bytes over 100k cycles for both pools.
> 
> **Editor IMGUI:** adds **`IndentLevelScope`** (alloc-free `readonly struct`, **`DisposalLease`** for single restore on copy/dispose, restores saved level not a blind decrement). Replaces manual `indentLevel++`/`--` and many `try`/`finally` blocks across property drawers, **`WGroupGUI`**, **`WButton`**, settings, and related utilities. **`IndentLevelScopeTests`** cover throws, nested scopes, and copy-dispose behavior.
> 
> **CHANGELOG** documents the pool GC fix ([#367]).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46ea7ce4cb3f2c2ece95139edf3cfd6df089ea4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Utils/PoolBuckets.cs
```diff
@@ -0,0 +1,198 @@
...
+    /// <remarks>
+    /// Array-backed rather than a lock-free linked structure because a pool exists to stop
+    /// allocating: <see cref="System.Collections.Concurrent.ConcurrentStack{T}"/> allocates a node
+    /// per push, which is 32 bytes charged to every single rent-and-return cycle.
+    /// </remarks>
+    internal sealed class PoolBucket<T>
+    {
+        private readonly Stack<T> _idle = new();
+#if !SINGLE_THREADED
+        private readonly object _gate = new();
```
wallstop: Should we just lock "_idle"?

Runtime/Utils/PoolBuckets.cs
```diff
@@ -0,0 +1,198 @@
...
+            {
+                return existing;
+            }
+
+#if SINGLE_THREADED
+            PoolBucket<T> created = new();
+            _buckets[size] = created;
+            return created;
+#else
+            return _buckets.GetOrAdd(size, new PoolBucket<T>());
```
wallstop: I think I have GetOrAdd extensions that "just work." here, also, this allocates always. Ours do not. Recommend utilizing here and anywhere else (do a sweep) that you might want to "GetOrAdd" in an allocating manner. We should never do that.

Runtime/Utils/PoolBuckets.cs
wallstop: Done in 6cd9fefe. The stack is private and never handed out, so nothing else can contend for its monitor and the separate gate object per bucket bought nothing.

Runtime/Utils/PoolBuckets.cs
wallstop: Done in 6cd9fefe, using `GetOrAdd(key)`, which constructs only on the miss the preceding `TryGetValue` just proved. It also dispatches on the dictionary type, so the SINGLE_THREADED and threaded bodies of that method collapse to the same line.

Swept `Runtime/` and `Editor/` as asked. Mine was the **only** call site passing an eagerly constructed value; the other eight all pass a non-capturing lambda, which Roslyn caches in a static field, so none of them allocate per call:

```text
AbstractRandom:810, StringWrapper:49, EnumExtensions:164/344,
ReflectionHelpers.TypeDiscovery:441, ReflectionHelpers:234/249/264,
Sibling/Parent/ChildComponentAttribute, EditorCacheHelper:170,
SerializableSetPropertyDrawer:1818
```

Re-measured after the change: every pooled rent path in the package still reports 0 bytes/cycle.

Issue #383, opened 2026-08-10
WallstopProto: hand Write the sub-message size Measure already computed


wallstop:
`WProtoWriter.TryWriteMessage` (session 174, #380) produces a sub-message's length prefix by
back-patching a finished payload: it reserves the prefix at its **minimum** width, writes the
payload, and shifts the payload right when the finished length needs a wider varint.

That is the right trade for small messages and the wrong one for large ones, and the crossover was
measured rather than guessed.

## The measurement

Depth-3 graph, one bulk member at the leaf. Shared host, so read the ratios rather than the
nanoseconds; the direction and magnitude are stable across runs.

| Sub-message size | back-patch vs re-measuring the child |
| ---: | --- |
| 0–128 B | **1.2–1.5x faster** |
| ~512 B | parity |
| 8 KB | 0.50x |
| 16 KB | 0.44x |
| 64 KB | 0.39x |
| 256 KB | 0.32x |

Both allocate 0 B/op. By depth at small payloads, back-patching is at parity to four levels and
**1.35x faster at twelve**, because re-measuring is quadratic in depth.

## Why the large case costs what it does

A prefix wider than one byte costs one `memmove` of the payload **per enclosing level**, while
re-measuring a bulk member is O(1) — `LengthDelimitedSize(blob.Length)` does not walk the bytes. At
256 KB and depth 3 that is four passes over the payload instead of one, which matches the measured
0.32x almost exactly.

## What would remove it

The size is already known. `Measure` computes every nested value's size during the single sizing pass
that allocates the destination buffer, and then throws it away; `Write` re-derives the prefix width
from the payload it just produced. Handing `Write` the number `Measure` already had would make the
shift unnecessary in every case, giving one measurement **and** one copy.

Three shapes were considered and none is free:

1. **A size-hint stack** threaded through a thread-static, pushed by `WProtoSizes.MessageSize` and
   popped by `TryWriteMessage`. Correctness must not depend on the hint — the pop is a *hint* for the
   reservation width, and a wrong hint degrades to today's shift rather than corrupting output. Needs
   care around a caller that measures twice, or measures and never writes.
2. **A size parameter on `IWProtoFormatter<T>`.** Cleanest to reason about, but the interface is
   public and hand-implementable by consumers, so this is a breaking change to the one type the
   design promises consumers can implement.
3. **Reserving the prefix from the writer's remaining capacity** rather than from a fixed minimum.
[...]

PR #384, opened 2026-08-10
Let a WallstopProto contract hold another one, and stop two local gates from lying

### LINKED ISSUE #380: WallstopProto: nested contracts need a decision on length prefixes versus the hook contract
`WProtoWriter` has no length back-patching, by design: `WProtoSizes` measures a sub-message so its length prefix can be written up front. That is what makes the writer a single forward pass over a `Span<byte>` with no scratch buffer.

It also means **a nested value is measured once per enclosing level**. A contract at depth 3 has `Measure` called on it three times for one serialization: once by the root computing its own total, once by its parent computing the length prefix, once by its own `Write`… and so on.

That collides head-on with the lifecycle hook contract, which `WProtoFormatterContractTests` pins today:

> before-serialization runs as the first statement of `Measure` and is **not** repeated in `Write` (a hook that rents pooled scratch would otherwise leak)

The 1:1 pairing of "one `Measure`, then one `Write`" is exactly what nesting breaks. A hooked contract nested two levels deep runs its before-serialization hook three times per write. If that hook rents pooled scratch — the case the contract was written to protect — it leaks two rentals every time.

## Why this is blocking rather than cosmetic

The source generator (#343, landed in session 173) refuses a `[WProtoMember]` whose type is another contract, with `WPROTO003`. That refusal is deliberate: emitting the obvious code would bake whichever answer we guessed into every generated formatter in every consumer's assembly, and the wire format is the one thing that cannot be changed later.

Nested messages are not exotic. Of the 65 `[ProtoContract]` types in `Runtime/`, the PRNG hierarchy alone nests through 17 `[ProtoInclude]` subtypes, and the surrogates nest by construction. Step 5 (the facade swap) cannot complete without this.

### 1. Length back-patching in the writer

Reserve a fixed-width varint (5 bytes for any length that fits 32 bits) for each sub-message, write the payload, then patch the length in place. Nothing is measured twice; the hook contract survives untouched.

Costs: the reserved prefix is up to 4 bytes larger than the minimal encoding, and **protobuf-net writes the minimal one**, so the bytes stop matching the oracle unless the payload is memmoved back over the slack. A memmove per sub-message is real work on a large save.

A variant worth measuring: patch *and* compact, so the output is still minimal, paying one `Buffer.BlockCopy` per nested message rather than one `Measure` per enclosing level. Which is cheaper is an empirical question and depends on nesting depth versus payload size — nobody should guess it.

### 2. Redefine the hook contract

State that before-serialization runs once per `Measure`, and that a nested value is measured once per enclosing level. Costs nothing at runtime, and removes the "rent pooled scratch in a hook" pattern the current wording protects — hooks would have to be idempotent and allocation-free.

This is the cheapest option and the one that changes a documented guarantee, so it needs to be a deliberate decision rather than a default.

### 3. A measurement cache threaded through the call

`Measure` returns sizes into a caller-supplied scratch list that `Write` then consumes. Preserves both the byte format and the hook contract, at the cost of an extra parameter on `IWProtoFormatter<T>` — a public interface consumers hand-implement — and a per-serialization allocation unless it is pooled.

## What would settle it

Measure options 1 and 3 against option 2 on a realistic nested corpus (the PRNG hierarchy is a good one: 17 subtypes, one level of include nesting, small payloads) for both throughput and allocation, and confirm option 1's compacted output stays byte-identical to protobuf-net. Then pick, write it down in `PLAN.md`, and lift `WPROTO003` for nested contracts.

## Related

- #343 — the WallstopProto rollout
- #377 — the nesting bound, which every nested read has to honour

### LINKED ISSUE #382: release: package.json ships pr-description.md while validate-npm-package forbids it
Two files disagree about the same filename, and the disagreement only bites during a release.

**`package.json`** lists it as something to ship:

```json
"files": [
  ...
  "scripts/postinstall-hooks.js.meta",
  "pr-description.md"
]
```

**`scripts/validate-npm-package.ps1`** lists it as something that must never ship:

```powershell
$forbiddenRootMarkdownArtifactPrefixes = @(
  'pr-description.md'
)
...
$errors += "Forbidden release artifact included in npm package: $entry"
```

## Why it is latent rather than currently broken

`pr-description.md` does not exist at the repo root today, so `npm pack` silently drops the `files` entry and the validator finds nothing to reject. Both checks pass. The contradiction is invisible.

It becomes a **release failure** the moment that file exists at pack time — which is precisely when someone is drafting a release description, i.e. during a release. `files` pulls it into the tarball, and the validator then fails the build for the entry `files` asked for.

## The fix

Remove `"pr-description.md"` from `package.json`'s `files` array. The validator's forbid-list is the intended behaviour (a PR description is a development artifact, not package content); the `files` entry is the mistake.

Worth adding a contract assertion alongside it: **no entry in `package.json` `files` may appear in `validate-npm-package.ps1`'s forbidden lists.** That is a cheap, purely static check, and it is the general form of this bug — the two lists are maintained independently today and nothing compares them, so the next disagreement will be just as invisible.

## Provenance

Found incidentally while reading `package.json` during session 173 (PR #381). Not fixed there: that PR was green across 53 checks including 14 Unity legs, and a `package.json` edit re-runs the entire matrix for a defect that cannot fire outside a release. It wants its own small change plus the parity check above.

wallstop:
Three things, in descending order of how much they matter.

## A WallstopProto contract can hold another one (#380)

A sub-message is length-delimited, so its size reaches the wire before its bytes exist. With the old
writer the only way to produce that number was to measure the child a second time inside `Write` —
and that unbalances the lifecycle hooks, because before-serialization belongs to `Measure` and
after-serialization to `Write`:

```text
depth-3 graph, one serialization
  measure-per-level : before=4  after=1
  back-patch        : before=1  after=1
```

Four rentals in, one release out, for exactly the pooled-scratch hook the contract's wording exists
to protect. That is why the generator refused nested members (`WPROTO003`) rather than guessing, and
it is the whole of the decision.

`WProtoWriter.TryWriteMessage(fieldNumber, formatter, value)` now writes the key, the prefix and the
payload as one operation. It reserves the prefix at its **minimum** width and shifts the payload only
when the finished length needs a wider varint. Reserving one byte rather than the maximum is what
makes the shift provably safe — the payload is always written to the *left* of where it ends up, so
it can never run past a buffer sized from `Measure` — and it means a sub-message under 128 bytes
moves no bytes at all. Every value is measured **exactly once per serialization at every depth**, so
the hook contract holds verbatim instead of being rewritten.

### The trade, measured

Shared host, so the ratios are the signal and the nanoseconds are not.

| Sub-message size | back-patch vs re-measuring |
| ---: | --- |
| 0–128 B | **1.2–1.5x faster** |
| ~512 B | parity |
| 8 KB | 0.50x |
| 64 KB | 0.39x |
| 256 KB | 0.32x |

By depth at small payloads: parity to four levels, **1.35x faster at twelve**, because re-measuring
is quadratic in depth. Both allocate **0 B/op**. The large case costs one `memmove` per enclosing
level while re-measuring a bulk member is O(1) — taken deliberately, since the small case is the
per-frame one and the large case is a save. The optimization that removes it is #383, with the
measured crossover.

### Two encoding rules came from the oracle, and one is backwards

Asked protobuf-net 3.2.56 rather than reasoning about it:

```text
all defaults                       -> 12-00
struct member = default(Point)     -> 08-01-12-00
nullable struct = null             -> 08-01-12-00
nullable struct = default          -> 08-01-12-00-1A-00
reference member = null            -> 08-01-12-00
reference member = empty instance  -> 08-01-12-00-22-00
```

A **struct** sub-message is written even when every member equals its default — the opposite of the
default-omission rule every scalar follows. Guessing would have produced a payload protobuf-net never
writes.

### Cycles became expressible, so measurement needed a bound

`WPROTO003` had made a cyclic contract graph impossible; lifting it makes one expressible, and
measurement walks the graph. `WProtoSizes.MessageSize` carries the reader's `MaxNestingDepth` and
reports passing it **by name**. There is no value it could return instead — a cyclic message has no
finite encoded size — and the alternative is a stack overflow, which cannot be caught. The writer
carries the same bound for a formatter that writes without measuring first.

## `package.json`'s files array is now compared against the validator that can reject it (#382)

The reported contradiction is not there: `git log --all -S 'pr-description.md' -- package.json`
returns nothing, so the `files` entry never existed. Corrected on the issue.

The concern behind it is real and nothing was checking it. The `files` array and
`validate-npm-package.ps1`'s lists are two hand-maintained descriptions of one payload, and the
existing contracts only assert that each list *contains* what a release needs. The two ways they can
disagree are not equally loud: an entry `files` ships that the validator forbids fails the release at
the worst moment, while an entry the validator allows that `files` omits means **npm simply does not
ship it** and nothing says so. Three static assertions close both directions, in the one contract
suite that runs in CI rather than only in the pre-push hook.

## Generated `.meta` files get the line endings `.gitattributes` asks for

Adding any new asset left `npm run validate:prepush` failing on a file nobody hand-wrote.
`.gitattributes` declares `*.meta text eol=crlf` and `check-eol` enforces that in the working tree,
but a generated file never passes through git's smudge filter and both generators wrote LF.

`npm run agent:preflight:fix` could not clear it either, which is the part that matters — it is
documented as the thing that makes `validate:prepush` pass. Its EOL normalization runs *before* the
step that creates the `.meta`, so the file it just wrote is never in the set it already normalized.
Both writers now emit CRLF at creation. Observed twice, which is why it gets a test:
`scripts/tests/test-analyzer-placement.js.meta` went in with LF in session 173 and this session
produced two more.

---

## Verification

| What | Result |
| --- | --- |
| Generator suite (analyzer-referenced, real output) | **36/36** |
| `dotnet` harness over the real wire core + protobuf-net 3.2.56 oracle | all checks pass |
| Byte identity vs protobuf-net, nested, 11 payload sizes to 20 KB | 11/11 |
| Prefix-width sweep vs oracle, payload 100–135 and label 100–140 | byte-identical throughout |
| Cross-deserialization (protobuf-net reads WProto's nested bytes) | 4/4 |
| Mutations on the nested-contract work | **8/8 caught** |
| Package manifest parity contract | 149 checks, 0 failures; 3 mutations caught |
| `.meta` line-ending contract | 1 mutation caught |
| Unity 6000.4 compile gate | all four nested contracts loaded, **each with a generated formatter** |

The compile gate answers the question session 173 could not: Unity's own Roslyn compiled the new
contracts and the shipped analyzer emitted a nested formatter for every one of them, including the
**recursive** one.

**The committed golden hex was verified without Unity.** The fixture asserts two literal byte
prefixes at the 127/128 boundary. Rather than let a hand-computed constant find out on a 90-minute
Unity leg, the same three contracts were mirrored into the generator test project, the real generator
asked what it emits, and the answer compared — all four constants matched, and the mirror was
deleted.

### The mutation harness lied, and the mutation it hid was the real one

The first pass reported eight of eight caught. One of them — a faulted nested write reported as
success — was failing tests that had nothing to do with it; re-run alone it was **36/36 green**. The
harness restored the generator's *source* after each mutation but only rebuilt the analyzer for
mutations that needed one, and building the generator binplaces the DLL the test project loads, so
three later mutations were scored against the previous mutation's binary.

The harness now rebuilds on every run and drives both suites. The survivor got the test it never had:
a formatter that ignores a refused write and returns `true` would otherwise produce a length prefix
that *correctly* describes a payload with a member missing — internally consistent, and silently
wrong. The interface is public and hand-implementable, so this is not hypothetical.

The tell for a contaminated mutation run is a mutation whose failing tests look unrelated to it.

## Not verified locally, and why

`Tests/Runtime/**` is a PlayMode assembly, and PlayMode through the MCP bridge is unproven — session
173 got a PlayMode run only by calling the runner directly, which locked the editor for the rest of
that session. Trading a working compile gate for one run was not worth it, given the same behaviors
are covered 36/36 against the same generator and the same runtime sources with eight mutations behind
them. CI is the gate for the fixture itself, for Unity 2021.3's Roslyn 3.9, and for IL2CPP AOT of
nested formatters.

Closes #380. Closes #382. Follow-up: #383.

---

> [!NOTE]
> **Medium Risk**
> Changes core serialization wire format and generator output used for saves/network data; risk is mitigated by protobuf-net oracle tests, nesting bounds, and extensive new coverage, but regressions would affect player data compatibility.
> 
> **Overview**
> **WallstopProto** now lets `[WProtoMember]` types be other `[WProtoContract]` types (including self-reference), lifting the old `WPROTO003` refusal. The generator emits nested writes via **`WProtoWriter.TryWriteMessage`**, which writes tag + payload first and **back-patches** the length prefix so each graph is **measured once**—keeping before/after serialization hooks paired at any depth. **`WProtoSizes.MessageSize`** mirrors that sizing and enforces **`MaxNestingDepth`** on measure (cycles throw `InvalidOperationException` instead of stack overflow). Wire rules follow protobuf-net: struct sub-messages always serialize; `IsRequired` writes default values but not null references.
> 
> Docs and tests cover nested round-trips, prefix-width boundaries, hook counts, and IL2CPP fixtures. **`.llm`** adds foreign-call/cast defensive patterns tied to `TryWriteMessage`.
> 
> **Release hygiene:** CI contract tests now cross-check **`package.json` `files`** vs **`validate-npm-package.ps1`** allow/forbid lists both ways; **`generate-meta.sh`** and **`agent-preflight.ps1`** write **CRLF** for new `.meta` files to match `.gitattributes`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1eff5f4517e713e76efab5df346b135f25bd45ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Serialization/WallstopProto/WProtoWriter.cs
```diff
@@ -353,6 +362,94 @@ public bool TryWriteRaw(ReadOnlySpan<byte> value)
...
+
+            if (!TryReserve(1, out int prefixStart))
+            {
+                return false;
+            }
+
+            int payloadStart = _position;
+            _depth++;
+            bool written = formatter.Write(ref this, value);
+            _depth--;
```
wallstop: Should this be try/finally?

Runtime/Core/Serialization/WallstopProto/WProtoWriter.cs
```diff
@@ -353,6 +362,94 @@ public bool TryWriteRaw(ReadOnlySpan<byte> value)
...
+            if (!written || _faulted)
+            {
+                _faulted = true;
+                return false;
+            }
+
+            int length = _position - payloadStart;
+            int prefixSize = WProtoSizes.Varint32Size((uint)length);
```
wallstop, line 444: is length ABSOLUTELY guaranteed to be positive here? Is this safe?

Runtime/Core/Serialization/WallstopProto/WProtoWriter.cs
```diff
@@ -353,6 +362,94 @@ public bool TryWriteRaw(ReadOnlySpan<byte> value)
...
+            if (prefixSize != 1)
```
wallstop: Should we do a smarter comparison that is safer? What if prefixSize = 0?

Issue #428, opened 2026-08-13
ci: Copilot reviewer quota leaves otherwise-green PRs red


wallstop:
## Problem

The automatic `copilot-pull-request-reviewer` check fails before reviewing when the requesting account has exhausted its monthly Copilot review quota. This leaves an otherwise-green pull request with an unavoidable red check.

Observed on #427, twice (before and after the review-fix push):

- workflow run: https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31734901523
- job: https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31734901523/job/94564063352
- terminal error: HTTP 402 / `You have exceeded your monthly quota`
- no code analysis or actionable review was produced
- 54 repository-controlled checks passed and 7 were intentionally skipped on the same head

The first attempt failed identically in run `31734114963`.

## Impact

- GOAL.md's literal all-checks-green acceptance cannot be achieved from repository changes.
- Pushing a valid reviewer fix reruns the reviewer and reproduces the same red check.
- Retrying cannot help until quota resets or the account/organization entitlement changes.
- The red status is indistinguishable at a glance from an actionable code-review failure.

## Acceptance

Choose and document one supported policy:

1. restore/increase the Copilot review quota so automatic review can complete;
2. make quota exhaustion neutral/skipped rather than failure, if GitHub exposes such configuration; or
3. remove the Copilot reviewer from required/all-green acceptance and rely on the configured Cursor review plus repository CI when Copilot reports quota exhaustion.

After remediation, rerun the failed reviewer on #427 and confirm that it completes without the HTTP 402 quota error.

wallstop:
By design

Issue #398, opened 2026-08-10
WallstopProto: no allocation benchmark, and the read path still allocates per call


wallstop:
## Context

Raised in review on #389: "Can we/should we pool/be more intelligent about allocations here and
elsewhere? I want to make these paths hyper-optimized and allocation friendly."

## Done already

`WProtoFacade.Serialize<T>(T value, ref byte[] buffer)` removes the per-call array allocation on the
write path. It reuses the caller's buffer when it fits, grows it when it does not, returns the byte
count, and `-1` when the type is not served. A caller serializing every frame can keep one scratch
buffer for the process lifetime.

## What still allocates

The **read** path, which a caller-supplied buffer cannot fix:

- `WProtoReader.TryReadString` calls `Encoding.UTF8.GetString`, allocating a string per field per
  message. The obvious candidates are a string interner keyed on the payload bytes, or a
  `ReadOnlySpan<byte>` accessor that lets the caller decide.
- `byte[]` members allocate per element on read.
- The generated read path allocates the collection for every repeated member, including when the
  caller intended to merge into an existing one.

Boxing is worth an explicit check too: `IWProtoScalarFormatter<T>` is an interface, and a call
through it on a struct `T` boxes unless the call site is generic-constrained. It should be fine as
written, but "should be" is not a measurement.

## The blocker is measurement, not ideas

**There is no allocation benchmark in this suite.** Every claim above is reasoning about code, not
data, and the fix philosophy here is data-backed. Adding pooling without a benchmark risks making
things slower and being unable to tell.

Suggested first step, before any further optimization:

1. A benchmark harness in `Generator~` measuring bytes allocated per serialize and per deserialize,
   for a representative contract (scalars, a string, a repeated field, a map, a nested message).
2. Record the current numbers as the baseline.
3. Only then decide what to pool - and gate each change on a measured improvement.

The `Generator~` harness already compiles the real runtime sources against NUnit, so this is a small
addition rather than new infrastructure. Note that GC behaviour under IL2CPP differs from Mono, so a
desktop measurement bounds the problem rather than settling it.

wallstop:
Session 184 now implements the benchmark-first step requested here on the local candidate branch.

[...]

Issue #343, opened 2026-08-05
feature/bufix: AOT-Compatible Protobuf Serialization


wallstop:
### Problem (one paragraph)

protobuf-net v3's runtime engine is fundamentally incompatible with IL2CPP: it constructs serializers by
reflectively instantiating `internal` closed generics (`ConcreteStub<T>`, `ListSerializer<T>`,
`StructValueChecker<T>`) via `MakeGenericType`, which IL2CPP cannot AOT-compile, and it calls the
IL2CPP-unsupported `RuntimeParameterInfo::GetTypeModifiers` icall during model build. This is
maintainer-confirmed (protobuf-net#699/#792/#1025/#1190); the real protobuf-net AOT story ("generators") is
still not shipped. No `link.xml`/`[Preserve]`/`AotHelper` manifest can fix it (the failing generics are
`internal` and reflectively reached). Yet we **require** runtime protobuf serialization in shipped IL2CPP
players (player data) **and** must stay byte/wire-compatible with data already written by protobuf-net —
**v2 *and* v3 both**, because this is a library and clients in the wild have persisted data with either. At
CompatibilityLevel 200 v2 and v3 emit the same wire format, but we treat that as a claim to *prove*, not
assume: the conformance suite uses a **dual oracle** (protobuf-net v2.4.x **and** v3.2.x) and fails if they
ever disagree on a contract we ship.

### Why an in-tree source generator is the *right* answer (not a fallback)

| Option | IL2CPP runtime | v3 wire-compat | Covers arbitrary **consumer** generic element types (`Deque<TheirStruct>`) | Verdict |
| --- | --- | --- | --- | --- |
| AOT hints / `[Preserve]` / `link.xml` | ❌ (internal reflective generics) | ✅ | ❌ | **Dead** (proven) |
| Downgrade to protobuf-net v2.4.x + precompiled model | ⚠️ (linker-version-sensitive) | ✅ (also Level 200) | ❌ (precompiled model is closed at package build) | EOL, partial |
| Migrate to Google.Protobuf / MemoryPack | ✅ | ❌ (`[ProtoInclude]`/surrogate encoding won't round-trip) | ✅ | Breaks existing data |
| **In-tree source-generated serializer (WallstopProto)** | ✅ (zero runtime reflection) | ✅ (reproduces Level 200 exactly) | ✅ (**generator runs at the consumer's build**) | **Chosen** |

The decisive property: a Roslyn source generator emits static serializer code **at the compilation that
[...]

PR #429, opened 2026-08-13
Size a repeated read from the count already in the packed run

wallstop:
## Size a repeated read from the count already in the packed run

Closes the read-allocation half of #398 (which is the open per-frame piece of #343), and records the
only answer this repository can give to #428.

### The problem, measured

WallstopProto's write path reuses a caller's buffer and allocates nothing. Its **read** path
allocated **5,272 B/op against protobuf-net's 4,112** for the same object graph — 1,160 bytes of
garbage per deserialization that the returned value does not account for. On a per-frame or per-save
read that is a frame-time spike.

#398 said to measure before pooling anything. The aggregate benchmark existed (session 184) but it
mixes five member shapes into one number and cannot say which one to fix, so the first change here is
an instrument: one contract per member shape, each measured against protobuf-net decoding the
**identical** graph. It attributed the entire gap to two shapes and exonerated three:

| Shape                        |     Before |   After | protobuf-net |
| ---------------------------- | ---------: | ------: | -----------: |
| `int[128]`                   | 1,744 B/op | **560** |          560 |
| `List<int>[128]`             | 1,208 B/op | **592** |          624 |
| `string`                     |    88 B/op |      88 |           88 |
| `Dictionary<string,int>[32]` | 3,384 B/op |   3,384 |        3,384 |
| nested contract              |    96 B/op |      96 |           96 |

### The fix

The cause was the accumulator, not the reader: an array member decoded into a `List<T>` that doubled
from empty and was then copied out of, so 128 `int`s left six abandoned buffers plus a duplicate of
the answer.

A packed run already carries the count. `WProtoReader.CountPackedElements` reads it off the encoded
bytes without consuming them — exactly, because a varint element ends at the byte whose continuation
bit is clear and a fixed-width run divides — and the generator spends it: arrays accumulate into
`WProtoArrayBuilder<T>` (a struct that reserves once and hands its buffer over when it comes out
exactly full), and every `List<T>`-backed accumulator is sized through `WProtoRepeated.Reserve`.
Collections with no portable capacity API (`HashSet<T>`, `Queue<T>`, `LinkedList<T>`, consumer types)
are untouched, and an **unpacked** run — what protobuf-net writes — has no knowable length and grows
exactly as before.

**Aggregate: 5,272 → 4,088 B/op, below the oracle, throughput unchanged** (2,153–2,237 ns/op against
a 2,213 baseline). No wire-format change; no golden vector moved.

Both gates now assert against protobuf-net's own allocation rather than a hand-written ceiling — two
implementations returning the same graph must allocate the same, and no constant needs re-tuning when
a runtime changes what a `Dictionary` costs.

### Verification

- Generator suite vs protobuf-net **3.2.56**: 318/318 (five runs); vs **2.4.9**: 317/317.
- `npm run typecheck:unity`: all four legs, zero warnings. `npm run validate:local`: green.
- Real Unity **6000.4.6f1** editor over the MCP bridge: **39/39 package assemblies fresh**, empty
  console, **210/210** WProto fixtures green in PlayMode (read-sizing, collections, maps, includes,
  surrogates, marshals, generics, facade, wire format).
- The editor caught a defect the desktop suite could not: a fixed 4 KB scratch in the new fixture
  could not hold 1,000 mixed-sign `int`s (a negative `int32` varint is ten bytes). Only the
  1000-element case failed; the scratch is sized from the encoding now.
- No Unity license is configured in this devcontainer, so the Docker EditMode/PlayMode legs are CI's.

### Also in this PR

**#428** — the automatic Copilot reviewer fails with HTTP 402 (`exceeded your monthly quota`) before
reading any code, and every push re-requests it and reproduces the same red. Nothing in the repository
can clear it. The supported policy (the issue's option 3) is now written where a landing session reads
it — `ship-changes` Step 10 — with the signature to recognize it by. Restoring the quota or dropping
the reviewer from the required set remains an owner action in organization settings.

### Follow-ups

What is left of #398 is the string interner and merging into a caller's existing collection. Both
change what a read hands back rather than how it accumulates, and neither is indicted by the
measurement: `string` and the map path already allocate exactly what protobuf-net does.

---

> [!NOTE]
> **High Risk**
> Changes deserialization allocation behavior and adds security bounds on save/payload restore paths across multiple collection types and generated formatters; incorrect limits or counting could affect compatibility or deny legitimate large saves.
> 
> **Overview**
> **Packed repeated deserialization** no longer grows `List<T>` accumulators and copies into arrays. The generator emits reservations from **`WProtoReader.CountPackedElements`**, uses **`WProtoArrayBuilder<T>`** for array members, and **`WProtoRepeated.Reserve`** for list-backed shapes—aggregate read allocation drops from ~5,272 B/op to ~4,088 B/op (at or below protobuf-net for the same graph), with new per-shape allocation benchmarks gating regressions.
> 
> **Untrusted capacity fields** can no longer drive multi‑GB allocations: **`SerializationCapacityLimits`** clamps growth-hint capacities (`Deque`, `BitSet` JSON) or refuses semantic ones (`SparseSet`), with tests for the six-byte `int.MaxValue` attack and configurable **`MaximumRestoredCapacity`**.
> 
> Docs/skills add **untrusted-payload-limits** (critical rule 17), serialization guide notes, **ship-changes** guidance for non-repo Copilot quota failures (#428), and benchmark throughput checks use fastest-round minima instead of medians to avoid CI flakes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebb18c3fd157fa8fd2861e25952f0b99f6f6e435. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Serialization/WallstopProto/WProtoArrayBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+            }
+
+            T[] exact = new T[_count];
+            Array.Copy(_items, exact, _count);
+            return exact;
+        }
+
+        private void Resize(int capacity)
+        {
+            T[] grown = new T[capacity];
```
wallstop, line 180: Isn't there an official Array. pattern for this?

Issue #445, opened 2026-08-15
test: contracts that name a workflow filename where they mean a requirement (4 found in one refactor)


wallstop:
## The pattern

Session 190 consolidated 24 lint workflows into one. Four separate contract tests broke — not
because the thing they protect stopped being true, but because each one **named a file or a job
where it meant a requirement**:

| Contract | Named | Actually required |
| --- | --- | --- |
| `test-unity-workflow-matrix-contract.ps1` | `pwsh-invocations-lint.yml` by path | *some* workflow checking out the cleanup policy derives its ref from `resolve-build-lock-pin.js` |
| `validate-devcontainer-config.ps1` | `validate-devcontainer.yml` must exist | devcontainer validation runs when the publish workflow changes |
| `test-sync-script-contracts.ps1` (×2) | `validate-links:` / `validate-link-format:` job blocks | both `-Mode Targets` and `-Mode Format` run, bounded |

All four were fixed in #444 by locating the owner instead of naming it. This issue is about the
**remaining instances**, because four in one refactor is a pattern rather than a coincidence.

## Why it matters

A contract that names a filename has two failure modes, and the quiet one is worse:

1. **Loud** — the file is renamed or removed, the contract fails on the name, and someone spends
   time discovering the requirement is still satisfied. That is what happened four times here.
2. **Quiet** — the *work* moves to another file while the named file still exists. The contract
   keeps passing against a file that no longer does the thing. Nothing goes red.

`PLAN.md` already records this lesson twice in other contexts ("a pinned SHA restated in a test is
not a pin", "derive the thing from where it actually lives"), which is why it is worth sweeping
rather than fixing case by case.

## Proposed work

1. **Sweep** `scripts/` for contracts that hard-code `.github/workflows/<name>.yml`. A first pass:

   ```bash
   rg -n "workflows/[a-z0-9-]+\.yml" scripts/
   ```

2. For each, decide whether the filename **is** the requirement (e.g. a contract specifically about
   `release.yml`'s publish steps — legitimate) or merely where it currently lives (rewrite to
   discover).
3. When rewriting to discover, two rules learned the hard way in #444:
   - **Strip comments before matching.** The first rewrite matched workflows whose *comments*
     mentioned `run-repo-lint.js`, which the slimmed workflows do in prose explaining where their
     jobs went. `test-workflow-repository-guard.ps1` already strips comments for exactly this reason.
[...]

Issue #435, opened 2026-08-14
ci: the MCP command sandbox cannot reach package types, so a change can be compiled in the editor but not exercised there


wallstop:
Session 187 verified a new serializer capability against a live Unity 6000.4.6f1 editor over the MCP
bridge and could confirm only half of what it wanted to.

**What works:** the editor picks up working-copy changes, recompiles, and reports errors. That is a
real gate — it proved the generated code for ten nested-collection shapes is valid under Unity's own
Roslyn with the shipped analyzer, with an empty console.

**What does not:** running anything against those types.

1. `Unity_RunCommand` compiles its script into a sandbox assembly that references `UnityEngine` and
   `UnityEditor` but **not** the package assemblies. `WProtoFormatterProvider`, `IWProtoFormatter<T>`
   and any test contract are unreachable by name (`CS0234`/`CS0246`). It also lacks a `System.dll`
   reference, so even `HashSet<int>` fails to compile (`CS0012` on `ISet<>`).
2. `System.Reflection` is on the sandbox's namespace denylist, so the reflection workaround is
   refused outright: `UNEXPECTED_ERROR: Script uses one or more unauthorized namespaces`.
3. Driving `TestRunnerApi` from the sandbox produced no callbacks. `RetrieveTestList` is asynchronous
   and the command returns before it fires; a filtered `Execute` (by `testNames`, then by
   `groupNames` regex, with the listener as a `HideAndDontSave` `ScriptableObject`) logged nothing at
   all, not even `RunStarted`. Plausibly the sandbox assembly is unloaded once the command returns,
   which would take the registered listener with it.

Related to #400 (the runner cannot see this package's tests because the host project lacks
`testables`), but distinct: even with discovery fixed, (1) and (2) block the direct route.

**What would fix it**, roughly in order of how much it buys:

- Have the sandbox assembly reference every loaded player/editor assembly, or accept an explicit
  reference list. This makes the bridge able to exercise package code, not just compile it.
- Failing that, a bridge-side "run tests and return results" verb that owns the async lifetime, so
  the listener does not live in a disposable assembly.
- At minimum, document the limitation where a session will hit it — the MCP notes in
  `.llm/context.md` currently read as though the editor can be used to test changes, and it can only
  be used to compile them.

wallstop:
Sharpening this from session 189, where it blocked the only behavioural in-editor check available.

[...]

Issue #434, opened 2026-08-14
WallstopProto: rectangular arrays (int[,]) need a shape header, which jagged arrays did not


wallstop:
Split out of #399, whose jagged/nested half shipped in session 187.

`int[][]`, `List<int[]>`, `List<List<int>>` and friends are served now by a wrapper message per
inner collection — `message Wrapper { repeated T values = 1; }`. A **rectangular** array is not the
same question wearing a different shape:

- A jagged array carries its structure in its own elements. Each row is a separate object with its
  own length, so a wrapper per row reconstructs it exactly.
- A rectangular array does not. `int[3,4]` is one object whose 12 elements have no per-row boundary,
  and `GetLength(d)` is metadata that lives nowhere on the wire. A flat run of 12 varints could be
  read back as `[3,4]`, `[4,3]`, `[2,6]`, `[12,1]` or `[1,12]` and nothing in the payload chooses.

So supporting it means **carrying the dimensions**, which is a wire-format decision rather than a
missing case. The natural proto3 spelling is:

```proto
message Rect {
  repeated int32 dims = 1;   // packed, one entry per rank
  repeated T values = 2;     // row-major
}
```

`protoc` would accept that, so it satisfies the same "maps to a concrete proto3 type" condition the
jagged work was approved under (#399, 2026-08-13).

**What has to be decided before it is written**

1. **Rank bound.** `T[,,,,]` is legal C#. Is there a maximum rank, or is it whatever `dims` holds?
2. **Untrusted `dims`.** This is squarely [critical rule 17](.llm/skills/untrusted-payload-limits.md):
   `dims` is a **capacity claim**, not a length. Six bytes can say `[46341, 46341]` and ask for 8 GB.
   The delivered `values` count is the only floor, so the reader must refuse any `dims` whose product
   does not equal the delivered element count — and compute that product in `long` so it cannot
   overflow into something plausible. (Session 186's sweep found exactly this class already;
   `SerializationCapacityLimits.TryAccept` is the existing verb for a semantic claim.)
3. **Zero-length dimensions.** `new int[0,5]` has zero elements but rank 2 and a real shape. It must
   not decode to `int[0,0]`, which means `dims` has to be written even when `values` is empty — the
   opposite of the omit-empty rule everywhere else.
4. **`Array.CreateInstance` under IL2CPP.** The generated reader needs to build a rank-N array from
   a runtime rank. `Array.CreateInstance(typeof(T), lengths)` is reflection, which this package does
[...]

PR #447, opened 2026-08-15
Cache the credential, fuzz the reader, and stop loading every prefab

### LINKED ISSUE #280: [Bug]: SendMessage cannot be called during Awake...
### Description

SendMessage cannot be called during Awake, CheckConsistency, or OnValidate (Water Color: OnMeshFilterChanged)
UnityEngine.MeshFilter:set_sharedMesh (UnityEngine.Mesh)
WaterMesh2D:BuildMesh () (at Assets/Scripts/Rendering/WaterMesh2D.cs:629)
WaterMesh2D:Generate () (at Assets/Scripts/Rendering/WaterMesh2D.cs:519)
WaterMesh2D:ApplySettings (WaterMesh2D/WaterSettings,WaterMesh2D/WaterShaderSettings) (at Assets/Scripts/Rendering/WaterMesh2D.cs:358)
WaterEnvironment:UpdateActionSize () (at Assets/Scripts/EnvironmentalEffects/WaterEnvironment.cs:258)
EnvironmentalEffectBase:OnValidate () (at Assets/Scripts/EnvironmentalEffects/EnvironmentalEffectBase.cs:204)
UnityEditor.AssetDatabase:LoadAllAssetsAtPath (string)
WallstopStudios.UnityHelpers.Editor.AssetProcessors.DetectAssetChangeProcessor:HasMatchingSubAsset (string,System.Type) (at ./Library/PackageCache/com.wallstop-studios.unity-helpers@3.5.0/Editor/AssetProcessors/DetectAssetChangeProcessor.cs:1054)
WallstopStudios.UnityHelpers.Editor.AssetProcessors.DetectAssetChangeProcessor:AppendCreatedAssets (System.Type,System.Collections.Generic.IReadOnlyList`1<string>,System.Collections.Generic.List`1<string>) (at ./Library/PackageCache/com.wallstop-studios.unity-helpers@3.5.0/Editor/AssetProcessors/DetectAssetChangeProcessor.cs:1022)
WallstopStudios.UnityHelpers.Editor.AssetProcessors.DetectAssetChangeProcessor:CollectCreatedAssets (System.Type,System.Collections.Generic.IReadOnlyList`1<string>,System.Collections.Generic.IReadOnlyList`1<string>) (at ./Library/PackageCache/com.wallstop-studios.unity-helpers@3.5.0/Editor/AssetProcessors/DetectAssetChangeProcessor.cs:989)
WallstopStudios.UnityHelpers.Editor.AssetProcessors.DetectAssetChangeProcessor:HandleAssetChanges (System.Collections.Generic.IReadOnlyList`1<string>,System.Collections.Generic.IReadOnlyList`1<string>,System.Collections.Generic.IReadOnlyList`1<string>,System.Collections.Generic.IReadOnlyList`1<string>) (at ./Library/PackageCache/com.wallstop-studios.unity-helpers@3.5.0/Editor/AssetProcessors/DetectAssetChangeProcessor.cs:440)
WallstopStudios.UnityHelpers.Editor.AssetProcessors.DetectAssetChangeProcessor:ProcessPendingAssetChangesCore () (at ./Library/PackageCache/com.wallstop-studios.unity-helpers@3.5.0/Editor/AssetProcessors/DetectAssetChangeProcessor.cs:397)
WallstopStudios.UnityHelpers.Editor.AssetProcessors.AssetPostprocessorDeferral:RunSafely (System.Action) (at ./Library/PackageCache/com.wallstop-studios.unity-helpers@3.5.0/Editor/AssetProcessors/AssetPostprocessorDeferral.cs:234)
WallstopStudios.UnityHelpers.Editor.AssetProcessors.AssetPostprocessorDeferral:DrainPending () (at ./Library/PackageCache/com.wallstop-studios.unity-helpers@3.5.0/Editor/AssetProcessors/AssetPostprocessorDeferral.cs:213)
WallstopStudios.UnityHelpers.Editor.AssetProcessors.AssetPostprocessorDeferral:DrainScheduled () (at ./Library/PackageCache/com.wallstop-studios.unity-helpers@3.5.0/Editor/AssetProcessors/AssetPostprocessorDeferral.cs:186)
UnityEditor.EditorApplication:Internal_CallDelayFunctions ()

### Steps to Reproduce

1. Use an asset postprocessor
2. Reload assets

### Expected Behavior

All good

### Actual Behavior

The above warning

### Severity

Cosmetic (visual/UX issue)

### Unity Version

Other

### Package Version

3.5.0

### Platform

- [x] Editor
- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Regression

### Additional Context

_No response_

### LINKED ISSUE #437: [Feature]: WallstopProto - Fuzz Testing
### Feature Description

We should fuzz test our proto implementation

### Use Case / Motivation

Security/correctness

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Unity Version

None

### Package Version

None

wallstop:
Fixes #280
Fixes #437

Four pieces of work, one session. Two of them are defects a shipped game can hit; one is a
credential path that was interrupting the owner; one is a CI lever measured and deliberately not
pulled.

## Credentials come from a cache, and never from a dialog

The Dev Containers credential helper answers by raising a dialog **on the owner's desktop**, once per
invocation. `.llm/context.md` and `ship-changes` each carried a copy-pasteable `git credential fill`
snippet, and `git push` / `git fetch` invoke the same helper on every operation — so the instructions
themselves were the defect. Session 191 raised four dialogs before being asked to stop.

`scripts/github-token.sh` is the only supported source now and never prompts: a non-empty
`$GITHUB_TOKEN` / `$GH_TOKEN`, else a 0600 cache, else exit 3 naming the command that fixes it. It is
also a git credential helper, and `normalize-container-git-config.sh` installs it as the **only** one
for github.com in the container's `~/.gitconfig`.

**The empty reset is the load-bearing half, not the add** — `credential.helper` is multi-valued and
git runs every value. The first version of the test could not have caught that: it asked
`git config --get-urlmatch`, which reports the URL-scoped section alone and answered "only our helper
is configured" for a config git would still have prompted from. It asks `git credential fill` itself
now, against a fake helper that records its own invocations.

Verified end to end: `git ls-remote` against this remote authenticates from the cache in **0 s, no
dialog**. The one deliberate prompt is `npm run github:token:bootstrap`, run by a human once per
container; `npm run github:token:store` takes a pasted token with no dialog at all.

A contract keeps the guidance from drifting back — no tracked file may carry a **runnable**
`git credential fill`. It distinguishes a code fence from prose, because the prohibition contains the
string, and it exempts a file that pins `GIT_CONFIG_SYSTEM` (a throwaway config cannot reach the
host's helper) rather than exempting a filename, which is the defect #445 is about.

## WallstopProto is fuzzed, and it found a hole in #434's rule (#437)

Four seeded strategies over every member shape — random bytes, mutations of valid payloads, generated
field sequences, generated capacity claims. The invariant is what a player needs from a save file it
did not write: the decode **terminates, allocates in proportion to what it was given, and reports
failure instead of throwing**, and anything accepted survives its own re-encoding.

It found a defect on unmutated `main`, in ten bytes: `0A080A06FFFFFFFF0700` throws
`OutOfMemoryException` out of `TryRead`.

#434's rule is that a dimension header must **equal** the element count it delivers — strictly
stronger than a ceiling, **except that a zero axis makes the product zero whatever the other axes
say**. `[int.MaxValue, 0]` delivers nothing, equals nothing, is accepted, and reaches
`new int[2147483647, 0]`, which the CLR refuses for the *shape* rather than for the elements
(measured: `[1<<30, 0]` allocates fine).

An axis is now refused as the capacity claim it is, via `SerializationCapacityLimits.TryAccept`. The
property that makes this safe: **it cannot refuse a shape that carries data.** With every axis at one
or more, each axis divides the product, so each is at most the delivered count — and `TryAccept`
raises its ceiling to exactly that.

*Proven able to fail*, which is the only thing that makes a fuzz suite worth its runtime:

| Mutation | Result |
| --- | --- |
| `TryConsume` stops bounding a run against the buffer | caught by three strategies independently |
| the axis refusal reverted | caught by the fuzz strategy **and** both named regression rows |
| `TryReadLength` stops checking length against `Remaining` | survived — **equivalent**, `TryConsume` enforces it again |

Two harness defects were found by measuring rather than reading, and both would have left it green
while proving less: an all-default contract encodes to **zero bytes**, so half the mutation seeds were
nothing at all (scalar reach 49.7% → past 90%); and **acceptance is the wrong coverage proxy**,
because a mutated rectangular payload is usually refused *deep*, by the very check #434 added.

## The watcher stops loading every imported prefab (#280)

Three sessions read this as a deferral bug. It never was — the reporter's own stack shows the work
already running from `EditorApplication.delayCall`, one tick after the import. Unity raises the
`Awake / CheckConsistency / OnValidate` guard around **every** `OnValidate` it runs, at any time, so
a deferred `LoadAllAssetsAtPath` produces the identical warning one tick later, from the consumer's
own components, with our frames underneath.

`GetMainAssetTypeAtPath` reports `GameObject` for a prefab, so a watcher on any other type fell
through to `HasMatchingSubAsset`, which loaded **every imported prefab** — once per watcher — and
discarded every object it loaded, because the whole decision is a `Type` predicate.

Nothing caught it because the code was *compliant with the rule as written*: the call was already
inside the drain, which was the only remedy the skill prescribed, and the reachability contract
deliberately stops its walk at the deferral boundary. The rule now reads **deferral is necessary, not
sufficient — never answer a metadata question with a load**, in the skill and in critical rule 16,
and the two comments that stated the wrong premise are corrected where they live.

And the existing regression test could not fail: its fixture has no user `OnValidate` at all, so it
passed with the bug fully present. The new one counts `OnValidate` invocations rather than warnings,
which proves the load **did not happen** rather than that it was quiet.

Watching prefabs by the components they contain is not supported behaviour (owner decision on the
issue); `SearchPrefabs` is the opt-in route, and it is opt-in because it pays exactly this cost.

## The CI lever that was measured and not pulled

`Stuck Job Watchdog` is the repository's most frequent workflow by run count, and it costs **$0.00** —
the repository is public. Billing rounds each job to a whole minute and the job takes **13 s**, so
removing its token step or its clone saves nothing; and GitHub already throttles the cadence, so the
cron requests 288 runs/day and delivers **65.5** (median gap **21.3 min**, not 5).

Of 82 runs whose logs were read: 77% clean queue, 23% evaluated a queued run and found it healthy,
**zero took any action**. Corroborated independently — the `watchdog-state` branch has exactly one
commit since 2026-06-28, and both recovery paths commit immediately. Left alone; the numbers are in
the progress log so the next look starts from data rather than from intuition.

## Verification

- `scripts/tests/test-github-token.sh` 20/20 and `test-normalize-container-git-config.sh` 12/12, with
  both mutations of the helper install caught and the baseline green after restore.
- `Generator~`: **374** cases against protobuf-net 3.2.56, **373** against 2.4.9 (362 baseline).
- All six `typecheck:unity` / `typecheck:tests` configurations.
- A real Unity 6000.4.6f1 editor via the MCP bridge: **65 of 65 package assemblies fresh, empty
  console.**
- `validate:tests:fast`, `lint:markdown`, `lint:docs`, `lint:changelog`, `lint:spelling`,
  `test:shell-portability` 18/18, `test:run-repo-lint` 11/11.

**Not verifiable locally:** the EditMode tests for #280. There is no Unity licence in this container
and the MCP command sandbox cannot reference package types (#435), so the bridge proves the tree
compiles in a real editor but cannot run a package test. The four-editor EditMode matrix on this pull
request is the signal for those.

## Review round

An adversarial pass over the first four commits found five things worth the round trip, all fixed
here and each verified by reverting the fix:

- **A credential leak introduced by the commit meant to prevent one.** The `url=` parser stripped
  userinfo before cutting the path, and `#*@` strips through the *first* `@` anywhere — so
  `https://evil.example/@github.com` read as host `github.com` and was served the token. Path first
  now, with four hostile and four legitimate rows covering it.
- **`erase` was unconditional**, and git calls it on any *rejected* credential for whatever host it
  was reaching — one unrelated 401 wiped the cache. And `write_cache` truncated rather than removed,
  which follows a symlink.
- **"Never prompts" was overstated**: with an empty cache git falls back to the editor's own
  `GIT_ASKPASS`, which a helper cannot suppress. The docs say so now.
- **`SearchPrefabs` does not do what three places said it did.** It searches prefabs for instances of
  the *handler's* type; it has no bearing on which assets match a watcher. The code comment, the
  CHANGELOG and the user docs all offered it as the route for watching prefab contents.
- **The strategy written for the axis fix exercised it once in 7,500 decodes** — it always emitted
  field 1, so the rank-three member (the shape #434's defect actually had) was unreachable. It emits
  every rectangular field now, and the found payload is a named regression case rather than
  iteration 42 of one seed.

---

> [!NOTE]
> **Medium Risk**
> Touches deserialization security (player saves/network) and editor import-time asset matching semantics (prefab sub-asset watching removed); credential changes affect all container git/GitHub access but are heavily tested.
> 
> **Overview**
> **GitHub in the devcontainer** no longer relies on `git credential fill` or the Dev Containers helper for routine work. **`scripts/github-token.sh`** resolves a token from env or a 0600 cache (exit 3 with fix instructions when empty), acts as the sole **github.com** credential helper via **`normalize-container-git-config.sh`**, and adds **`npm run github:token:*`** plus contract tests that block runnable `git credential fill` snippets from returning to agent docs.
> 
> **Asset change watching (#280)** stops answering type questions by loading **`.prefab`** files: **`HasMatchingSubAsset`** returns false for prefab paths so **`LoadAllAssetsAtPath`** no longer deserializes every component (and their **`OnValidate`**) on each import. Nested sub-assets in **`.asset`** files and sprite-on-texture behavior stay covered; docs and skills state that deferral alone does not make loads safe and that watching prefab *contents* is unsupported.
> 
> **WallstopProto rectangular arrays (#437)** refuse hostile **empty** shapes where a zero axis hides a huge other axis: **`WProtoRectangular.TryAcceptShape`** now takes **`largestDimension`** and applies **`SerializationCapacityLimits`**, with generator emission of **`LargerDimension`**; new **`FuzzTests`** and regression cases pin OOM-from-ten-bytes and re-encode invariants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d55a86f1d45e3f6c6bb839ce8faf224827015ce5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Generator~/WallstopStudios.UnityHelpers.Proto.Generator/RectangularArray.cs
```diff
@@ -523,10 +523,16 @@ private void EmitDimensionCapture(Writer writer, string source)
...
+        /// <para>
+        /// The largest axis is folded beside the product because the product stops bounding the axes
+        /// the moment one of them is zero: <c>[int.MaxValue, 0]</c> multiplies to zero, equals an
+        /// empty run, and reaches <c>new int[2147483647, 0]</c>.
+        /// </para>
         /// </remarks>
         private void EmitShapeCheck(Writer writer)
         {
             string product = "(long)" + Dimension(0);
+            string largest = Dimension(0);
```
wallstop: We get Dimension(0) twice, why not replace with one call?

scripts/github-token.sh
```diff
@@ -0,0 +1,286 @@
...
+# github.com and its siblings (gist., www., raw.) share one credential; anything else is not ours.
+is_github_host() {
+    local candidate
+    candidate="$(printf '%s' "$1" | tr '[:upper:]' '[:lower:]')"
+    candidate="${candidate%%:*}"
+    case "$candidate" in
+        github.com | *.github.com) return 0 ;;
+        *) return 1 ;;
+    esac
+}
```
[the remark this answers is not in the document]
wallstop: Correct, and it is worse than a gap: registering a URL **resets** the inherited helper for it, so a URL that is registered and then declined has nothing left to answer it — strictly worse than never registering it.

Fixed as a class rather than an instance in `d55a86f1`:

- the URL list now lives in `github-token.sh --hosts` and `normalize-container-git-config.sh` reads it, because two copies is how they came to disagree;
- `githubusercontent.com` and its subdomains join the served set, kept a superset of the registered URLs;
- a contract asserts that **every URL `--hosts` claims is one `get` actually serves**. Deleting the `githubusercontent` branch fails it, which is exactly the bug you reported.

Generator~/WallstopStudios.UnityHelpers.Proto.Generator/RectangularArray.cs
wallstop: Taken — `Dimension(0)` is read once into `firstDimension` and shared by the product and the largest-axis folds (`d55a86f1`).

PR #464, opened 2026-08-16
Refuse to conjure a singleton whose state is authored, and hold three converter lists to one

### LINKED ISSUE #321: [Feature]: On-Demand Singletons
### Feature Description

Control whether singletons are auto-created or not through attributes! Something like this:

```csharp
using System;

/// <summary>
/// Marks a <see cref="SingletonBehaviour{T}"/> that is allowed to be created on demand when no
/// instance exists in any loaded scene.
/// </summary>
/// <remarks>
/// Only valid for singletons that hold no serialized state, because a runtime-created instance is a
/// bare component with every [SerializeField] left null. Anything authored in a scene or prefab must
/// NOT carry this attribute - for those, a missing instance is a load-order bug and needs to be loud
/// rather than papered over with a stand-in that silently behaves like the real thing.
/// </remarks>
[AttributeUsage(AttributeTargets.Class, Inherited = true)] 
public sealed class RuntimeCreatableSingletonAttribute : Attribute
{
}
```

### Use Case / Motivation

Sometimes people don't want auto-loading singletons. If we don't have this kind of behavior, we should add it (or, make it configurable in other ways)

### Proposed Solution

The attribute thing, or an overridable property, or similar

### Alternatives Considered

_No response_

### Unity Version

None

### Package Version

None

### LINKED ISSUE #459: test: nothing stops a new write-only JSON converter, the way Touch and GameObject were added
## Context

Found while closing out the JSON half of #437 in #458.

`GameObjectConverter` and `TouchConverter` write a value and cannot rebuild it. That is a deliberate
one-way encoding — what they write is a diagnostic record — and #458 made them say so with
`NotSupportedException` instead of `NotImplementedException`, and declared them as write-only targets
in `JsonConverterFuzzTests`.

## The gap

**The declaration is a list in a test file, and nothing checks it against the converters.** A
forty-seventh converter can be added tomorrow with a `Read` that throws, or with a `Write` whose
output its own `Read` refuses, and:

- the fuzz suite will not notice, because a target is only fuzzed if someone added it to `Targets`;
- the write-only list will not notice, because it only constrains the names already on it.

That is exactly how the `WGuid.Empty` defect in #458 survived — a converter refusing its own output,
with a passing round-trip suite next to it, because every round-trip test used a non-empty id.

## What would close it

The WProto side already has the right shape in `ContractMirrorTests`: it parses `Runtime/` and fails
when a contract is neither mirrored nor listed with a stated reason. The JSON equivalent:

1. Enumerate every `JsonConverter<T>` / `JsonConverterFactory` registered in
   `SerializerEncoding.GetNormalJsonOptions()`.
2. Fail when one is absent from `JsonConverterFuzzTests.Targets` and absent from the write-only list.
3. Fail when a converter on the write-only list *does* read something back, and when one not on it
   refuses a value it just wrote.

Point 3 is the load-bearing one — it is the assertion that would have caught #458's worst defect
without anyone thinking to write a test for empty GUIDs.

## Not urgent, but not free either

Every converter is a save-file entry point, and the cost of missing one is a save that cannot be
loaded rather than a wrong pixel.

wallstop:
## Two halves of the package disagreed about what a missing singleton means

`ScriptableObjectSingleton<T>.Instance` returns `null` and logs a warning when no asset exists.
`RuntimeSingleton<T>.Instance` always created a `GameObject`. So a manager authored in a boot scene,
reached from any other scene, handed back a **bare component with every `[SerializeField]` left
null** — a stand-in that behaves like the real thing right up until it writes something.

`[SingletonCreation(SingletonCreationPolicy.NeverCreate)]` (#321) is that decision made explicit.

```csharp
[SingletonCreation(SingletonCreationPolicy.NeverCreate)]
public sealed class SaveManager : RuntimeSingleton<SaveManager>
{
    [SerializeField]
    private SaveSettings _settings;
}
```

Three design points, each a measurement rather than a preference:

- **The policy is read from the type, not from a virtual property.** There is no instance to ask when
  the question is whether to make one. That is why the owner's proposal on #321 was an attribute, and
  it is the only mechanism that can work.
- **The default does not move.** `CreateOnDemand` is what a singleton with no attribute gets, because
  auto-creation is a documented headline feature and flipping it would break every consumer.
  `NeverCreate` is a denial a consumer opts into.
- **`Instance` still never throws** (critical rule 13). It returns `null` and logs **one warning per
  type**, re-armed by `ClearInstance()`, gated to editor/development builds — the same shape
  `ScriptableObjectSingleton.WarnNoInstancesFound` already shipped. An unrecognized policy value
  creates rather than refuses: a singleton silently ceasing to exist is a worse answer to a value this
  build does not know about.

The policy governs **creation only**. An instance you authored is still found and still served.

### Two annotations that compile and do nothing are now reported

Editor-side, `[SingletonCreation]` on a type that is not a `RuntimeSingleton<>` is inert, and
`NeverCreate` paired with `[AutoLoadSingleton]` at any phase before `AfterSceneLoad` is a
contradiction — those phases run before a single `GameObject` exists, so a lookup that may not create
can only ever find nothing. `AfterSceneLoad` is exempt and is the control the refusal needs.

The rule takes its attributes as **arguments** rather than reading them off the type, so its eleven
tests drive every branch without annotating a deliberately wrong type — which `TypeCache` would find
on every editor load, and this very diagnostic would then report forever. That is the technique
`RuntimeMismatchSingleton` already uses for the auto-loader's own mismatch rules.

## The JSON converter list existed three times, and nothing compared the copies (#459)

`GetNormalJsonOptions`, `GetPrettyJsonOptions` and `GetFastJsonOptions` each carried a **verbatim
copy of the same 46-entry `Converters` initializer**. A forty-seventh converter added to one and
missed on another would have changed what a value round-trips through based only on which options the
caller reached for, with nothing to report it.

One list now. The enum converter is inserted at index 4, the position it occupied in normal and
pretty, so **every registration order is byte-identical to before** — verified by extracting all three
initializers from `main` and diffing them against the shared method's emission order.

That also makes `ColorBlockConverter`'s `#if !UNITY_DISABLE_UI` guard usable for the first time. With
three unguarded references, defining the symbol compiled the converter out and then broke the runtime
assembly on the references to it — a guard that could not be used was indistinguishable from
protection that existed. Session 195 filed this as too small to raise on its own; with one call site
it is three lines.

### The contract #459 asked for

`JsonConverterCoverageTests` holds the shipped registrations and the fuzz suite to each other:

| Contract | What it catches |
| --- | --- |
| `EveryRegisteredConverterIsCoveredDeclaredOrExplicitlyNotOurs` | A converter that ships and is fuzzed by nothing |
| `EveryDeclaredExceptionStillNamesARegisteredConverter` | A declaration naming a converter nobody registers |
| `TheWriteOnlyDeclarationAgreesWithTheFuzzTargets` | The two statements of "this one cannot read" drifting apart |
| `EveryShippedConfigurationRegistersTheSameConvertersInTheSameOrder` | The three lists diverging again, and pins the one deliberate difference |

It asks each converter what it claims through `CanConvert` rather than reading its generic argument,
because six of the registrations are factories whose converted type is decided per closure.

**And `AcceptedPayloadsSurviveTheirOwnReEncoding` could not make #459's load-bearing assertion.** It
`continue`s past any payload it fails to decode — correct, since most of its corpus is deliberately
hostile — so a converter whose **own seed** does not read back is silently passed over rather than
reported. That is exactly the shape of the `WGuid.Empty` defect #458 fixed. The new
`AReadableConverterReadsBackTheValueItWrote` makes it explicit, one case per target.

## Verification

**Local, in a live Unity 6000.4.6f1 editor against this working tree** (the reflection loop from
[unity-devcontainer-testing](.llm/skills/unity-devcontainer-testing.md)):

| Fixture | Result |
| --- | --- |
| `JsonConverterCoverageTests` | 4/4 |
| `JsonConverterFuzzTests` | 303/303 (was 256; +47 round-trip cases) |
| `SingletonCreationDiagnosticsTests` | 11/11 |
| `RuntimeSingleton` policy, driven directly | `NeverCreate` → null + the exact warning; `(SingletonCreationPolicy)200` and `CoroutineHandler` → created |

**Proven able to fail.** Five mutations, each caught by the test named for it:

| Mutation | Red |
| --- | --- |
| Drop `GameObject` from the write-only declaration | `EveryRegisteredConverterIsCoveredDeclaredOrExplicitlyNotOurs` |
| Make `Touch` a readable target | `TheWriteOnlyDeclarationAgreesWithTheFuzzTargets`, `AReadableConverterReadsBackTheValueItWrote`, `MalformedPayloadsAreReportedAsJsonException` |
| Register the enum converter in every configuration | `EveryShippedConfigurationRegistersTheSameConvertersInTheSameOrder` |
| Declare a type nothing registers | `EveryDeclaredExceptionStillNamesARegisteredConverter` |
| Remove the policy check from `Instance` | `NeverCreatedSingleton.Instance` created a stand-in — the exact assertion in `InstanceReturnsNullWhenCreationIsRefused` |

**No regression from the options refactor, measured rather than assumed.** The whole
`Tests.Serialization` namespace was run under one crude harness (no `[SetUp]`, so file-IO fixtures
throw) on both heads: `main` at `ed1396cf` reported **884 cases / 32 failed**, this branch **935 / 25**.
Every failure here is present on `main` by the same name and count — a strict subset. The seven that
disappear are fixture-ordering artifacts of the same harness, not signal.

`npm run typecheck:unity` green in all six configurations (default, define-off, Odin runtime, and the
three test variants). `npm run agent:preflight` and `npm run lint:changelog` green.

## A matching `grep` reported failure (found by this PR's own CI)

`Repo Lint` went red on this branch with

```text
FAIL excludeDirs missing 'obj' (.NET build output)
PASS excludeDirs contains 'bin' (binary build output)
```

Both entries are on **the same line of the same string**, and nothing about
`scripts/lint-meta-files.ps1` changed. The check was

```bash
echo "$SCRIPT_CONTENT" | grep '\$excludeDirs' | grep -qF "'obj'"
```

and under `set -o pipefail` that is a race: `grep -q` exits at its first match, the upstream `grep`
is still writing and dies of SIGPIPE, and pipefail reports the **pipeline** as 141 — so a successful
match reads as a failed check. Measured rather than reasoned about: with a producer large enough to
still be writing, the identical pipeline returns 141 on **30 of 30** runs while every line matches.
It passes whenever the producer finishes first, which is why 26 of the previous 27 `Repo Lint` runs
were green; eight concurrent workers over one tree (#446) is the pressure that finally lost it.

Each exclusion array is extracted once and answered with bash string matching now — no pipeline to
lose — and it still discriminates: deleting `'obj'` from `lint-meta-files.ps1` reddens it.

**The sweep fixed twelve more sites where a spurious 141 would change an answer** rather than be
swallowed: an assertion that flips its branch, or a `$(...)` that aborts under `set -e` — including
`scripts/unity/run-tests.sh`, which parses the Unity result counts CI reports. The ~40 remaining
sites feed the consumer from `echo`/`printf` of a small string, where the producer finishes inside
one write; those are filed on #465 with the lint rule that would close them, because the honest
remedy there is a rule rather than forty edits.

The sentence worth keeping: **a pipeline whose consumer may stop reading early does not report
whether it matched — it reports whether the producer got to finish.**

## The JSON contract had a hole of its own

It enumerated `JsonSerializerOptions.Converters`, and **three of this package's adapters register
through a type-level `[JsonConverter]` attribute instead** (`SerializableList<T>`,
`SerializableType`, `SerializableNullable<T>`), so they never appear in that list and were invisible
to both the contract and the fuzz corpus. `EveryTypeLevelConverterIsCoveredByAFuzzTarget` covers that
second channel, and the two uncovered types have seeds now. Neither had a defect — stated because a
new gate that finds nothing is worth saying plainly rather than dressing up.

Local: `JsonConverterCoverageTests` 5/5, `JsonConverterFuzzTests` **315/315**, and the full 60-check
`Repo Lint` registry green under the same 8-worker concurrency CI uses.

## Review round — five real findings, one of them a regression this PR introduced

**The `sed` range read the wrong array.** Replacing the racing pipeline with
`sed -n '/\$excludeDirs *=/,/)/p'` widened the assertion: sed never closes a range on its start
line, and `$excludeDirs` is declared on **one** line, so the range ran on to the closing paren of
`$excludeFilePatterns` fifteen lines later. Moving `'obj'` from one array to the other then *passed*
the check written to catch exactly that drift. One line is matched as one line now, and the
moved-entry case reddens it.

**`#if !UNITY_DISABLE_UI` is deleted rather than honoured.** Honouring it was the wrong half of the
choice session 195 left open. Three other runtime files — `Partials/UI.cs`,
`MatchColliderToSprite.cs`, `EnhancedImage.cs` — name `UnityEngine.UI` with no guard at all, so
defining the symbol still breaks the assembly; all the guard bought was turning one loud compile
error into a silent fall-through to System.Text.Json's reflection path, which is the thing these
converters exist to avoid under IL2CPP. A symbol nothing defines, that three sibling files cannot
honour, is protection that does not exist. The CHANGELOG entry claiming otherwise is gone with it.

**A `NeverCreate` singleton rescanned the whole scene on every access.** `CreateOnDemand` never had
the problem — it creates one and caches it — but with nothing in the scene `_instance` stays null and
`FindAnyObjectByType` repeats forever, which is precisely what the `if (X.Instance != null)` guard
this PR's own documentation recommends does per frame.

*The first fix was wrong, and probing rather than reasoning is what caught it.* The invariant is that
nothing can appear without `Awake` assigning the cache — and **Unity does not run `Awake` outside
play mode**. `AddComponent` in the editor produced a live, active instance that the remembered
refusal then hid. It is play-mode-only now; edit mode still scans, and the probe confirms all three
states.

**CS0414 in every release player build.** `_creationRefusedWarningLogged` was assigned and never read
when neither `UNITY_EDITOR` nor `DEVELOPMENT_BUILD` is defined. The repo's `TypeCheck` project
`NoWarn`s CS0414, so no local gate would have caught it — a consumer's build has no such suppression.
The field and its uses are inside the same `#if` now.

**The editor diagnostic could not see an inherited `[SingletonCreation]`.** `TypeCache` reports only
direct annotations while the runtime resolves with `inherit: true`, so an annotated abstract base
with a concrete `[AutoLoadSingleton]` subclass — the exact contradiction the rule reports — was
invisible: the base is skipped as abstract, and the subclass carries no attribute of its own. Both
attribute sets are enumerated now.

**And two tests read stronger than they were.** `Assert.IsTrue(restored != null)` is vacuous for
every struct target (`Deserialize` boxes) and never compared the value at all — it is a re-encoding
equality now, which works regardless of a target's equality semantics and catches a converter that
reads its own output back as a *different* value. And the order test compared the three lists only to
each other, which cannot see a reorder of the one shared list; the count and
`JsonStringEnumConverter`'s index are pinned against history instead.

Re-verified: **438/438** in a live editor (`JsonConverterCoverageTests` 5, `JsonConverterFuzzTests`
315, `JsonConverterTests` 107, `SingletonCreationDiagnosticsTests` 11), six `typecheck:unity`
configurations, and **60/60** `Repo Lint` checks under the same 8-worker concurrency CI uses.

## Second review round — owner feedback and Bugbot

**"Should this utilize `[Preserve]` or similar?"** Yes, and it generalizes to a class of eight. The
failure mode is the same for every member: the IL2CPP linker drops an attribute read only by
reflection, `GetCustomAttributes` finds nothing, and the reader falls back to its default —
`NeverCreate` silently resumes conjuring stand-ins, `[ScriptableSingletonPath]` looks in the wrong
folder. Nothing throws, and no editor run reproduces it.

`SingletonCreationAttribute`, `ScriptableSingletonPathAttribute`, `RandomGeneratorMetadataAttribute`,
`EnumDisplayNameAttribute`, `WShowIfAttribute`, and the three relational attributes
(`Child`/`Parent`/`SiblingComponentAttribute`, reached only through
`IsAttributeDefined<TAttribute>`).

*Stated plainly rather than overclaimed:* `link.xml` already preserves the whole runtime assembly, so
none of these is a live bug today. It earns its place because `link.xml`'s own comment calls that
wholesale preserve a "low-maintenance choice" — a later narrowing must not silently re-open it — and
because the marker belongs at the declaration rather than in a file the next reader will not check.

*And the sweep is executable.* `scripts/lint-preserve-attributes.js` (registered as
`preserve-attributes`) requires `[Preserve]` on every attribute used as a reflection target under
`Runtime/`, covering both the direct and the generic path, and is proven discriminating in both.
Two guards against itself: a scan that finds **nothing** fails rather than passes, and an attribute
reflected but declared outside `Runtime/` must be listed as foreign with its full name.

**Bugbot: "refusal cache hides pre-Awake instances" — right that it was unsound, wrong about why.**
Scene objects *are* findable before `Awake`, but a search that finds one caches it, so that path
never refuses. The actual hole is the one the edit-mode probe had already found wearing a different
disguise: the cache assumed only `Awake` can introduce an instance, and that is false in the editor
(Unity does not run `Awake` there) *and* false for a subclass whose override forgets `base.Awake()`.

The bound is **a frame** now, which needs no invariant at all. Worst case is one search per frame —
which is exactly what a single guarded call site already cost — and every stronger claim turned out
to be wrong somewhere.

*The typecheck caught the sweep's own defect*: two of those files keep usings inside
`#if UNITY_EDITOR`, so appending after the last one put `UnityEngine.Scripting` in a block the default
configuration compiles out.

Re-verified: **438/438** in a live editor, six `typecheck:unity` configurations, **61/61** `Repo Lint`
checks.

Closes #321. Closes #459. Files #465.

### DIFF
--- a/Runtime/Core/Attributes/SingletonCreationAttribute.cs
+++ b/Runtime/Core/Attributes/SingletonCreationAttribute.cs
@@ -4,6 +4,7 @@
 namespace WallstopStudios.UnityHelpers.Core.Attributes
 {
     using System;
+    using UnityEngine.Scripting;
     using WallstopStudios.UnityHelpers.Core.Helper;
 
     /// <summary>
@@ -35,6 +36,7 @@ namespace WallstopStudios.UnityHelpers.Core.Attributes
     /// </code>
     /// </example>
     [AttributeUsage(AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
+    [Preserve]
     public sealed class SingletonCreationAttribute : Attribute
     {
         /// <summary>
--- a/Runtime/Core/Attributes/SingletonCreationAttribute.cs
+++ b/Runtime/Core/Attributes/SingletonCreationAttribute.cs
@@ -1,0 +1,56 @@
+// MIT License - Copyright (c) 2026 wallstop
+// Full license text: https://github.com/wallstop/unity-helpers/blob/main/LICENSE
+
+namespace WallstopStudios.UnityHelpers.Core.Attributes
+{
+    using System;
+    using WallstopStudios.UnityHelpers.Core.Helper;
+
+    /// <summary>
+    /// Decides whether <see cref="WallstopStudios.UnityHelpers.Utils.RuntimeSingleton{T}.Instance"/>
+    /// may create the singleton when none exists in any loaded scene.
+    /// </summary>
+    /// <remarks>
+    /// A singleton authored in a scene or a prefab holds serialized state, and a created stand-in has
+    /// every <c>[SerializeField]</c> left at its default - so it behaves like the real thing while
+    /// carrying none of the data. Annotate those with
+    /// <see cref="SingletonCreationPolicy.NeverCreate"/> so a missing instance is reported instead of
+    /// substituted. Leave a singleton that holds no serialized state alone; creating one on demand is
+    /// what makes <c>Instance</c> work from anywhere.
+    ///
+    /// The decision is read from the type rather than from an overridable member because there is no
+    /// instance to ask when the question is whether to make one.
+    /// </remarks>
+    /// <example>
+    /// <code>
+    /// [SingletonCreation(SingletonCreationPolicy.NeverCreate)]
+    /// public sealed class SaveManager : RuntimeSingleton&lt;SaveManager&gt;
+    /// {
+    ///     [SerializeField]
+    ///     private SaveSettings _settings;
+    /// }
+    ///
+    /// // Returns null instead of a SaveManager with no settings.
+    /// SaveManager manager = SaveManager.Instance;
+    /// </code>
+    /// </example>
+    [AttributeUsage(AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
+    public sealed class SingletonCreationAttribute : Attribute
+    {
+        /// <summary>
+        /// Gets the policy <c>Instance</c> applies when no instance exists.
+        /// </summary>
+        public SingletonCreationPolicy Policy { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SingletonCreationAttribute"/> class.
+        /// </summary>
+        /// <param name="policy">
+        /// What <c>Instance</c> should do when no instance of the singleton exists.
+        /// </param>
+        public SingletonCreationAttribute(SingletonCreationPolicy policy)
+        {
+            Policy = policy;
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix defining `UNITY_DISABLE_UI` breaking the runtime assembly. The JSON options named `ColorBlockConverter` while the converter itself was compiled out ([#459](https://github.com/Ambiguous-Interactive/unity-helpers/issues/459)).
 - Fix a save holding an unset `WGuid` failing to load: JSON wrote the empty GUID and then refused to read it back ([#437](https://github.com/Ambiguous-Interactive/unity-helpers/issues/437)).
 - Fix a JSON payload with `min` greater than `max`, or with `max` missing, crashing a `Range<T>` load with `ArgumentException` instead of reporting corrupt data ([#437](https://github.com/Ambiguous-Interactive/unity-helpers/issues/437)).
 - Fix a `Gradient` with more than eight colour or alpha keys silently losing the extras and filling the player log with errors. The payload is now refused ([#437](https://github.com/Ambiguous-Interactive/unity-helpers/issues/437)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `[SingletonCreation(SingletonCreationPolicy.NeverCreate)]`, which stops `RuntimeSingleton<T>.Instance` creating a bare stand-in when no instance exists. It returns `null` and names the type instead. See [Controlling on-demand creation](./docs/features/utilities/singletons.md#controlling-on-demand-creation) ([#321](https://github.com/Ambiguous-Interactive/unity-helpers/issues/321)).
 - Add `AnimationClip.GetSpriteFramesFromClip()`, which pairs every sprite a clip references with the binding that supplies it, a `GetSpritesFromClip(path, propertyName, type)` overload that filters on that binding, and `UnityExtensions.SpriteBindingProperty`. Editor-only. See [Sprites from an AnimationClip](./docs/features/utilities/math-and-extensions.md#sprites-from-an-animationclip) ([#451](https://github.com/Ambiguous-Interactive/unity-helpers/issues/451)).
 - Add `AssetChangeDetectionUtility.Enabled`, `ResetEnabledToDefault()` and `EnabledScope(bool)` to turn the `[DetectAssetChanged]` watcher on and off. The returned `AssetChangeDetectionEnabledScope` restores the previous setting on dispose ([#327](https://github.com/Ambiguous-Interactive/unity-helpers/issues/327)).
 - Add `SingleThreadedThreadPool.DrainAsync()`, which closes the pool and waits for queued work to finish instead of dropping it, plus `IsAcceptingWork` ([#318](https://github.com/Ambiguous-Interactive/unity-helpers/issues/318)).
@@ -71,6 +72,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix defining `UNITY_DISABLE_UI` breaking the runtime assembly. The JSON options named `ColorBlockConverter` while the converter itself was compiled out ([#459](https://github.com/Ambiguous-Interactive/unity-helpers/issues/459)).
 - Fix a save holding an unset `WGuid` failing to load: JSON wrote the empty GUID and then refused to read it back ([#437](https://github.com/Ambiguous-Interactive/unity-helpers/issues/437)).
 - Fix a JSON payload with `min` greater than `max`, or with `max` missing, crashing a `Range<T>` load with `ArgumentException` instead of reporting corrupt data ([#437](https://github.com/Ambiguous-Interactive/unity-helpers/issues/437)).
 - Fix a `Gradient` with more than eight colour or alpha keys silently losing the extras and filling the player log with errors. The payload is now refused ([#437](https://github.com/Ambiguous-Interactive/unity-helpers/issues/437)).
--- a/Editor/Tags/AttributeMetadataCacheGenerator.cs
+++ b/Editor/Tags/AttributeMetadataCacheGenerator.cs
@@ -84,6 +84,15 @@ namespace WallstopStudios.UnityHelpers.Editor.Tags
 
                 AutoLoadSingletonEntry[] autoLoadEntries = BuildAutoLoadSingletonEntries();
 
+                foreach (
+                    string problem in FindSingletonCreationProblems(
+                        TypeCache.GetTypesWithAttribute<SingletonCreationAttribute>()
+                    )
+                )
+                {
+                    Debug.LogWarning(problem);
+                }
+
                 // Get or create the cache asset
                 AttributeMetadataCache cache = GetOrCreateCache(out bool metadataChanged);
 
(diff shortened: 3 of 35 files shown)


Runtime/Core/Attributes/SingletonCreationAttribute.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// {
+    ///     [SerializeField]
+    ///     private SaveSettings _settings;
+    /// }
+    ///
+    /// // Returns null instead of a SaveManager with no settings.
+    /// SaveManager manager = SaveManager.Instance;
+    /// </code>
+    /// </example>
+    [AttributeUsage(AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
```
wallstop, line 38: Should this utilize [Preserve] or similar?

wallstop: Yes — taken, and swept.

**The class is "an attribute this package reads by reflection at runtime"**, and the failure mode is
identical for every member: the IL2CPP linker drops the attribute, `GetCustomAttributes` finds
nothing, and the reader falls back to its default. `[SingletonCreation(NeverCreate)]` silently
resumes conjuring empty stand-ins; `[ScriptableSingletonPath]` looks for the asset in the wrong
folder. Nothing throws, and no editor run reproduces it.

Eight, found by scanning every reflection target under `Runtime/`:

| Attribute | Read by |
| --- | --- |
| `SingletonCreationAttribute` | `RuntimeSingleton.ResolveCreationPolicy` |
| `ScriptableSingletonPathAttribute` | `ScriptableObjectSingleton.GetResourcesPath` |
| `RandomGeneratorMetadataAttribute` | `RandomGeneratorMetadata` |
| `EnumDisplayNameAttribute` | `EnumExtensions` |
| `WShowIfAttribute` | `SerializableNullable.TryGetValueFieldAttribute` |
| `ChildComponentAttribute`, `ParentComponentAttribute`, `SiblingComponentAttribute` | `BaseRelationalComponentAttribute` via `IsAttributeDefined<TAttribute>` |

**One thing stated plainly rather than overclaimed:** `link.xml` already preserves the whole runtime
assembly, so none of these is a live bug today. It earns its place anyway — `link.xml`'s own comment
calls that wholesale preserve a *"low-maintenance choice"*, so a later narrowing must not silently
re-open it, and the marker belongs at the declaration where the next reader is looking rather than in
a file they will not think to check. (I nearly shipped the opposite mistake in this same PR — a
CHANGELOG entry claiming a `UNITY_DISABLE_UI` fix that was not one — so the framing is deliberate.)

**The sweep is executable rather than remembered.** `scripts/lint-preserve-attributes.js` (registered
in `Repo Lint` as `preserve-attributes`) requires `[Preserve]` on every attribute used as a
reflection target under `Runtime/`. It covers both paths — the three relational attributes are
reached *only* through a generic `TAttribute`, which no scan of call sites can see, so they are
required on the strength of their base. Proven discriminating both ways: removing either marker
reddens it, naming the attribute and its reader.

Two guards it needed against itself: a scan that finds **nothing** fails rather than passes (a broken
scan is indistinguishable from a clean repository otherwise), and an attribute that is reflected but
declared outside `Runtime/` must be listed as foreign with its full name rather than waved through.

The typecheck then caught the sweep's own defect — two of these files keep usings inside
`#if UNITY_EDITOR`, so appending after the last one put `UnityEngine.Scripting` in a block the
default configuration compiles out.

Issue #486, opened 2026-08-17
lint: the PowerShell linters spend ~70% of a 14-minute serial repo-lint on file enumeration and reads, not scanning


wallstop:
Measured while answering [#438](https://github.com/Ambiguous-Interactive/unity-helpers/issues/438), which asked whether a sibling repo's quadratic-scanner fix applied here. It did not — but the ablation technique that answered it pointed at this instead.

## The finding

`lint-no-regions` spends **28 ms** running its regex over all 21.6 M characters, inside a check that takes **50 s**. Essentially all of the rest is file enumeration and reading.

Stage timings over `Runtime` + `Editor` + `Tests` (1643 files / 21.59 M chars):

| Stage | 9p mount (devcontainer) | tmpfs |
| --- | ---: | ---: |
| `Get-ChildItem -Recurse -Include *.cs -Path $root` | **28.5 s** | 0.25 s |
| `Get-ChildItem -Recurse -File -Filter *.cs` | 4.6 s | 0.037 s |
| `[IO.Directory]::EnumerateFiles(…, AllDirectories)` | **0.82 s** | 0.005 s |
| `Get-Content -Raw` per file | **39.7 s** | 0.29 s |
| `[IO.File]::ReadAllText` per file | **3.4 s** | 0.061 s |
| the `#region` regex over all 21.6 MB | 28 ms | — |

`node scripts/run-repo-lint.js --jobs 1` is `real 13m54s` against `user 3m14s + sys 0m58s` — about **70% I/O wait**. `df -T` shows the repo on a `9p`/`v9fs` mount of the Windows host.

Two antipatterns, both mechanical to fix:

1. **`Get-ChildItem -Recurse -Include`** enumerates everything and post-filters, ~35x slower than `[IO.Directory]::EnumerateFiles` here. Six scripts: `lint-unity-file-naming.ps1:84`, `lint-no-regions.ps1:88`, `lint-odin-undo-safety.ps1:91`, `lint-tests.ps1:1305`, `lint-drawer-multiobject.ps1:169`, `lint-csharp-naming.ps1:122`.
2. **`Get-Content -Raw` per file**, ~12x slower than `[IO.File]::ReadAllText`. Eight scripts.

## The honest caveat, which is why this is scoped as it is

[...]

Issue #460, opened 2026-08-16
ci: copilot-pull-request-reviewer reports failure on every pull request, with no output


wallstop:
## What was measured

The `copilot-pull-request-reviewer` check run is `failure` on the head commit of every recent pull
request, including three that were reviewed and merged:

| PR | head | conclusion |
| --- | --- | --- |
| #458 | `46e422d3` | failure |
| #456 | `1add1dff` | failure |
| #454 | `298f2bfd` | failure |
| #447 | `d55a86f1` | failure |

The check run carries **no `output.title` and no `output.summary`** — there is nothing to read and
nothing to act on. It does not appear on `main`'s commits at all, only on pull requests.

## Why this is worth fixing rather than ignoring

A check that is red on every pull request is indistinguishable from a check that is red because of
the pull request. The working rule for the last several sessions has been "main CI green before
starting, PR CI green before finishing", and this one silently trains that rule to be read as "green
except the one that is always red" — which is the state in which a *real* reviewer failure goes
unnoticed.

It also costs a line of explanation in every session log and every PR hand-off.

## Options

1. **Fix it** — if Copilot code review is intended to run here, the failure has a cause worth knowing
   (permissions on the app installation, or the repository transfer to `Ambiguous-Interactive`).
2. **Turn it off** — disable Copilot pull request review for this repository, so the check stops
   being produced.

Either is better than the current state. This is not a request to *replace* it: the automatic bot
reviewers that do work are unaffected, and nobody should be manually requesting reviews from bots.

## How to reproduce

```bash
gh api "repos/Ambiguous-Interactive/unity-helpers/commits/<pr-head-sha>/check-runs" \
  --jq '.check_runs[] | select(.name|test("copilot")) | {name, conclusion, output}'
```

wallstop:
**Cause found, on #458.** The check run itself carries no output, but the bot posts the reason as a
pull request review:

> Copilot was unable to review this pull request because the user who requested the review has
> reached their quota limit.

So this is a **Copilot quota exhaustion**, not a permissions or transfer problem — which narrows the
options to two:

1. Raise or renew the quota on the account Copilot review is billed to.
2. Turn Copilot pull request review off for this repository, so the check stops being produced at
   all.

Option 2 is the one that costs nothing and removes the permanently-red check; option 1 is only worth
it if the review output is actually being read. Either way the current state — a red check on every
pull request whose reason is only visible if you go read the review body — is the one to leave.

wallstop:
[...]

PR #502, opened 2026-08-19
Construct a JSON converter where its closure lives, declare a hook where every reader runs it, and report the field Unity threw away

### LINKED ISSUE #437: [Feature]: WallstopProto - Fuzz Testing

### LINKED ISSUE #497: editor: Unity silently drops a framework generic in a serialized field, and nothing reports the loss
## What happens

Unity's serializer declines every type out of the framework assemblies, and it declines **silently**.
Measured in a live editor (6000.4.6f1) on a `ScriptableObject` holding
`public (int, float) frameworkPair;`:

```text
SerializedObject.FindProperty("frameworkPair")  ->  null
JsonUtility.ToJson(asset)                       ->  the field is absent entirely
```

Not an empty `SerializedProperty`, not a warning, not a console message. The field simply is not
there. Anything authored into it is gone on the next domain reload, and nothing anywhere says so.

`[Serializable]` is **not** the discriminator, which is what makes this hard to reason about from the
outside: `ValueTuple<int, float>` *is* `[Serializable]` and is still dropped.

## Why this is worth a diagnostic rather than a doc line

`SerializableValueTuple<T1, T2>` (#289, shipped in #496) closes the tuple case, but tuples were only
the instance someone happened to report. The rule is general — any framework generic in a serialized
field behaves this way — and the failure mode is the worst kind: **silent data loss at authoring
time**, discovered when a designer's values are missing from a build.

Candidates a developer would plausibly reach for, all silently dropped today:

| Field | Serialized |
| --- | --- |
| `(int, float)` | no |
| `KeyValuePair<string, int>` | no |
| `Nullable<T>` | no (this is why `SerializableNullable<T>` exists) |
| `Dictionary<K, V>` | no (this is why `SerializableDictionary<K, V>` exists) |
| `Tuple<T1, T2>` | no |

Three of those already have a package stand-in. The package therefore already *knows* the answer for
the common cases — it just has no way to tell anyone who did not find the docs first.

## Shapes this could take

1. **An editor-time validation pass** over `MonoBehaviour` / `ScriptableObject` fields, reporting a
   `[SerializeField]`/public field whose type Unity will not serialize, and naming the package
   stand-in where one exists. Closest to where the mistake is made. Fits the existing
   `[DetectAssetChanged]` / validation machinery.
2. **A Roslyn analyzer.** Catches it at compile time rather than on inspection, but "will Unity
   serialize this type" is a runtime question about Unity's native serializer, and encoding its rules
   statically risks false positives on generic user types, which Unity *does* serialize since 2020.
3. **A drawer-level notice** — the field is invisible to a drawer precisely because it is not
   serialized, so there is nothing to draw on. Probably not viable; recorded so it is not
   re-proposed.

Option 1 looks right: it can ask Unity rather than model it, and it can be advisory rather than a
build break.

## Not obviously free

Enumerating "every serialized field in the project" has a cost, and a check that fires on a type the
developer deliberately left unserialized (a runtime-only cache) is noise. Whatever ships should
probably be opt-in or at least suppressible per field, and should say **which stand-in to use**
rather than only that something is wrong — the fix is the useful half.

Split out of #289, whose tuple half is closed. Found in session 204.

### LINKED ISSUE #500: WallstopProto: a lifecycle hook declared on a base type is never called, and nothing says so
## What is unasked

`CollectHooks` walks `contract.GetMembers()`, so a lifecycle hook is only ever found on the contract
that declares it:

```csharp
private static Hooks CollectHooks(GeneratorExecutionContext context, INamedTypeSymbol contract)
{
    ...
    foreach (ISymbol symbol in contract.GetMembers())
```

So a `[WProtoAfterDeserialization]` (or `Before`, or either serialization hook) declared on a **base**
type is never called for a derived contract. It is not diagnosed either — `WPROTO005` fires for a hook
on a type with no `[WProtoContract]`, which a base in an `[WProtoInclude]` chain *does* have, so the
hook looks wired up from every angle and silently never runs.

Found while sweeping the generator's `GetMembers()` call sites after Bugbot caught the same shape in
`WPROTO033` (fixed in #496): an uninitialized allocation zeroes the whole object, so that diagnostic
had to walk the base chain. This one is the same question asked of hooks, and unlike that one the
answer is **not** obviously "walk the base".

## Why it is not obviously a bug

Three things have to be decided together, and only the first is a fact about this repository:

1. **What does protobuf-net do?** It is the oracle for every other lifecycle question here, and the
   package's whole contract is to agree with it. Whether it invokes an inherited
   `[ProtoAfterDeserialization]` on a subtype is measurable in about ten minutes with the differential
   harness in `Generator~/.../OracleDifferentialTests.cs` — and it should be measured for **each** of
   the four hooks separately, because session 202 found two of five seeded shapes disagreeing with the
   guess.
2. **How does it interact with `[WProtoInclude]`?** A subtype's members are written by the base's
   formatter with the subtype's under an include tag. If both base and subtype declare the same hook,
   "run each once" and "run the most derived" are different behaviours, and `WPROTO006` currently
   refuses two methods carrying the same hook attribute *on one type* — which says nothing about one
   each on two types in a chain.
3. **What about `constructAtEnd`?** A `[WProtoBeforeDeserialization]` on such a contract already runs
   after construction rather than before, because there is no earlier moment. An inherited one would
   inherit that caveat.

## Current exposure

Small but non-zero. One hook exists in the package today — `DotNetRandom`'s
`[WProtoAfterDeserialization]` — and it is declared on the concrete contract, so it runs. Nothing here
is broken. The exposure is a **consumer** who puts a hook on their own base class and gets silence.

## Suggested shape

1. Measure protobuf-net for all four hooks, inherited, with and without an include chain.
2. If it invokes them, walk the base chain in `CollectHooks` and pin the behaviour with a differential
   test per hook.
3. If it does not, add a diagnostic — a hook on a base of a contract, never called, is worth a warning
   rather than silence, and that is cheap either way.

Either outcome is better than the present state, where the answer depends on a `GetMembers()` call
nobody chose deliberately.

Found in session 204.

### LINKED ISSUE #501: serialization: JSON converter factories are not AOT-safe, so a consumer's own generic closure throws ExecutionEngineException on IL2CPP
## What happens

`Serializer.JsonStringify` / `JsonDeserialize` of a **generic** type throws on an IL2CPP player when
nothing in the build names the closed converter:

```text
System.ExecutionEngineException : Attempting to call method
'...SerializableValueTupleJsonConverter`2[[System.Int32],[System.Single]]::.ctor'
for which no ahead of time (AOT) code was generated.
```

Measured on Unity **2021.3** standalone IL2CPP, on the gated leg of #496.

## Why

System.Text.Json resolves a converter for a generic type through a `JsonConverterFactory`, whose
`CreateConverter` builds the closed converter reflectively:

```csharp
Type converter = typeof(SomeConverter<,>).MakeGenericType(arguments);
return (JsonConverter)Activator.CreateInstance(converter);
```

IL2CPP generates code only for closures it can see statically. `SomeConverter<int, float>` appears
nowhere in source, so the constructor it is asked to call does not exist in the player.

**This is not specific to any one type.** The package ships **eight** such factories:

| Factory |
| --- |
| `RangeConverterFactory` |
| `SerializableNullableJsonConverterFactory` |
| `SerializableListConverterFactory` |
| `SerializableSetConverterFactory` |
| `SerializableDictionaryConverterFactory` |
| `SerializableSortedDictionaryConverterFactory` |
| `DequeConverterFactory` |
| `CyclicBufferConverterFactory` |

They pass today because the closures they are exercised with — `Deque<int>`, `SerializableNullable<float>`
and so on — are referenced elsewhere in this package or its tests, which is what gives IL2CPP a
reason to compile them. **A consumer's own closure has no such reference.** `Deque<TheirStruct>` in a
JSON payload is the same crash, and nothing in the package or its tests would catch it, because the
type does not exist here.

That is exactly the shape #343 fixed for protobuf, one serializer over: WallstopProto exists because
protobuf-net's reflective model cannot run under IL2CPP, and the source generator emits per-closure
formatters so it does not have to. JSON never got the equivalent.

## Discovered rather than reasoned about

Found while giving `ValueTuple` an AOT protobuf formatter in #496. The obvious companion change — a
`[JsonConverter]` factory so JSON worked too — was written, shipped to CI, and **failed on the
standalone leg with the error above**. The JSON half was reverted; the protobuf half (a
`WProtoRootMarshal`) works and stayed. Tuples are simply the case where *no* closure is referenced
anywhere, so it fails immediately rather than waiting for a consumer to hit it.

## Options

1. **Generate the registrations.** The generator already scans closed constructions for
   `WProtoRootMarshal`; the same scan could emit `options.Converters.Add(new SomeConverter<int, float>())`
   for every closure of a registered factory. Exact, no reflection, and it reuses machinery that
   exists. Largest piece of work.
2. **AOT hint class.** A never-called method referencing `new SomeConverter<T1, T2>()` for the
   closures the package knows about. Cheap, and helps only closures the package can name — so it
   fixes this repository's tests and not a consumer's game. A fig leaf unless paired with (1).
3. **Document the limitation** and tell consumers to use protobuf in a player, or to register a
   concrete converter themselves. Honest, free, and leaves the failure in place.
4. **`link.xml` preservation** — worth checking whether it can force the closures, though it
   generally preserves types that exist rather than creating generic instantiations.

Recommend (1), with (3) shipped immediately since it costs nothing and the failure is live today.

## Reproducing

Serialize any generic type through the package's JSON options, closed over arguments that appear
nowhere else in the build, from an IL2CPP standalone player.

Found in session 204.

wallstop:
## What this changes

Seven issues touched. Everything below was measured rather than reasoned about, and three of the
measurements contradicted the issue that asked for them.

### #501 — a generic type's JSON converter is now constructed where its closure is written

A `JsonConverterFactory` builds its converter with `MakeGenericType` and `Activator.CreateInstance`. IL2CPP compiles a generic closure only when source references it, and **value types do not share a compiled body**, so `Deque<TheirStruct>` in a player had no converter to construct and the first save threw `ExecutionEngineException`. The package's own tests never caught it, because the closures they exercise are named right here -- which is exactly what gives IL2CPP a reason to compile them.

The generator now finds every closed construction of a type paired through `[assembly: WJsonConverter]` and constructs the converter in the assembly that wrote the closure. Every factory asks `WJsonConverterRegistry` before it reaches for reflection, so the reflective path stays as the editor and Mono fallback it was always safe on. Eleven pairings ship; `WALLSTOP_DISABLE_GENERATED_JSON_CONVERTERS` turns them off.

**An assembly that does not reference System.Text.Json gets no registrar at all.** A live editor found that one: the registration names a `JsonConverter`, and eighteen of this repository's own test assemblies -- every assembly definition with `overrideReferences` and no `System.Text.Json.dll` listed -- stopped compiling at once. Every `dotnet` gate passed, because every one of those compilations references it. `JsonRegistrarEmissionTests` pins it.

**Registering the container was necessary and not sufficient**, and only the gated IL2CPP leg could show it. On a 2021.3 player the registration worked -- execution reached `DequeConverter<sbyte>.Read` -- and then threw one level deeper on `ListOfTConverter<List<sbyte>, sbyte>::.ctor`, because every collection converter here read its payload with `Deserialize<List<T>>` and System.Text.Json resolves *that* through its own factory. They read and write elements one at a time now, which asks only for `T`'s converter. All four gated IL2CPP legs pass.

The remaining boundary is documented rather than hidden: this covers the package's own generic types, not System.Text.Json's reflective handling of a consumer's own `struct`, which reaches `ObjectDefaultConverter<T>` through the same call.

### #500 — a lifecycle hook is now declared where every reader runs it

The issue asked whether protobuf-net invokes an inherited hook. Measured on both oracles, and **they disagree with each other**, which is why this is a warning rather than a behaviour change:

| Hook placement | WallstopProto | protobuf-net 2.4.9 | protobuf-net 3.2.56 |
| --- | --- | --- | --- |
| On the root of the chain | runs once | runs once | runs once |
| On a subtype | runs, innermost first | runs, outermost first | **never runs** |
| On a type with no chain | runs once | runs once | runs once |

So the issue's premise is wrong for the case it names -- a hook on the root already runs under all three -- and the real divergence is the opposite one. `WPROTO034` says so and names the fix; `AbstractRandom` now owns the after-deserialization hook and dispatches to a `protected virtual` method `DotNetRandom` overrides.

`WPROTO033` claimed protobuf-net skips a hook on a `SkipConstructor` contract. **It does not** -- measured on both oracles, all four hooks -- and being a subtype was the real reason `DotNetRandom`'s hook never ran there. The message is corrected.

### #497 — the serialized field Unity threw away is now reported

Unity declines every framework type in a serialized field and says nothing. The validator does not model that rule: it constructs the type, wraps it in a `SerializedObject` and asks which fields arrived, so it cannot misreport the generic user types Unity has serialized since 2020. It names the package stand-in where there is one, in C# keywords so the suggestion can be pasted; `[NonSerialized]` silences a field, because that is already what it means.

Scoped to the selection rather than the project: a first draft walked every loaded assembly, ran the startup half of the whole project and stopped the editor responding. The probe host is now deactivated before a component is added, so no `Awake` runs at all.

### #437 — fuzz the values, not only the bytes

`FuzzTests` feeds hostile bytes to the reader; this feeds random valid values to the writer and asks that both serializers produce and accept the same wire. Equality is re-encoding rather than a per-contract comparer, so every member is covered without any being named. Two facts fell out, both deliberate design rather than defects: a repeated packable scalar is written packed here and one field per element by protobuf-net; and 2.4.9 omits an empty-string map key where 3.2.56 and this package write the explicit zero-length one.

### #486 — enumerate the files, do not enumerate everything and filter

Six linters swapped `Get-ChildItem -Recurse -Include` for `[IO.Directory]::EnumerateFiles`: **227 s to 148 s** over this repository on the devcontainer's 9p mount. Output is byte-identical over the clean tree *and* over a tree seeded with a `#region` and a mis-named member, so the paths that print file names were exercised rather than only the no-findings path; exit codes match too. The issue's caveat stands -- CI runners are not on a 9p mount, so this is a local win.

### #460 — answered on the issue, not closed here

The `copilot-pull-request-reviewer` check is GitHub's own hosted agent workflow (`dynamic/agents/...`), triggered by repository ruleset **15358326**, failing on `statusCode: 402, errorCode: "quota"`. Billing state, not code, so both remedies are the owner's. [RCA on the issue](https://github.com/Ambiguous-Interactive/unity-helpers/issues/460#issuecomment-5339410958). It is the one red check on this PR.

## Review round

Cursor Bugbot found two real issues in the #497 validator, both fixed:

- **A field on a nested `[Serializable]` type was never reached** -- the parent field still produces a property, so the inner loss was invisible. The walk now recurses, materializes one collection element before walking its element type (element fields exist only under `Array.data[0]`), refuses a type it is already inside, and stops at Unity's nesting limit.
- **A type that would not construct was counted as clean** -- a selection where every type failed reported that every field survived across zero types. Skipped types are now named, and their presence makes the report a warning.

## Evidence

- Generator suite green on **both** oracles: 448 tests on `v3`, 447 on `v2`.
- `typecheck:unity` green across all six configurations; the `validate:tests` aggregate green.
- Live Unity editor (6000.4.6f1): the new AOT fixture passed 3/3 with 59 converters registered, the existing JSON fixtures passed **155/155**, and the editor-side validator passed 6/6 -- the only local gate that compiles the Editor assemblies at all (#347).
- CI: all four editmode and all four playmode legs green, and **all four gated IL2CPP standalone legs green**, which is the gate #501 exists for and the one that caught the `ListOfTConverter` half.

The one red check is `copilot-pull-request-reviewer`, which is out of quota on every pull request in this repository (#460, RCA linked above).

Closes #501
Closes #500
Closes #497
Closes #437

---

> [!NOTE]
> **Medium Risk**
> Touches IL2CPP JSON and protobuf serialization startup paths and an editor tool that instantiates selected types; behavior is heavily tested but regressions would surface at player save/load or during editor scans.
> 
> **Overview**
> **Editor tooling** adds a selection-scoped **Validate Serialized Fields** command that probes `MonoBehaviour` / `ScriptableObject` types via `SerializedObject`, recursively walks nested serializable fields, skips `[SerializeReference]`, and reports dropped public/`[SerializeField]` members with package **stand-in** type names.
> 
> **JSON / IL2CPP** introduces `[assembly: WJsonConverter]`, `WJsonConverterRegistry`, and generator-emitted `WJsonGeneratedRegistrar` (only when System.Text.Json is referenced). Closed generic constructions found in source get **pre-built** converters; factories consult the registry before `MakeGenericType`. Collection converters switch to **`WJsonArray`** element-by-element I/O to avoid nested `ListOfTConverter` AOT failures. Built-in container pairings ship behind `WALLSTOP_DISABLE_GENERATED_JSON_CONVERTERS`.
> 
> **WallstopProto generator** extracts shared **`ClosureScan`** from marshal logic, adds **`JsonConverterMap`**, and new warnings **WPROTO034** (lifecycle hooks on `[WProtoInclude]` subtypes), **WPROTO035/036** (bad or duplicate JSON converter pairings). **`AbstractRandom`** moves after-deserialization work to the chain root via a protected virtual hook (**`DotNetRandom`** overrides).
> 
> **Tests** add differential **value fuzz** against protobuf-net (repeated, maps, polymorphism), **hook ordering** fixtures, JSON registrar emission tests, and validator fixture types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 727a42557f7f09b189ea515ab9efc3161ddf658c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Serialization/JsonConverters/WJsonArray.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        /// <summary>
+        /// Reads a JSON array into a list, or returns <c>null</c> for a JSON null.
+        /// </summary>
+        /// <typeparam name="T">The element type.</typeparam>
+        /// <param name="reader">Positioned on the array's first token.</param>
+        /// <param name="options">The options to read elements with.</param>
+        /// <param name="owner">The type being read, named in any exception.</param>
+        /// <returns>The elements, or <c>null</c>.</returns>
+        internal static List<T> ReadList<T>(
```
wallstop, line 49: This could take in a list ref/buffer such that the ReadArray utilizes Buffers (list) as a scratch space to avoid double allocation.

Issue #523, opened 2026-08-21
feature: a pool for Unity objects whose checked-out items outlive the scope that took them — `ObjectPool.Dispose` strands them, and every consumer re-derives the same three details


wallstop:
Proposed from downstream (IshoBoy, session 133), where three components independently built the same pool and **three of them got some part of it wrong**. Filing because the package already owns the neighbouring problem well and this is the one hole in it — but see *"Why this might not belong here"* at the bottom, which is a real question rather than a formality.

## The defect this is about

`UnityEngine.Pool.ObjectPool<T>.Dispose()` destroys what is **in** the pool. An item that is checked out and still doing something is not in it — and when the factory built a root `GameObject`, nothing else owns that item either. So disposing the pool:

1. leaves the checked-out object standing in the scene, with whatever tween or animation is driving it still running, and
2. leaves its eventual ending calling `Release` **into a disposed pool**.

Neither logs. The symptom shows up a level later as a stray object, or as a throw from a callback nobody is watching.

## Why it kept happening

The three call sites in that project:

| owner | in-flight drain | pool disposed |
| --- | --- | --- |
| `Tree` (pooled fall visuals + shadows) | yes, hand-rolled | yes |
| `Elevator` (pooled dust puffs) | **no** | yes |
| `BeachBallObserver` (pooled prop visuals) | **no** | **no** |

One of three had it right, and only because it had been bitten. The correct version is ~40 lines and **three of those lines are non-obvious**, which is the argument for one home rather than three:

- **The removal from the in-flight list is unconditional; the release is not.** An item destroyed out from under its flight is Unity-null but is still the entry sitting in the list. A guard that returns early on it leaks one dead reference per use — exactly the unbounded growth the list exists to prevent. `ReferenceEquals(x, null)` asks *"was anything handed in"*, `x == null` asks *"is it gone"*, and the two questions want different answers here.
- **A destroyed item must not go back into the pool.** Pooling a husk makes it the next thing `Get` hands out.
- **`Dispose` drains into a copy before destroying anything**, so a `Release` arriving from a destroyed object's own ending finds nothing to hand back twice — `ObjectPool<T>` throws on a double release.

## Why the package's existing pooling does not cover it

This is not a gap in `WallstopGenericPool<T>` — it is a different lifetime shape, and I think that is the interesting part.

[...]

Issue #522, opened 2026-08-21
test: 14 of the 16 protobuf-net surrogates have no byte-parity gate, and drift already shipped once


wallstop:
`ProtobufUnitySurrogates.cs` declares **16 surrogates**. Only **two** of them are covered by a
byte-for-byte parity assertion against protobuf-net.

`WProtoFacadeTests.APortedTypeIsServedAndMatchesProtobufNetByteForByte` gates `FastVector2Int` and
`FastVector3Int`. The other fourteen have no such gate:

```text
Vector2  Vector3  Quaternion  Color  Color32  Rect  RectInt  Bounds
BoundsInt  Vector2Int  Vector3Int  Resolution  Parabola  ImmutableBitSet
```

## Why this is worth closing, with a concrete instance

A surrogate and a WallstopProto formatter are two independent encoders for one type, and nothing but a
parity test stops them drifting. Session 211 hit exactly that drift, and the gate is the only reason it
was caught:

`FastVector2IntSurrogate` mirrored the serialized cached hash by reading **`v.GetHashCode()`**. That was
identical to the stored `_hash` field until #309's fix made `GetHashCode()` answer the origin's hash for
a zero-initialized instance. From that moment protobuf-net wrote **6 bytes** for
`default(FastVector2Int)` while the formatter wrote **0** — a silent wire divergence between the two
encoders for the single most common value in grid code.

```text
FastVector2Int  Expected is <System.Byte[6]>, actual is <System.Byte[0]>
                Values differ at index [0]  Missing: < 24, 205, 175... >
```

It was caught **only** because `FastVector2Int` is one of the two gated types. Had it been `Vector2Int`
or `Bounds`, it would have shipped.

Fixed in PR #520 by giving both fast vectors an `internal SerializedHash` that reads the stored field,
and by extending that one test to cover `default` and the zero construction for both ranks.

## The obstacle, so nobody rediscovers it

`AssertServedAndIdentical<T>` asserts `WProtoFacade.TrySerialize(value, out byte[])` first. A type served
through `[assembly: WProtoSurrogate(...)]` is **not** registered as its own `IWProtoFormatter<T>`, so that
call answers `false` and the assertion fails for a reason that has nothing to do with byte parity. This
is the same trap recorded on #399: `IsRegistered<T>()` is not "is this type supported" — a surrogated or
root-marshalled type answers `false` while being perfectly well served.

[...]

Issue #521, opened 2026-08-21
random: UnityRandom is the one generator whose snapshot cannot resume its stream


wallstop:
Measured while characterising every generator's period algorithmically (#516). **19 of the 20 generators
resume their exact stream from their own `RandomState` snapshot. `UnityRandom` does not — it diverges at
the very first draw.**

## The measurement

Live Unity `6000.4.6f1`. For each generator: construct, draw 37 (so the snapshot is taken mid-stream, the
way a save actually happens), snapshot `InternalState`, draw 64 more, then rebuild through the
`RandomState` constructor and draw 64 and compare.

```text
RESTORES CORRECTLY (19):
  BlastCircuitRandom, DotNetRandom, FlurryBurstRandom, IllusionFlow,
  LinearCongruentialGenerator, PcgRandom, PhotonSpinRandom, RomuDuo, SplitMix64,
  SquirrelRandom, StormDropRandom, SystemRandom, WaveSplatRandom, WDoomRandom,
  WyRandom, XoroShiroRandom, XorShiftRandom, Xoshiro128StarStar, Xoshiro256StarStar

DOES NOT RESTORE (1):
  UnityRandom: diverges at draw 0 of 64
```

An independent diagnostic in the same run agrees: counting the state bits ever *observed to change*
across 512 draws gives `UnityRandom` **2^0** — no bit of its snapshot ever moves — against 2^32 to 2^32832
for everything else.

## Why

`UnityRandom.InternalState` builds its snapshot from `_seed`, the value the instance was constructed with:

```csharp
return new RandomState(
    (ulong)(_seed ?? 0),
    gaussian: _seed != null ? 0.0 : null,
    ...
```

The live position belongs to `UnityEngine.Random`'s engine globals, not to this object, so the snapshot
records where the stream *started* — and for a parameterless instance, not even that. The
`UnityRandom(RandomState)` constructor recovers `_seed` and calls `RestoreCommonState`, and does not call
`InitState` at all, so the restored instance simply continues from wherever the engine happens to be.

## Why it matters more than the existing "global state" caveat

The type's docs already warn that "determinism depends on controlling Unity's global state elsewhere in
your project". A reader reasonably concludes that if nothing else touches `UnityEngine.Random`, a save
and load will reproduce the sequence. **It will not**, and this is the worst shape a determinism bug can
take: the load *succeeds*, nothing throws, and the run is silently different.

[...]

Issue #513, opened 2026-08-21
feature: no `VisualElement` extensions exist, so every UI Toolkit consumer re-writes "is this drawn" and "is this inside that"


wallstop:
Raised in review on [IshoBoy PR #606](https://github.com/Ambiguous-Interactive/IshoBoy/pull/606):
*"Can we create an issue upstream (unity helpers) to add these methods, and/or see if we already
have them there?"*

## We already looked, and there are none

```
$ rg -n "this VisualElement" Runtime/
(no matches)
$ rg -l "VisualElement" Runtime/
Runtime/Visuals/UIToolkit/MultiFileSelectorElement.cs
Runtime/Visuals/UIToolkit/LayeredImage.cs
```

The package ships two UI Toolkit **elements** and no extension surface for
`UnityEngine.UIElements` at all — `Runtime/Core/Extension/` has `Animator`, `AsyncOperation`,
`Circle`, `Color`, `Dictionary`, `Direction`, `Enum`, `IEnumerable`, `IList`, `IReadonlyList`,
`Vector`… and nothing for the element tree.

## The two we wrote, and why they are not one-liners

Both came out of a focus bug: menu screens are swapped by assigning `style.display`, and the
element holding keyboard focus goes `DisplayStyle.None` while it still holds it — leaving the
cursor inside something nobody can see.

**1. "Is this element actually drawn?"**

```csharp
internal static bool IsShown(VisualElement element)
{
    for (VisualElement walked = element; walked != null; walked = walked.hierarchy.parent)
    {
        if (walked.style.display.value == DisplayStyle.None)
        {
            return false;
        }
    }
    return true;
}
```

Three things a caller gets wrong here:

- **it has to walk to the root.** An element with `display: Flex` under a hidden ancestor is not
  drawn, and asking only the element answers `true`;
- **`hierarchy.parent`, not `parent`** — the latter skips content containers, which is a different
  tree from the one display inherits down;
- **inline style rather than `resolvedStyle`.** `resolvedStyle` needs a layout pass, so a tree that
  has just been shown has not got one yet and answers with the previous frame's value. An unset
  `style.display` reads as `DisplayStyle.Flex`, which is the right default — so the naive version
  works only for elements somebody explicitly styled.

A `resolvedStyle`-based overload (`IsShownResolved`?) is the honest companion for callers who *are*
past layout and want USS taken into account. Both are defensible; having neither means every
consumer picks one without knowing there was a choice.

[...]

PR #525, opened 2026-08-21
Draw one entry not four, resume the stream a snapshot promised, and gate every surrogate

### LINKED ISSUE #524: [Feature]: WDoomRandom serves int table + higher period
### Feature Description

We should make WDoomRandom store ints and have a higher period, like 1024 or 2048, so that it isn't absolutely atrocious

### Use Case / Motivation

Slightly better randoms

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Unity Version

_No response_

### Package Version

_No response_

wallstop:
Seven issues. Five receive code, two were verified delivered and closed, and one receives a recorded negative result.

## #524 — `WDoomRandom` draws one entry instead of a quarter of one

The table held bytes and `NextUint()` consumed four, so a caller saw **64** values before the sequence repeated. It now walks 1024 whole `uint`s, one per draw: **16x the period at a quarter of the work**, with distinct entries so a cycle still emits no value twice.

PractRand 0.95 was built and run locally rather than the failure depth being guessed — clean at 4KB, `Gap-16` FAIL at 8KB, which is one period (4KB) doubled. `expected-outcomes.json` records `8KB`/`Gap-16` against the previous `2KB`/`BCFN`.

`WDoomRandom` has never shipped in a release, so this folded into its `Added` entry and the session-211 `Changed` entry correcting its period was removed rather than amended.

## #521 — the one generator whose snapshot could not resume

`UnityRandom.InternalState` carried the seed it was constructed with. The live position belongs to `UnityEngine.Random`'s globals, so a save loaded without error and ran a different sequence.

The issue left the tradeoff open deliberately, because restoring writes a **global** as a side effect of deserializing an object. **Decision: restore**, for three reasons — the type's identity *is* that global; `new UnityRandom(seed)` already calls `InitState`, so construction moving the engine is existing documented behaviour; and the opt-in alternative buys nothing for the case that motivates the issue, since a game saving an `IRandom` cannot know it holds a `UnityRandom`.

The position travels as `JsonUtility` text of `UnityEngine.Random.State`, through `RandomState`'s payload for JSON and a new `[ProtoMember(7)]` for protobuf. The protobuf half needed `AbstractRandom.OnBeforeSerialization`, the mirror of the `OnAfterDeserialization` already there, because a hook on a subtype of a `[WProtoInclude]` chain runs under some readers and not others (WPROTO034).

`GeneratorSnapshotRestoreTests` no longer has an exception list. It was written to name `UnityRandom` "so that the day it gains a real snapshot this test fails and says so"; that day arrived.

## #522 — the gate covered 2 of 16, and its premise was wrong in two places

Every `[WProtoSurrogate]` registration is now compared at three values, with the covered set derived from the assembly attributes so a fifteenth surrogate fails the completeness test the day it is registered.

Two corrections the measurement forced, both worth more than the test:

- **`ImmutableBitSet` is not byte-identical to protobuf-net and must not be.** It is the one surrogate holding a repeated scalar, and this package writes a packed run where protobuf-net writes a field key per element — deliberate, documented, roughly half the bytes. The gate would have "found" a design decision. So the claim asserted for all fourteen is the stronger one: **each encoder reads what the other wrote**, compared by re-encoding rather than by value, because a float member holding NaN is never equal to itself.
- **The facade is the wrong door.** `WProtoFacade.TrySerialize` answers `false` for a surrogated type while serving it perfectly well as a member — the `IsRegistered<T>()` trap #399 recorded.

Red-green: flipping one `WProtoMember` number on `ColorSurrogate` reds both gates and shows the alpha channel silently dropped in both directions. Reverted, green again.

## #523 — `TrackedObjectPool<T>`

A pool for `UnityEngine.Object` items whose lifetime ends in a callback rather than a scope. It deliberately does **not** build on `WallstopGenericPool<T>`: that pool's `PooledResource<T>` lease *is* a lexical scope, which is right for the scratch buffers it serves and cannot strand anything — and it cannot refuse a return, which is exactly what a destroyed Unity object needs.

The three lines that are easy to get wrong are the three the tests name: removal from tracking is unconditional while re-pooling is not, a husk is never pooled or handed out, and `Dispose` drains before destroying so a `Release` from a destroyed item's own ending is refused rather than counted twice.

## #486 — nine linters were paying for reads, not scans

The enumeration half of this issue had already landed; the read half had not.

| linter | before | after | |
| --- | ---: | ---: | ---: |
| `lint-license-headers` | 19.9s | 4.2s | 4.75x |
| `lint-no-regions` | 18.5s | 4.5s | 4.15x |
| `lint-unity-file-naming` | 20.2s | 5.6s | 3.58x |
| `lint-conditional-call-chains` | 8.2s | 2.7s | 3.03x |
| `lint-duplicate-usings` | 26.4s | 9.1s | 2.90x |
| `lint-csharp-naming` | 30.7s | 16.6s | 1.84x |
| `lint-tests` | 90.5s | 78.0s | 1.16x |
| **total (9)** | **216.6s** | **122.8s** | **1.76x** |

Findings identical, asserted rather than assumed: stdout, stderr and exit code compared before and after over the whole tree, and again over a tree seeded with a `#region`, a duplicate `using`, a missing licence header and two underscore names — so six of the nine were actually reporting rather than printing "clean" twice. All byte-identical both times. The 146 existing harness cases pass, plus two new ones for the only real semantic difference: an empty file and a single-line file, where `Get-Content` answers `$null` and a bare string against `ReadAllLines`' empty array.

A local developer-experience win, not a CI one — on ext4 the same swaps are worth about half a second.

## Verified delivered and closed

**#512** and **#513** both received code in #518 and neither had been closed. Verified against `main` before closing.

**#511** receives a negative result: the Wayback Machine holds a snapshot of the original `photonspin` repository root but not its source, so the constant stays unsettled on primary evidence and the determination stands.

## Verification

- Unity `6000.4.6f1` through the MCP bridge: the random fixtures (50 tests), the serialization sweep (50 fixtures, 526 passed), and the new pool fixture (9 tests). The three remaining sweep failures are `[OneTimeTearDown]` gaps in the ad-hoc loop, not in the tree — each passes in isolation, and the traps are now recorded in `unity-devcontainer-testing.md`.
- `Generator~` differentials in Release against protobuf-net 3.2.56 (465 passed) and 2.4.9 (464 passed).
- `typecheck:unity` clean across all configurations, `mkdocs build --strict`, `lint:docs`, `lint:markdown`, `lint:spelling`, `lint:changelog`, `agent:preflight`, `validate:prepush`.

Closes #524, #521, #522, #523, #486.

---

> [!NOTE]
> **Medium Risk**
> Serialization and RNG restore now write UnityEngine.Random.state and add a protobuf field; spatial-hash bucket order can change. Not auth/security-critical, but save-file and query-order behavior is user-visible.
> 
> **Overview**
> Adds **`TrackedObjectPool<T>`** for `UnityEngine.Object` items whose lifetime ends in a callback, not a scope. `Dispose` destroys in-flight items; destroyed objects are never pooled or handed out. Distinguishes `ReferenceEquals` vs Unity `== null` so a husk is still removed from tracking.
> 
> **`UnityRandom` snapshots now resume.** `InternalState` carries `UnityEngine.Random.state` (JsonUtility text). Restore writes that global back. Malformed or pre-3.6 payloads leave the engine alone. `AbstractRandom` gains `OnBeforeSerialization` so protobuf captures the position. Snapshot-restore tests no longer exempt this generator.
> 
> **`WDoomRandom`** walks 1024 distinct `uint`s (one `NextUint` per entry) instead of packing four table bytes. Period is 1024 draws; `NextTableByte` is gone.
> 
> Spatial-hash 2D/3D buckets remove with **`RemoveAtSwapBack`** (query order within a cell can change). New **`WProtoSurrogateParityTests`** compare every registered surrogate with protobuf-net (byte match except packed `ImmutableBitSet`; both encoders still round-trip). Linters read files via `File.ReadAllLines`/`ReadAllText` instead of `Get-Content`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e4081229ee3ec593b7865522b81a9880d36d5e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Random/UnityRandom.cs
```diff
@@ -119,8 +138,63 @@ public override uint NextUint()
...
+                UnityEngine.Random.state =
+                    UnityEngine.JsonUtility.FromJson<UnityEngine.Random.State>(_engineState);
+            }
+            catch (ArgumentException)
+            {
+                // A payload whose text does not describe an engine state leaves the engine where it
+                // is. The rest of the generator still works, which is what a save file written by a
+                // different version needs.
+            }
+        }
```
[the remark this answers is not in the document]
wallstop: Confirmed and fixed in 2cbb8bbb -- and the measurement says it is worse than reported.

Measured in Unity 6000.4.6f1:

```text
JsonUtility.FromJson<Random.State>("{\"foo\":1}")  ->  {"s0":0,"s1":0,"s2":0,"s3":0}
after assigning it, three consecutive draws       ->  2147483648, 2147483648, 2147483648
```

So the global stream is not rewound, it is **stuck**: an all-zero xorshift state emits one constant forever, for every consumer of `UnityEngine.Random` in the project. `catch (ArgumentException)` never fired, exactly as you said, because the text was valid JSON.

The fix is a round trip rather than a wider catch: re-serialize what was parsed and compare it to the payload. Foreign JSON parses to zeros and re-serializes to something the payload does not say, so it is refused; a genuinely zeroed state is refused outright, because no live engine produces one. Neither check knows how many fields `Random.State` has -- the zeroed shape is computed from the type rather than written out.

Four cases now pin it: foreign keys, an empty object, an explicitly zeroed state, and text that is not JSON.

Runtime/Utils/TrackedObjectPool.cs
```diff
@@ -0,0 +1,265 @@
...
+
+            int index = IndexOfInFlight(taken);
+            if (index < 0)
+            {
+                return false;
+            }
+
+            // Unconditional, and this is the line that is easy to get wrong: a destroyed item is
+            // still the entry in this list, and skipping the removal leaks it forever.
+            _inFlight.RemoveAt(index);
```
wallstop: Should this RemoveAtSwapBack (for perf)?

Runtime/Utils/TrackedObjectPool.cs
```diff
@@ -0,0 +1,265 @@
...
+
+            if (taken == null)
```
wallstop: What's special about null? What about value types? Unity engine types?

Runtime/Utils/TrackedObjectPool.cs
wallstop: Yes — done in 6e408122, using the package's own `IListExtensions.RemoveAtSwapBack`.

Nothing reads this list in order: `Dispose` drains all of it and `IndexOfInFlight` scans all of it, so the shift was pure cost.

I swept the class rather than the line. 58 `RemoveAt` call sites in `Runtime/`, of which **two more qualify**: `SpatialHash2D` and `SpatialHash3D` remove from a bucket, which is an unordered set, on the path that runs whenever anything moves. Both loops return immediately after the removal, and the only order-sensitive assertion in `SpatialHashTests` is a single-element query, so nothing observes it.

Everything else deliberately stays as `RemoveAt`, because the order is the meaning: `_colorStack` and `_materialStack`, a ring buffer's contents, the breadcrumb bar, the purge queue ordered by return time, and the convex/concave hulls. The many `RemoveAt(Count - 1)` calls already shift nothing.

The rule is now in `high-performance-csharp.md` with both preconditions (no caller observes the order; the index is not already the last) and the forward-loop trap, so the next change does not have to rediscover which lists are which. A `Changed` entry records that two items in one spatial-hash cell can come back in a different order than before — they were never ordered, but they were consistently unordered.

Runtime/Utils/TrackedObjectPool.cs
wallstop: Good catch — the line was doing something the operator does not say, and it is now named.

**What's special about null:** nothing, and that was the bug in the writing. Two *different* questions were being asked with two operators that both read as null checks:

- `ReferenceEquals(x, null)` — *was anything handed in*
- `x == null` — *is it gone*, which for a `UnityEngine.Object` is true for a **destroyed** instance as well as for a null reference

This type needs both, and the difference is the whole reason it exists: an item destroyed while checked out is still the entry in the tracking list, so the removal must not be guarded by `== null` while re-pooling must be. Guard the wrong one and you leak one dead reference per use.

Both are now `WasHandedIn` and `IsGone`, so no call site depends on a reader knowing which operator lies.

**Value types:** there is no case. `where T : UnityEngine.Object` makes `T` a reference type, and that constraint is exactly what makes the distinction expressible — the type's remarks now say so, and point at `WallstopGenericPool<T>` for the plain-object, scope-lifetime case.

**Unity engine types:** every `T` here is one by constraint, and `GameObject`, a `MonoBehaviour` subclass and `Texture2D` all follow the same rules — the fixture uses `Texture2D` precisely because the lifetime is identical and it needs no scene.

`create-csharp-file.md` rule 7 previously read as a flat ban on `ReferenceEquals`, which would have told the next agent to "fix" this pool back into leaking. It now carries the one legitimate use and the requirement to name it.

PR #528, opened 2026-08-22
Shrink grid payloads by 58%, and stop a CI test flaking

### LINKED ISSUE #386: research: Investigate if this library does string operations more correctly than ours
https://github.com/Antyss77/StringExtension

It has lots of different casing utilities, it's unclear if they are more or less correct (or performant) than our approaches

### LINKED ISSUE #519: perf: FastVector2Int/3Int put their derived hash on the wire, 43% of every cell
Measured while auditing #309. `FastVector2Int` and `FastVector3Int` put their **cached hash on the wire**,
and it is the largest field in the message.

## The measurement

Live Unity `6000.4.6f1`, current `main`, through `Serializer.ProtoSerialize`:

```text
(5, 3)  ->  08 05 10 03 18 F9 F0 A8 B3 05        10 bytes
            ^^^^^ x=5  ^^^^^ y=3  ^^^^^^^^^^^^^^ field 3 = the cached hash, 6 of the 10 bytes
```

A 1,000-cell `List<FastVector2Int>` payload is **14,167 bytes, 14.17 per cell**. The hash is a
well-distributed 32-bit value, so it is almost always a **5-byte varint plus a 1-byte tag**: about
**43% of every cell**, and it is *derived data* — a pure function of `x` and `y`.

A tilemap of 10,000 cells therefore carries roughly **60 KB of recomputable bytes**. `FastVector3Int`
is the same shape with one more component.

## It is already ignored on read

`FastVector2IntWProtoFormatter.TryRead` reads field 3 and **discards it**, rebuilding through the public
constructor so the hash is recomputed from `x` and `y`. Its own remark says why: a tampered payload must
not produce a value whose `GetHashCode` disagrees with its `Equals`. So nothing consumes the field; it is
written purely so the encoding stays byte-for-byte identical to what protobuf-net writes, which it must
while `[ProtoMember(3)]` is on the field.

## The decision this needs

Removing `[ProtoMember(3)]` from `_hash` makes both encoders agree on a two-field message and drops
~43% of the payload for the most common gameplay type in the package. The cost is a **wire-format
change**:

- **Reading old data keeps working.** Field 3 becomes an unknown field and `TrySkipField` already handles
  it, and the hash is recomputed either way.
- **Old readers cannot read new data.** A 3.5.1 reader handed a two-field payload gets `_hash == 0`. Note
  that this direction was *already* broken before the #309 fix, since a zero cache did not equal any
  constructed value; after that fix a `_hash` of 0 means "the origin", which is wrong for every other cell.
- **The generated path needs the same treatment.** The generator materializes a readonly struct through a
  private constructor that assigns only `[ProtoMember]` members, so with field 3 gone `_hash` would be left
  at 0 unless the hand-written formatter keeps ownership of this type, or the generator learns to
  reconstruct through the author's constructor.

That last point is the interesting one, and it generalizes past these two types: **a contract cannot
currently declare a member as derived** — recomputed on read rather than carried. Every such member is
either on the wire or silently zero.

## Suggested scope

- [ ] Decide whether the 43% is worth a wire break for `FastVector2Int` / `FastVector3Int`, or whether the
      compatibility is worth more.
- [ ] If yes: a `[WProtoDerived]`-style member marker, so the generator emits a reconstruction through the
      author's constructor instead of an assignment, and the property generalizes beyond these two types.
- [ ] Either way: a size assertion on the per-cell payload, so this is a number a test holds rather than
      one someone re-measures.

Relates to #504 (strictly beating protobuf-net, of which payload size is a part), #343 and #399.

### LINKED ISSUE #526: ci: the watchdog cleanup test's delayed-registration ordering assertion is a race, and it flakes
`test:shell-portability` H1-H3 failed once on PR #525 with

```text
FAIL Unity watchdog/container cleanup regression
  Detail: cleanup did not retry inspection across delayed daemon registration
```

and passed on a re-run of the same commit, on the two commits before it, and locally (18/18) on the same tree. Nothing in that pull request touches shell scripts, the watchdog, or Docker.

## Why it is timing-dependent rather than wrong

The assertion is an ordering one, `scripts/tests/test-shell-portability.sh:1051-1053`:

```bash
"$first_inspect_line" -ge "$registration_line" || "$registration_line" -ge "$inspect_line"
```

It requires three events in strict order: cleanup inspects **before** the daemon finishes registering, registration completes, and cleanup inspects **again**. The fixture arranges that with a sleep loop and a release file, so the middle of the sandwich is a race: on a runner slow or loaded enough for the first inspect to land after the release, `first_inspect_line >= registration_line` and the test reports a cleanup regression that did not happen.

The property under test is real and worth keeping — cleanup must retry inspection across a delayed registration. It is the *evidence* that is racy: line order in a shared event log, with the interleaving arranged by wall-clock sleeps.

## Suggested fix

Make the fixture deterministic rather than fast: gate the registration release on the first inspect having been observed (block until `first_match_line -E '^docker inspect '` is non-empty, then release), instead of releasing on a timer and asserting the order came out right. That turns the race into a barrier and the assertion into a tautology of the thing actually being tested — that a second inspect happens after registration.

## Evidence

- Failing job: run `32539161904`, Contract Suites, commit `6e408122`
- Passing: the same commit after `rerun-failed-jobs`, plus commits `435a8e35` and `2cbb8bbb`
- Local: `npm run test:shell-portability` -> `Tests run: 18, Tests passed: 18`

wallstop:
**Why:** a grid cell was spending 6 of every 10 bytes on a hash the reader recomputed anyway, a CI test flaked on a race, and there was no way to turn a title into a URL-safe key.

**What:**

- `FastVector2Int`/`FastVector3Int` stop writing their cached hash — a 1,000-cell tilemap drops from 14,167 to 5,870 bytes. Old saves still read.
- The Unity watchdog cleanup test waits for a barrier instead of a sleep, so it no longer flakes under load.
- New `string.Slugify()`: `"Level 10: The Descent"` → `"level-10-the-descent"`.
- Agent guidance now asks for short pull request titles and bodies.

Fixes #519
Fixes #526
Fixes #386

---

> [!NOTE]
> **Medium Risk**
> Protobuf wire format for FastVector2Int/3Int is one-way: 3.5.1 cannot read new payloads. Old saves still deserialize because field 3 is skipped and z stays on tag 4.
> 
> **Overview**
> **FastVector protobuf no longer writes the cached hash.** Readers already recompute it from `x`/`y`/`z`, so a typical cell drops from ~10 bytes to the components only (a 1,000-cell tilemap: 14,167 → 5,870). Field 3 is skipped as unknown, so **legacy payloads still read**; `FastVector3Int.z` stays on tag 4 so a leftover hash is never interpreted as `z`. New payloads will not round-trip on 3.5.1.
> 
> Adds **`string.Slugify()`**: lowercase ASCII, single hyphens, accents folded (`"Café"` → `"cafe"`). Not a kebab-case substitute — punctuation and non-ASCII scripts are stripped (CJK-only input yields empty).
> 
> The Unity watchdog cleanup test waits on an inspect-complete barrier instead of a fixed sleep, so delayed Docker registration no longer races under load.
> 
> Agent/PR template now asks for a short **Why** / **What** body instead of long validation dumps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4dd50f72769c7742d5cf26b97a643bbb2f0c338b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Extension/StringExtensions.cs
```diff
@@ -1112,6 +1179,83 @@ public static string ToTitleCase(this string value, bool preserveSeparators = tr
...
+                    continue;
+                }
+
+                if (separatorPending)
+                {
+                    builder.Append(SlugSeparator);
+                    separatorPending = false;
+                }
+
+                builder.Append(isUpper ? (char)(current + 32) : current);
```
wallstop: Why not ToUpper/ToLower?

wallstop: Good catch — done in `0e5d1729`, and it turned out to matter for more than style.

`char.ToLowerInvariant` is now what runs, which is what the rest of this file already does in ten
other places. The `isUpper`/`isLower`/`isDigit` trio collapsed into one `isAsciiAlphanumeric` test
too, since the only reason to tell them apart was the arithmetic that is gone.

The thing worth recording is **why it has to be `ToLowerInvariant` and not `ToLower`**. Measured on
the live editor rather than assumed:

```text
culture=tr-TR   char.ToLower('I') = U+0131   char.ToLowerInvariant('I') = U+0069
```

Under a Turkish culture `ToLower` maps `I` to the dotless `ı`, which is **not ASCII** — so plain
`ToLower` would emit it and break the guarantee the method's own docs make. That is a live hazard in
Unity's Mono, not a theoretical one.

`SlugifyIsAsciiUnderATurkishCulture` pins it: it sets `tr-TR`, asserts `"HILL".Slugify()` is `"hill"`
and `"InstanceID".Slugify()` is `"instance-id"`, and restores the culture in a `finally`. tr-TR is the
one culture where lowercasing ASCII is not a no-op, so it is the only case that can catch this.

234 fixtures in `StringExtensionTests`, 0 failures.

Issue #529, opened 2026-08-22
[Feature]: RelationalComponent perf


wallstop:
### Feature Description

We should investigate and explore as many options as possible to increase the performance of our relational components - they are a critical piece of this package and should be as fast as possible.

### Use Case / Motivation

Extremely fast GetComponent/DI

### Proposed Solution

Heavy red-green, science, research.

### Unity Version

None

### Package Version

None

wallstop:
Code in #532. Posting the measurement history rather than just the result, because the first answer
was wrong and the way it was wrong is the useful part.

## The finding

A `[SiblingComponent]` array field cost **1.389 µs** where Unity's own `GetComponents(Type)` cost
**0.218 µs**. The gap is non-generic `Array` element access — `Array.GetValue`/`SetValue` measure
~68 ns per element against a typed indexer's ~1 ns.

**Casting the array to `IList` is not the fix**, which is worth recording because it is the obvious
one: 0.3461 µs for five elements against `Array.GetValue`'s 0.3370. Mono routes an array's `IList`
indexer through the same element-type re-check, so the fix had to leave the non-generic `Array` API
entirely rather than swap one accessor for another.

## The first numbers did not reproduce

Measured with one 100k run per shape, the change reported array **-23%**, `List<T>` **-8%**,
composite **-21%**. A repeat run reproduced none of it. Both sides were re-measured as **best of
three trials**, each gated on the loaded assembly actually being the variant under test — the
baseline by checking `Runtime` and `Tests` out at `main` and confirming the old `GetArray` symbol was
loaded before any number was reported.

| field shape           | baseline | pooled lease | final | vs baseline |
| --------------------- | -------: | -----------: | ----: | ----------: |
| array                 |    1.389 |        1.114 | 1.086 | **-22%**    |
| `List<T>`             |    0.996 |  1.069 (+7%) | 0.999 | parity      |
| `HashSet<T>`          |    1.022 |            — | 1.030 | parity      |
| single + array + list |    2.786 |        2.543 | 2.653 | -5%         |

## The regression the repeat run caught

The middle column is what was nearly shipped. Leasing a buffer from `Buffers<Component>.List` per
call **costs more than the array allocation it removes** on the `List<T>` path — a 7% regression
sitting inside a change advertised as a speed-up. The array path hid it, because there the lease is
paid for several times over by dropping `Array.CreateInstance` + `Array.Copy`.

[...]

PR #532, opened 2026-08-22
Speed up relational component assignment

### LINKED ISSUE #531: serialization: a refused protobuf surrogate registration is only reported to the log, where a game cannot act on it
Follow-up from PR #530, which fixed the failure but left the detection one-way.

## What #530 fixed

`RuntimeTypeModel.Default` is process-global and freezes a type the first time it serializes one, so
a surrogate registration can be **refused** rather than applied. All twenty of this package's
registrations shared one `try`, so a single refusal skipped every registration after it. Each is now
guarded independently, records the type in `ProtobufUnityModel.RegistrationFailures`, and logs an
error naming it.

## What is still one-way

`RegistrationFailures` is `internal`. A refused surrogate is **silent by nature** — the type keeps
serializing, with different bytes and no exception — so the only signal a shipped game gets is a
`Debug.LogError` at startup, on a path where the consequence is *saves written in a format the
package does not document and a later build will not read the same way*.

That is the one failure mode a game genuinely wants to assert on before it writes its first save,
and today it cannot:

```csharp
// What a consumer should be able to write, and cannot:
if (!Serializer.ProtobufSurrogatesReady(out IReadOnlyList<string> refused))
{
    // refuse to autosave, or fall back to JSON, rather than write bytes we cannot read back
}
```

## Why it is not obviously just "make the field public"

- The list is populated by a **static constructor**, so anything reading it has to have triggered
  that constructor first — which is exactly the ordering trap that caused the bug. The accessor has
  to call `EnsureInitialized()` itself rather than document that the caller should.
- The useful answer is arguably not "which types failed" but "is the wire format this package
  documents actually in effect", which is a different, smaller surface.
- It wants a decision about whether a refusal should be recoverable at all. It is not: protobuf-net
  cannot retro-fit a surrogate onto a frozen type, so the honest API reports rather than repairs.

## Suggested scope

- [ ] A public, `EnsureInitialized`-safe way to ask whether every surrogate registered, and which
      did not.
- [ ] Document it beside the AOT/surrogate section in the serialization guide, with the cause a
      consumer can actually act on: something reached `ProtoBuf.Serializer` before this package's
      `Serializer` was first touched.
- [ ] Decide whether `Serializer.ProtoSerialize` should refuse outright for an affected type rather
      than write bytes the package does not document.

Relates to #343, #504.

wallstop:
**Why:** relational components are the package's fastest path to a component reference, and array fields spent more than six times the cost of Unity's own query on themselves.

**What:**

- A `[SiblingComponent]` array field costs 22% less. `List` and `HashSet` fields are unchanged.
- Sibling collection fields no longer allocate per call.
- New `Serializer.ProtobufSurrogatesReady()` tells a game if protobuf will write the documented bytes.
- New public `ReflectionHelpers.CreateTypedArray()` builds a run-time-typed array about twice as fast.
- `audit-license-years.sh` drops 1,812 process launches; identical output.

Fixes #531
Refs #529 — one measured round, not the whole ask. Parent and child are converted but unmeasured.

---

> [!NOTE]
> **Medium Risk**
> Touches the hot parent/child/sibling assignment paths and protobuf surrogate reporting. Behavior should match, but IL2CPP fallbacks and scratch-list reuse are easy to get wrong.
> 
> **Overview**
> Makes `[SiblingComponent]` array assignment cheaper (~22%) and stops sibling collection queries from allocating a typed copy every call. Parent and child fast paths now filter and assign from reused `List<Component>` scratch instead of `Array.CreateInstance` / `SetValue`.
> 
> Adds public `ReflectionHelpers.CreateTypedArray()` (cached generic builder, loop not covariant `CopyTo`) and `Serializer.ProtobufSurrogatesReady()`, so a game can refuse a protobuf save when Unity types were frozen before this package’s surrogates. IL2CPP always reports ready.
> 
> Agent skills now require 50-char PR titles, measured replies to “could this be faster?”, and best-of-three Unity timings (allocation counters are unusable on this Mono). `audit-license-years.sh` reads the first line without forking `head`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b212ff3fe5469252e22ba99dd3e23eea470f3912. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Helper/ReflectionHelpers.cs
```diff
@@ -1141,6 +1141,158 @@ public static Func<int, Array> GetArrayCreator(Type elementType)
...
+        }
+
+        private static Array BuildTypedArray<TSource, TElement>(List<TSource> source, int count)
+            where TSource : class
+            where TElement : class
+        {
+            TElement[] result = new TElement[count];
+            for (int i = 0; i < count; ++i)
+            {
+                result[i] = source[i] as TElement;
```
wallstop, line 1304: Is it possible to optimize this with CopyTo or similar, for bulk operations?

wallstop: Measured it rather than reasoned about it, and the answer is two-part.

**Swapping this loop for a bulk call: no, it is 3.6-3.9x slower.** Building a `BoxCollider[]` from
the `List<Component>` this method receives, best of three trials on 6000.4.6f1 Mono (us):

| n   | this loop  | `List.CopyTo` | `Array.Copy` |
| --: | ---------: | ------------: | -----------: |
|   5 | **0.0878** |        0.3439 |       0.3692 |
|  50 | **0.6721** |        2.4317 |       2.2156 |
| 500 | **6.4056** |       23.3469 |      20.8070 |

The destination is a `TElement[]` and the source a `List<TSource>`, so a bulk call has to write
through a **covariant view** -- `BoxCollider[]` handed around as `Component[]` -- and the runtime
re-checks the element type on every store. The ratio is **flat from 5 to 500**, which is the tell:
this is per-element cost, not fixed overhead that amortizes. It also throws `InvalidCastException`
when the element type is an **interface**, which this method supports and has a test for, because an
interface array is not a covariant view of a class array.

**But the instinct is right, and the win is upstream.** `CopyTo` becomes a real memmove once the
source is already `List<TElement>`. Whole pipeline, query included:

| n   | today (non-generic query + loop) | generic query + `CopyTo` | generic + `ToArray` | available |
| --: | -------------------------------: | -----------------------: | ------------------: | --------: |
|   5 |                           0.2665 |                   0.2373 |          **0.2314** |     1.15x |
|  50 |                           1.6472 |               **1.2129** |              1.2189 |     1.36x |
| 500 |                          15.4724 |                  11.0717 |         **10.9960** |     1.41x |

So 15% at five elements rising to 41% at five hundred -- real, and it grows with the collection.

**I did not take it in this PR**, because it needs `GetComponents<T>` closed at run time, and this
file's own comments record that construct throwing in player builds until it was rewritten
non-generically. Trading the AOT-safe property for 15% at the common size, at the end of a session,
on a path a shipped game depends on, is not a trade I should make unilaterally. Filed as
[#534](https://github.com/Ambiguous-Interactive/unity-helpers/issues/534) with these numbers and the
fallback design it would need.

Two things changed from your question:

- The rationale is now a `<remarks>` block on the method, so the next reader gets the measurement
  instead of re-asking.
- [high-performance-csharp](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/.llm/skills/high-performance-csharp.md)
  gained a section on it. The existing guidance said bulk copy "is much faster than individual adds",
  which is true for exact-type copies and is exactly the intuition that inverts here.

wallstop: Swept the rest of the runtime for the same class, since a wrong bulk copy is worth finding
everywhere rather than only where it was noticed.

**No other instance.** Every remaining `Array.Copy` / `CopyTo` in `Runtime/` is an exact-type copy,
where the bulk call is genuinely the fast one:

| Site                                                       | Copy                              |
| ---------------------------------------------------------- | --------------------------------- |
| `Serializer`, `WProtoWriter`, `StringExtensions`           | `Span`/`ReadOnlySpan` — no covariance possible |
| `Serializer` buffer growth                                 | `byte[]` to `byte[]`              |
| `SystemRandom`, `DotNetRandom`, `PerlinNoise` state copies | `int[]` to `int[]`                |
| `Helpers.GetAssetPaths`                                    | `ICollection<string>` to `string[]` |
| `IListExtensions`, the five sort implementations           | `IList<T>` to `T[]`               |

The covariant-destination shape existed only in the three relational attribute files, and this PR
removes all of it — the old `Array.CreateInstance` + `Array.Copy` + `Array.GetValue`/`SetValue`
chain that `CreateTypedArray` replaced was exactly that pattern, which is why the rewrite is a win
in the first place.

PR #535, opened 2026-08-22
Stop relational child/parent fields allocating

### LINKED ISSUE #486: lint: the PowerShell linters spend ~70% of a 14-minute serial repo-lint on file enumeration and reads, not scanning

### LINKED ISSUE #533: ci: the SINGLE_THREADED branch compiles nowhere in the local typecheck, and CI runs two legs of it
`npm run typecheck:unity` builds `Runtime/**` three ways — the `WALLSTOP_PROTO` default, the
define-off legacy fallback, and `WALLSTOP_UNITY_HELPERS_ODIN_INSPECTOR`. It does **not** build the
`SINGLE_THREADED` branch, and CI runs two `SINGLE_THREADED` legs.

## Why this is the same shape as #347

[context.md](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/.llm/context.md)
records why the Odin configuration exists:

> The Odin configuration exists because Odin changes the **base class** of `RuntimeSingleton<T>`,
> `ScriptableObjectSingleton<T>` and `AttributeEffect`, and that branch compiled nowhere in
> automation until #347 -- which is how #275 shipped a compile break to consumers.

`SINGLE_THREADED` is the same kind of switch and is unguarded in the same way. It changes real
declarations, not just call sites — `ReflectionHelpers` alone swaps five caches between
`ConcurrentDictionary` and `Dictionary` under it, each with a differently-shaped cached getter:

```csharp
#if SINGLE_THREADED
        private static readonly Dictionary<Type, Func<int, Array>> ArrayCreators = new();
#else
        private static readonly ConcurrentDictionary<Type, Func<int, Array>> ArrayCreators = new();
#endif
```

A change that adds a cache to that file compiles clean locally, passes every local gate, and can
still red both `SINGLE_THREADED` legs — for the cost of a full matrix run to find out.

## How small the fix is

Found while adding one such cache in #532. Verifying it by hand needed four lines in
`Generator~/WallstopStudios.UnityHelpers.TypeCheck/WallstopStudios.UnityHelpers.TypeCheck.csproj`,
mirroring the switch already there for Odin:

```xml
    <DefineConstants Condition="'$(WallstopSingleThreaded)' == 'true'"
      >$(DefineConstants);SINGLE_THREADED</DefineConstants>
```

With that, `dotnet build … -p:WallstopSingleThreaded=true` **built clean in 25 s**. The change was
reverted rather than landed, to keep #532 to its subject.

## Suggested work

- [ ] Add the `WallstopSingleThreaded` property to the TypeCheck csproj, and the matching one to
      `WallstopStudios.UnityHelpers.TestCheck`.
- [ ] Add `typecheck:unity:singlethreaded` and `typecheck:tests:singlethreaded`, and chain them into
      `typecheck:unity` / `typecheck:tests` beside the Odin ones.
- [ ] Update the local-gates list in `.llm/context.md`, which currently says the tree is built
      "three ways".

Cost is ~25 s added to a gate that already runs five builds, against one full Unity matrix per
missed break.

### LINKED ISSUE #534: perf: relational queries could fill a typed buffer, worth 1.15x-1.41x, at the cost of a run-time-closed generic
Found while answering review feedback on #532, which asked whether the typed-array build could use a
bulk copy. It cannot where it stands, but the bulk win exists one step upstream.

## The measurement

`ReflectionHelpers.CreateTypedArray` receives a `List<Component>` and must produce a
`TElement[]`. A bulk call there writes through a covariant view and re-checks the element type per
store, so it is **3.6-3.9x slower than the element loop** and illegal for interface element types.
Details are in the method's `<remarks>` and in
[high-performance-csharp](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/.llm/skills/high-performance-csharp.md).

The gain is in never producing the untyped list. Whole pipeline, query included, best of three
trials on 6000.4.6f1 Mono (us):

| n   | today: `GetComponents(Type, List<Component>)` + loop | `GetComponents<T>(List<T>)` + `CopyTo` | + `ToArray` | available |
| --: | ---------------------------------------------------: | -------------------------------------: | ----------: | --------: |
|   5 |                                               0.2665 |                                 0.2373 |  **0.2314** |     1.15x |
|  50 |                                               1.6472 |                             **1.2129** |      1.2189 |     1.36x |
| 500 |                                              15.4724 |                                11.0717 | **10.9960** |     1.41x |

15% at the common size, 41% at five hundred, and it **grows with the collection** rather than
shrinking -- the opposite shape from most micro-optimizations.

## Why it was not taken in #532

`GetComponents<T>(List<T>)` has to be closed at run time for an element type known only from a field.
That is the construct the relational fast path already tried and reverted. Its comment is still in
the tree:

> AOT-safe: the non-generic Type overload avoids the runtime generic-method + Expression.Compile
> path, which IL2CPP cannot service (the old compiled path threw at runtime in player builds).

So this is not a free 15%. It is 15% in exchange for re-entering a construct that has broken player
builds in this package once already.

## What it would need

- [ ] A per-element-type cached delegate that closes `GetComponents<T>(List<T>)`, built the way
      `CreateTypedArray`'s builder is, including the same **fail-soft** guard: a refusal caches null
      and the non-generic path serves that type forever after.
- [ ] The failure must also be caught on **first invocation**, not only at delegate construction.
      `MakeGenericMethod` succeeding does not prove IL2CPP can execute the result, and #532's guard
      only covers construction.
- [ ] A typed `List<TElement>` buffer per element type, with the same retention bound the scratch
      lists use (`MaximumRetainedScratchCapacity`).
- [ ] Proof on a real IL2CPP standalone player, not just the editor. The gated standalone legs
      exercise this repository's element types; a consumer's own types are different instantiations,
      so a green leg is necessary and not sufficient.
- [ ] Re-measure with the rig #532 used: best of three trials, each gated on the loaded assembly
      being the variant under test. A single run of this workload produced numbers that did not
      reproduce.

## Scope note

The same shape applies to `GetComponentsInParent`/`GetComponentsInChildren`, which have **no**
non-generic caller-buffer overload at all today, so they still allocate a `Component[]` per call.
Those would gain the allocation removal as well as the copy, and were never measured -- see the open
items on [#529](https://github.com/Ambiguous-Interactive/unity-helpers/issues/529).

wallstop:
**Why:** every `[ChildComponent]` / `[ParentComponent]` collection assignment allocated a `Component[]` that was copied out and thrown away, because the non-generic Unity queries have no caller-buffer overload.

**What:**

- Child and parent collection queries now fill a reused buffer of the field's element type; the non-generic path stays as a cached, fail-soft fallback.
- The sibling path was measured and deliberately left alone -- 0.998x, and nothing to stop allocating.
- New differential tests assert the fast path and the fallback agree, so the fallback is no longer only reachable on platforms no test runs on.
- `SINGLE_THREADED` now compiles in `npm run typecheck:unity`, which built the tree three ways while CI ran two legs of it.
- Three PowerShell linters go 25.2 s -> 20.1 s with byte-identical output.

Fixes #534
Fixes #533
Fixes #486

Review feedback addressed in 11cf9e11: the collector cache now fills through `GetOrAdd` (state-taking overload, no closure), and the hand-rolled `[ThreadStatic]` scratch is gone in favour of `Buffers<T>` — measured at 0.976-0.983x across three orderings, so the specialization bought nothing. Both measured rules are now in `high-performance-csharp`. The remaining 17 non-atomic caches are inventoried on #536.

---

> [!NOTE]
> **Medium Risk**
> Touches relational component assignment (Awake-time scene wiring) and adds a runtime-generic AOT fallback cache. Behavior is intended to match the old path, but IL2CPP refusal handling and cache races are the main review surface.
> 
> **Overview**
> `[ChildComponent]` and `[ParentComponent]` collection assignment no longer allocates a throwaway `Component[]` on every call. Queries now close Unity’s generic, buffer-taking overloads over the field’s element type via a new `RelationalComponentCollector`, with a cached fail-soft fallback to the non-generic path when AOT refuses the instantiation. Sibling queries are unchanged (measured: no allocation to remove). Tests assert the fast path and fallback agree on the same hierarchies.
> 
> Local typecheck now builds `SINGLE_THREADED` as a fourth define branch (runtime + tests), matching CI legs that swap cache declarations. Concurrent collector population uses `GetOrAdd`; `Buffers<T>` is the scratch default.
> 
> Separately, three PowerShell linters walk files with `EnumerateFiles`/`ReadAllLines` instead of `Get-ChildItem`/`Get-Content` on the 9p mount.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a032cb0309a92e078af6e4578709ac9d33a0034. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Attributes/RelationalComponentCollector.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// </param>
+        /// <returns>The collector, or null when this runtime refuses it.</returns>
+        internal static RelationalComponentCollector For(Type elementType, Component probe)
+        {
+            if (elementType == null || FallbackOnly)
+            {
+                return null;
+            }
+
+            if (Collectors.TryGetValue(elementType, out RelationalComponentCollector cached))
```
wallstop, line 92: This is non-atomic - should we utilize something like TryAdd, with a non-allocating delegate?

Runtime/Core/Attributes/RelationalComponentCollector.cs
```diff
@@ -0,0 +1,247 @@
...
+                }
+            }
+
+            internal override int CollectParentsInto(
+                Component source,
+                bool includeInactive,
+                List<Component> destination
+            )
+            {
+                List<TElement> buffer = Rent();
```
wallstop: Should we utilize Buffers at all here? Why specialize?

Runtime/Core/Attributes/RelationalComponentCollector.cs
wallstop: You're right — fixed in 11cf9e11.

`GetOrAdd` with the **state-taking** overload, so the lambda stays `static` and no closure allocates for `probe`:

```csharp
return Collectors.GetOrAdd(elementType, static (type, live) => Create(type, live), probe);
```

The `probe == null` guard stays in *front* of it rather than moving inside the factory. That is deliberate: a factory handed a null probe would cache a refusal permanently for a reason that has nothing to do with whether this runtime supports the instantiation. `SINGLE_THREADED` keeps the plain form, where there is no race to lose.

Worth noting what the race actually cost, since it is more than duplicated work: the factory here **probes** — it invokes every entry point once against a live component — so two racers each built *and probed*, and each returned a different instance than the one that ended up cached. Benign given the collectors are stateless, but there was no reason to leave it.

**Swept the rest of the codebase for the same shape.** 71 raw hits, but 53 are inside `#if SINGLE_THREADED` branches where the field is a plain `Dictionary` — false positives. **18 are genuinely concurrent**, and mine was one. The other 17 are all memoize-after-miss caches whose computed value is deterministic (type -> root type, type -> factory delegate), so a lost race stores an equal value and is harmless — a real smell, not a real bug. I left them out of this PR rather than sprawl it, and filed them with the inventory. Also added the rule to `high-performance-csharp` so new caches get the atomic form by default.

Runtime/Core/Attributes/RelationalComponentCollector.cs
wallstop: Measured it rather than argue from the neighbours — and **the answer is you're right, so `Buffers<T>` it is.** 11cf9e11 deletes the hand-rolled scratch (~35 lines, plus `Rent`/`Return`/`[ThreadStatic]`).

A/B in one domain, best of three 200k-iteration trials, on `AssignChildComponents`:

| ordering | `[ThreadStatic]` | `Buffers<T>.List` lease | ratio |
| --- | ---: | ---: | ---: |
| scratch first | 4.7469 µs | 4.6658 | 0.983x |
| pooled first | 4.6483 µs | 4.5375 | 0.976x |
| interleaved | 4.7969 µs | 4.7000 | 0.980x |

Three orderings, all agreeing: the lease is ~2% **faster**, not slower. So the specialization bought nothing here.

**Why I reached for it in the first place, and why that reasoning was wrong:** session 215 measured a pool lease costing **7%** on the sibling path and left that comment in all three fast invokers, so I copied the pattern. But that call is ~1 µs and this one is ~4.7 µs — the lease's cost is roughly fixed, so it is a large fraction of a cheap call and negligible in an expensive one. **Neither number generalizes to the other call site**, which is the actual lesson, and it is now in `high-performance-csharp` as a table of both results plus the explanation. I did **not** touch the sibling invoker's `[ThreadStatic]` — that one has its own measurement and it still stands.

The checklist entry there said "All temporary collections use `Buffers<T>`" unqualified, which is what made this worth writing down: the rule is right, and hand-rolling is an optimization that needs its own number for its own call site.

PR #537, opened 2026-08-22
Fix a second await of one AsyncOperation hanging

### LINKED ISSUE #536: perf/consistency: 17 ConcurrentDictionary caches fill through the indexer after a miss, which is not atomic
Found while answering review feedback on #535, which asked whether a `TryGetValue`-miss-then-indexer-store into a `ConcurrentDictionary` should be atomic. It should, that one is fixed, and a sweep found the same shape in 17 more places.

## The shape

```csharp
// Not atomic: two callers racing a first use each build, and each returns a
// different instance than the one that ends up cached.
if (!Cache.TryGetValue(key, out Value cached))
{
    cached = Build(key);
    Cache[key] = cached;
}
```

`GetOrAdd` — with the state-taking overload where the factory needs an argument, so the lambda stays `static` and no closure allocates — stores exactly one winner and hands it to every caller.

## Scope: 18 sites, of which 1 is fixed

The sweep is easy to get wrong. A naive grep finds **71** indexer writes to fields declared `ConcurrentDictionary`, but **53 of those are inside `#if SINGLE_THREADED` branches** where the field is a plain `Dictionary` and the `#else` branch already uses `GetOrAdd` correctly. Any re-run of this sweep must track preprocessor state or it will report a problem four times bigger than it is.

The 17 remaining:

| file | lines |
| --- | --- |
| `Runtime/Core/Extension/AsyncOperationExtensions.cs` | 59, 67 |
| `Runtime/Core/Serialization/Serializer.cs` | 1338, 1339, 2082, 2102, 2116, 3223 |
| `Editor/CustomDrawers/SerializableDictionaryPropertyDrawer.cs` | 8099, 8114, 8163, 8186 |
| `Editor/CustomDrawers/SerializableSetPropertyDrawer.cs` | 4181, 5064, 5069 |
| `Editor/Utils/WButton/WButtonValueUtility.cs` | 96, 102 |

## Severity: a smell, not a live bug — which is why this is filed rather than fixed

**Every one of these caches a deterministic value** (type -> protobuf root type, type -> parameterless-constructor delegate, type -> "unsupported" marker). A lost race therefore stores an *equal* value, so no caller can observe a wrong answer. That is the opposite of the case on #535, whose factory had a side effect: it *probed* the runtime by invoking the built collector, so a race ran the probe twice.

So the value here is consistency and removing a foot-gun, not fixing a defect. Two things would raise the priority:

- If any of these factories gains a side effect (a log, a registration, a diagnostic) it becomes a real bug silently.
- The `Serializer.cs` ones are the only runtime, non-editor entries besides `AsyncOperationExtensions`, and serialization is the most likely place for a genuinely multi-threaded caller.

## Suggested work

- [ ] Convert the 6 `Serializer.cs` sites and the 2 `AsyncOperationExtensions.cs` sites, which are runtime rather than editor.
- [ ] Convert the 9 editor-drawer sites, which are main-thread-only in practice and lowest value.
- [ ] Check each conversion keeps its `#if SINGLE_THREADED` counterpart on the plain-`Dictionary` form.
- [ ] Consider a lint check: an indexer write to a `ConcurrentDictionary` field outside a `SINGLE_THREADED` branch is almost always this pattern.

The rule is already recorded in [high-performance-csharp](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/.llm/skills/high-performance-csharp.md) under "Populating a cache must be atomic", so new code should not add to the list.

wallstop:
**Why:** a second `await` of one `AsyncOperation` left the first awaiter hung forever.

**What:**

- Combine awaiter continuations instead of overwriting them.
- Fill every `ConcurrentDictionary` cache atomically; add a linter for it.
- Read `SerializableDictionary`/`HashSet` JSON without reflection.
- Assign a hierarchy 30% faster when most components have no relational fields.

Fixes #536

---

> [!NOTE]
> **Medium Risk**
> Touches concurrent caches, protobuf root registration, and awaiter continuations. Wrong GetOrAdd vs overwrite (already called out in the skill) can silently keep a bad cached factory.
> 
> **Overview**
> A second `await` of the same `AsyncOperation` no longer overwrites the first continuation, so every waiter of one `LoadSceneAsync` (or similar) handle resumes.
> 
> Caches that used `TryGetValue` then the indexer now fill through `GetOrAdd`/`TryAdd`, and cache-factory lambdas are `static` (or a cached method-group field) so a hit does not allocate a delegate. `HasRelationalAssignments` drops its lock for a concurrent dictionary, which is most of `AssignHierarchy` when a component has no relational fields. JSON converters write `_keys`/`_values`/`_items` directly instead of reflecting them on every deserialize.
> 
> A new `lint-concurrent-cache-fill` check (with self-tests) fails non-atomic ConcurrentDictionary fills and non-static factories; deliberate overwrites need a `// concurrent-overwrite:` reason.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3336e8b230d3ff9ae3456429a32d58c8bd55667. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/Utils/WButton/WButtonValueUtility.cs
```diff
@@ -92,14 +92,16 @@ internal static bool TryCreateInstance(Type type, out object value)
 
             try
             {
-                Func<object> factory = ReflectionHelpers.GetParameterlessConstructor(type);
-                ParameterlessFactoryCache[type] = factory;
+                Func<object> factory = ParameterlessFactoryCache.GetOrAdd(
+                    type,
+                    ReflectionHelpers.GetParameterlessConstructor(type)
```
wallstop, line 97: Is this allocation-free? I think this captures, shouldn't it be (inputType => ...)

Please scan for functions/lambdas that will allocate in GetOrAdd and similar calls, and ensure that they are allocation-free.

Editor/CustomDrawers/SerializableDictionaryPropertyDrawer.cs
```diff
@@ -8160,17 +8160,20 @@ private static bool TryInstantiateWithoutConstructor(Type type, out object insta
...
+                            {
+                                return Activator.CreateInstance(type, nonPublic: true);
+                            }
+                            catch
+                            {
+                                return null;
+                            }
                         }
-                    };
+                    );
```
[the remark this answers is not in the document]
wallstop: Already fixed, in `e12933b1` — you were reviewing `fbc3fad7`. Same diagnosis, and it was mine to make: converting those two stores to `TryAdd` in the atomicity sweep turned a required overwrite into a no-op.

Proved both ways on a live 6000.4.6f1 editor, three consecutive calls with a type whose non-public parameterless constructor throws:

| | call 1 | call 2 | call 3 |
| --- | --- | --- | --- |
| `TryAdd` | True | **False** | **False** |
| indexer overwrite (restored) | True | True | True |

Both sites are indexer writes again, each carrying a `// concurrent-overwrite:` comment saying why the write has to win.

Editor/Utils/WButton/WButtonValueUtility.cs
wallstop: **That line specifically: no.** It is `GetOrAdd(TKey, TValue)` — the *value* overload — so there is no lambda and nothing to capture. `ReflectionHelpers.GetParameterlessConstructor(type)` is an argument, evaluated eagerly, and the early `TryGetValue` above means we only reach it on a miss. Same as before the change.

**Your broader suspicion was right, though, and the scan found eight real ones.** Four of them in `Serializer`.

### I made the compiler answer instead of reading the lambdas

Guessing which lambdas capture is how I introduced the bug in the thread above, so: `static` was added to **every** lambda handed to `GetOrAdd`/`AddOrUpdate`/`GetValue` in the tree, and whatever failed with `CS8820` was a real capture. `Runtime` through `typecheck:unity`; `Editor` cannot be typechecked locally (#347) so it went through the MCP editor's own Roslyn.

**5 capturing factories in `Runtime`, 2 in `Editor`, out of 96 lambdas.** The other 89 were already non-capturing, so the compiler was already caching them in a static field — those are free, and `static` on them is a guarantee rather than a saving.

### What it cost — 400,000 warm-cache hits per shape, 6000.4.6f1, control moved 30.6 MB

| factory shape | bytes/call |
| --- | ---: |
| lambda capturing a method parameter | **115.8** |
| method group, e.g. `GetOrAdd(key, Build)` | **106.3** |
| `static` lambda + state-taking overload | 0.0 |
| cached `static readonly Func<...>` field | 0.1 |

The method-group row is the one I would not have predicted: C# 9 does not cache a method-group conversion (that landed in C# 11), so `TypeAccessors.GetOrAdd(collectionType, CreateAccessors)` was allocating a delegate on **every** collection serialize, cache hit included.

**One correction to my own first measurement.** It put the capturing lambda at *0 B/call* — because I declared the captured local beside the loop, and Roslyn creates the display class once, where the variable is declared. Real code captures a **parameter**, so a fresh display class is built per call. Reproducing the method rather than the shape moved it from 0 to 115.8. I nearly reported the zero.

### Fixed

| site | was | now |
| --- | --- | --- |
| `Serializer.GetWrapperAccessors` | captured `isSet` | state-taking overload; `bool` as `TArg` is generic, so not boxed |
| `Serializer.GetAccessors` | method group | cached `static readonly` field |
| `Serializer.SerializeSpecialCollection` / `Deserialize…` | method group | cached fields |
| `ReflectionHelpers.Factory` ×4 | captured `property` | read `property` off the key tuple |
| `WButtonParameterDrawer` | captured `label`, **every IMGUI repaint** | plain `TryGetValue` + store |
| `MethodAnalyzer` | captured `classInfo` | state-taking `AddOrUpdate` |
| `Buffers` ×4 | non-`static` outer pool lambdas | `static`; the inner capture of the factory's own parameter is required and stays |

### Made permanent

`lint-concurrent-cache-fill` gained a second rule: a cache-factory lambda must be `static`. It is not about the `static` keyword saving anything — it is that the compiler then **rejects** a capture, so the expensive shape stops compiling.

It keys off top-level `=>` arrows rather than splitting on commas, because an earlier draft split `WallstopGenericPool<Dictionary<TKey, TValue>>` at the generic comma and reported four false positives in `Buffers.cs`. Depth also excludes a nested `onRelease: set => set.Clear()`, which is an argument to the constructed value rather than to the cache. Self-test is 12 cases, four of them pinning exactly these shapes.

**The method-group half is not enforceable** and I want to be explicit about that: `GetOrAdd(key, Build)` is lexically identical to `GetOrAdd(key, localDelegate)`. Only review catches a new one. That limitation is written into the linter's own header.

PR #542, opened 2026-08-23
Warn when a lookup factory allocates per call

### LINKED ISSUE #400: mcp: the Unity test runner cannot see this package's tests (host project lacks `testables`)
## The gap

The Unity MCP bridge can compile this package (content assertions on newly-added types resolve, and
`CompilationPipeline.GetAssemblies` reports all 39 `WallstopStudios.UnityHelpers*` assemblies). But
its **test runner cannot see this package's tests**.

Measured in session 175 via the `UnityHelpersMcpTestRunner` bridge in the host project's
`Assets/Editor`:

```
ListTests  -> 189 test entries, 0 matching "WProtoRepeated"
Run(editmode, "WProtoRepeatedContractTests", ...) -> pass=0 fail=0 skip=0 seconds=0.3
```

A filter matching nothing returns `pass=0 fail=0`, which reads exactly like success. That is the
trap: a run that verified nothing is indistinguishable from a clean run unless you check the counts.

## Why it happens

The host Unity project (`D:/Code/Packages`) does not list `com.wallstop-studios.unity-helpers` in the
`testables` array of its `Packages/manifest.json`. Unity only enumerates tests from packages named
there. Compilation does not require it, which is why the compile gate works and the test gate does
not.

## Why it matters

This exact gap cost a CI round trip in session 175. The packed-encoding change updated the
`Generator~` differential suite but missed the Unity suite's committed byte vectors; six tests failed
on all four playmode legs. Nothing available locally could have caught it:

- `npm run typecheck:unity` excludes `Tests/` - and a type-checker would not flag a wrong string
  literal anyway.
- The `Generator~` harness compiles the runtime sources but not the Unity test assembly.
- The MCP test runner cannot see the tests, per above.

## Fix

Add the package to `testables` in the host project's manifest:

```json
{
  "testables": ["com.wallstop-studios.unity-helpers"]
}
```

That file lives outside this repository, so this needs the maintainer rather than a PR here.

## Guard worth adding regardless

Whatever the runner situation, a helper that runs package tests should **fail when the filter matches
zero tests**. `pass=0 fail=0` currently reads as success, and that is a false green of the same shape
as the "Files checked: 0" spelling bug this repo already fixed once.

### LINKED ISSUE #457: ci: measure whether the fast->standalone failure gate still earns its 4.3 runner-minutes
## The question

`unity-tests-standalone` has `needs: unity-tests`, so a fundamentally broken tree never starts the
IL2CPP legs. Session 194 measured what that gate costs and what it saves, and the two numbers are
close enough that the decision should be made on the failure *mix* rather than on the headline rate.

## What it costs

Measured across the six most recent green Unity Tests runs, by attributing each same-runner idle gap
to the tier boundary blocking it:

| Edge | median | total over 6 runs |
| --- | ---: | ---: |
| `unity-tests` → `unity-tests-standalone` | **4.3 runner-min** | 29.2 |
| `unity-tests-standalone` → `unity-tests-single-threaded` | 2.8 runner-min | 24.9 |

Median wall clock 37.2 min against a 25.0 min packing floor (sum of leg durations over two runners).

The second edge was removed in #456 — its justification was a premise `PLAN.md` already records as
disproven. This issue is only about the first.

## What it saves

Four standalone legs at **2.0–4.3 min** each in those runs, so skipping the tier saves roughly **10
runner-minutes** when it fires.

The workflow comment says the gate protects against "~30-min-each standalone builds". That is **10x
stale** — the number it was written against no longer exists. `#456` corrects the comment; it does not
change the gate.

## The part that is not measured

**25 of the last 60 completed Unity Tests runs (42%) ended red.** At that rate the gate fires often
enough to be worth 4.3 runner-minutes on every green run. But a red *run* is not the same as a red
*fast tier*, and this repository has several failure modes that are not:

- `Stale pull request run for <sha>` supersession (#332)
- build-lock input contract failures (#356, #411)
- `runner-preflight` (#415)
- an OS kill stranding the seat (#353)

None of those is a broken tree, and for none of them does the gate save anything — the standalone legs
would have been skipped or failed regardless.

## The work

1. For the last 60 Unity Tests runs, pull `/actions/runs/{id}/jobs` and classify each failure by which
   job failed first, separating a genuine `unity-tests` (fast tier) failure from the four modes above.
2. If real fast-tier failures are a small fraction, the gate's expected saving falls below its 4.3
   runner-minute cost and the edge should be removed — putting all 14 legs in flight at once and
   taking wall clock toward the packing floor.
3. If they are common, keep it, and record the measured rate beside the `needs:` so the next session
   does not re-derive it.

Either way the outcome is a number in the workflow comment rather than an assumption.

## Not in scope

The `max-parallel` question (already removed, sessions 166/169) and sibling contention on the shared
runners (0.4–41.5 runner-min in the same six runs), which is an organization capacity decision rather
than a workflow one.

*Found while answering "why do the tiers wait?" in session 194.*

### LINKED ISSUE #538: analyzer: catch method groups passed to cache factories, which the source linter cannot
## What is not covered

Session 217 found that a **method group** handed to a cache factory allocates a delegate on every call, cache hit included:

```csharp
// C# 9 does not cache a method-group conversion (C# 11 does). This builds a new
// Func<Type, Accessors> on every call, even when the key is already present.
return TypeAccessors.GetOrAdd(collectionType, CreateAccessors);
```

Measured on 6000.4.6f1, 400,000 warm-cache hits, against a control that moved 30.6 MB:

| factory shape | bytes/call |
| --- | ---: |
| method group | **106.3** |
| lambda capturing a method parameter | **115.8** |
| `static` lambda + state-taking overload | 0.0 |
| cached `static readonly Func<...>` field | 0.1 |

Three such sites were found by hand in `Serializer.cs` (`GetAccessors`, `SerializeSpecialCollection`, `DeserializeSpecialCollection`) and replaced with cached fields.

## Why the linter cannot catch it

`scripts/lint-concurrent-cache-fill.ps1` now enforces that every **lambda** handed to `GetOrAdd`/`AddOrUpdate`/`GetValue` is `static`, which makes the compiler reject a capture (`CS8820`). That half is solved and permanent.

The method-group half is **not lexically decidable**:

```csharp
GetOrAdd(key, CreateAccessors)   // method group -- allocates every call
GetOrAdd(key, cachedFactory)     // local holding a delegate -- free
```

Both are a bare identifier in argument position. Telling them apart needs symbol resolution, not a regex. A casing heuristic (PascalCase = method) would be wrong the first time someone names a field `Factory`. The limitation is written into the linter's header so the next person does not assume coverage exists.

## What would fix it

This repository already ships a Roslyn analyzer (the `WPROTO###` family, `Generator~/`). An analyzer has the semantic model, so this is a few lines:

- On an `IMethodReferenceOperation` in an argument position of `GetOrAdd` / `AddOrUpdate` / `GetOrCreateValue` / `GetValue`, report a diagnostic naming the fix ("hold it in a `static readonly Func<...>` field").
- Same check catches the equivalent shape in `ConditionalWeakTable.GetValue` and any future cache API added to the list.

Scope is small and it closes the class permanently, which the source-based linter cannot.

## Priority

Low-to-medium. Every site known today is fixed, and the shape is rare — it took a deliberate sweep to find three. But it is invisible when it reappears: the code reads correctly, nothing fails, and a delegate is allocated per call on a serialization path.

Recorded in [high-performance-csharp](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/.llm/skills/high-performance-csharp.md) under "A cache factory must be `static`, and a method group is not free".

### LINKED ISSUE #539: docs: four skills sit in the CRITICAL size band, so any edit to them blocks its own commit
## What happens

`scripts/lint-skill-sizes.ps1` warns CRITICAL at 480 lines and errors at 500. Four skills are in the CRITICAL band today:

```text
[skill-sizes] CRITICAL: avoid-magic-strings.md: 487 lines - only 13 lines from 500 limit! Consider splitting now
[skill-sizes] CRITICAL: editor-caching-patterns.md: 484 lines - only 16 lines from 500 limit! Consider splitting now
[skill-sizes] CRITICAL: unity-performance-patterns.md: 495 lines - only 5 lines from 500 limit! Consider splitting now
[skill-sizes] CRITICAL: use-discriminated-union.md: 494 lines - only 6 lines from 500 limit! Consider splitting now
[skill-sizes]   OK: 52, Warnings: 51, Critical: 4, Errors: 0
```

Repository-wide that is only a warning, so `lint:repo` stays green. But `npm run agent:preflight` checks **changed** skill files and **fails** on CRITICAL. So the band is invisible until you edit one of these four, at which point your unrelated change cannot be committed until you either split the file or cut the paragraphs you just added.

## Why it is worth doing before someone hits it

Session 217 hit it. Adding three measured findings to `high-performance-csharp.md` pushed it from ~420 to 480 and blocked the commit. The session condensed its own prose to 469 to get under the line — which worked, but it traded away detail to pay for a limit that has nothing to do with the change, and it leaves the file 11 lines from blocking the next person for the same reason.

Four files are now in that position. Each is one modest addition away from the same forced choice.

## Options

1. **Split each into a focused sibling**, which is what the linter's own message advises ("Consider splitting now"). `high-performance-csharp.md`, for instance, has grown a coherent cluster on cache population cost — atomicity, factory allocation, lock-vs-lookup — that would stand on its own and links back cleanly. Adding a skill means a trigger comment plus `pwsh -NoProfile -File scripts/generate-skills-index.ps1`; both are validated by `lint:llm`.
2. **Raise the limit** if 500 no longer reflects what an agent can usefully read in one file. That is a judgement about context budget, not about these four files.
3. **Make CRITICAL a warning in agent-preflight too**, so the band informs rather than blocks. Cheapest, and the weakest: it removes the forcing function that keeps these files from reaching the hard 500 error.

(1) is the intended path and the only one that actually reduces the files. Filing rather than doing it because splitting four skills is a documentation change with its own review surface, and it should not ride along inside an unrelated pull request — which is precisely the trap this issue is about.

wallstop:
**Why:** A lookup that rebuilds its factory delegate on every call allocates on exactly the hits it exists to make cheap, and nothing in the toolchain could see it.

**What:**

- Add `WUH001` in a new `WallstopStudios.UnityHelpers.Analyzers` assembly: a method group passed to `ConcurrentDictionary`, `ConditionalWeakTable`, or this package's own `IDictionary` extensions. On by default, a warning at most, switchable off.
- Fix the one site it found, in `Editor/Utils/WButton/`, which no local gate compiles.
- Run `validate:tests:fast` as a concurrent registry: 553 s serial to ~150 s.
- Split four skills out of the CRITICAL 480-500 line band.
- Record the standalone gate's measured failure mix beside its `needs:`.

Fixes #538
Fixes #539
Fixes #457
Fixes #400

---

> [!NOTE]
> **Medium Risk**
> Adds a Roslyn analyzer that runs on consumer compilations (warning-only, suppressible) and rewires the fast test aggregate to a concurrent registry, so a runner bug can skip or flake many checks.
> 
> **Overview**
> Ships a second analyzer assembly (`WUH###`, warning-only, never an error) so consumers get a default-on diagnostic for a shape a source linter cannot see: a **method group** passed as a lookup factory, which allocates a new delegate on every call under Unity’s C# 9.
> 
> **`WUH001`** covers `ConcurrentDictionary`/`ConditionalWeakTable` factory members and every delegate-taking member of this package’s `DictionaryExtensions`. It stays silent on C# 11+ (cached conversions), cached fields/locals, and lookalike types. CI byte-compares the committed DLL, typecheck/testcheck load it, and `wproto-generator` now builds/tests it without renaming the required check.
> 
> The one in-tree hit is `WButtonStateRepository.GetOrCreate`, which now holds a `static readonly` `CreateValueCallback`.
> 
> `validate:tests:fast` is no longer a serial `&&` chain: shared `check-runner.js` runs the registry concurrently (exclusive checks first), reports every failure in one pass, and contract tests assert coverage. LLM skills over the 480-line band are split out; the standalone Unity job comment records measured gate savings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1488a89ab403bc8334249f277532908b0867ccb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Generator~/WallstopStudios.UnityHelpers.Proto.Generator/CacheFactoryAnalyzer.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// </remarks>
+        private const int FirstLanguageVersionThatCachesMethodGroups = 1100;
+
+        /// <summary>
+        /// Cache types whose fill factories run on every lookup rather than once. Matched on
+        /// <see cref="ISymbol.MetadataName"/>, which carries the arity, so a consumer type that
+        /// merely shares the name is not caught.
+        /// </summary>
+        private static readonly ImmutableHashSet<string> CacheTypes = ImmutableHashSet.Create(
+            "System.Collections.Concurrent.ConcurrentDictionary`2",
```
wallstop: What about Dictionary?

Generator~/WallstopStudios.UnityHelpers.Analyzers.Tests/CacheFactoryAnalyzerTests.cs
```diff
@@ -0,0 +1,298 @@
...
+    /// concurrent cache's factory allocates a delegate on every call, cache hit included.
+    /// </summary>
+    /// <remarks>
+    /// The negative cases carry the weight here. The whole reason this is an analyzer rather than a
+    /// regex is that <c>GetOrAdd(key, Factory)</c> and <c>GetOrAdd(key, cachedFactory)</c> are the
+    /// same token in argument position, so a test suite that only proved the positive would not
+    /// distinguish this from the casing heuristic it exists to replace (#538).
+    /// </remarks>
+    [TestFixture]
+    public sealed class CacheFactoryAnalyzerTests
```
wallstop, line 27: This is great, but I don't think it belongs in our proto generator. This should be somewhere else.

Generator~/WallstopStudios.UnityHelpers.Proto.Generator/WProtoDiagnostics.cs
```diff
@@ -384,5 +384,34 @@ internal static class WProtoDiagnostics
...
+        /// rest of the family. What it must never do is fail their build: every other diagnostic
+        /// here is an error because the alternative is an exception from inside a shipped player,
+        /// and this one reports an allocation in code that is otherwise correct. A warning is
+        /// therefore the ceiling, and a consumer who does not want it turns it off in a ruleset or
+        /// <c>.editorconfig</c>.
+        /// </para>
+        /// </remarks>
+        internal static readonly DiagnosticDescriptor CacheFactoryAllocatesPerCall =
+            new DiagnosticDescriptor(
+                "WPROTO038",
```
wallstop: This is great and useful, but it should not be using out shorthand for "wallstop proto issues" (WPROT), this needs to be something that is a unity helpers diagnostic (potentially wallstop prefixed)

Generator~/WallstopStudios.UnityHelpers.Proto.Generator/CacheFactoryAnalyzer.cs
wallstop: (or, our IDictionary overloads and extensions?)

PR #551, opened 2026-08-23
Stop two per-frame allocations

### LINKED ISSUE #546: feature(random): a non-throwing range draw, for the authored min/max pair that a designer may legitimately collapse
## The shape

`AbstractRandom.NextFloat(min, max)` / `Next(min, max)` throw `ArgumentException` when `max <= min` (`Runtime/Core/Random/AbstractRandom.cs`). That is the right contract for a programmatic range.

It is the wrong contract for the commonest consumer shape: **two `[SerializeField]` floats that a designer fills in.** Setting both ends of a spread equal is the obvious way to ask for *no* spread — a fixed spawn delay, a fixed spacing, no scatter. So the inspector's most natural "turn this off" gesture is precisely the one that raises.

## Why the throw is expensive at that call site

These draws are almost always inside a coroutine or a periodic tick. An exception there stops the loop **permanently** while the console shows a single failed operation, so the symptom is "this system quietly stopped existing" rather than "something threw".

Two live examples we just fixed downstream (Ambiguous-Interactive/IshoBoy#666): a cloud spawner with `_minSpawnDelayTime` / `_maxSpawnDelayTime` and `_minDistance` / `_maxDistance`. Authoring either pair equal would have rendered **zero clouds for the rest of the level**.

## What we did locally, and what we would rather have

We added a two-method extension in our own tree:

```csharp
public static float NextFloatInRange(this IRandom random, float low, float high)
{
    if (random == null || high <= low) { return low; }
    return random.NextFloat(low, high);
}
```

**The low bound rather than zero**, deliberately: these are a *range*, not a scatter. An author who writes `3 .. 3` means three seconds. (A symmetric `[-s, s]` whose collapse genuinely is zero is a different shape and keeps its own guard.)

## The ask

Something equivalent on `IRandom` / `AbstractRandom` — name to taste (`NextFloatInRange`, `NextFloatClamped`, `RangeOrLow`) — for `float`, `int`, and `double` if it is cheap. It is a handful of lines, it belongs next to the methods whose contract it softens, and every consumer that authors its bounds needs it.

Two things we would want written into its doc comment, because both cost us time to work out:

1. **Why it is not the default.** The throwing overload is correct for a computed range; this one exists for an *authored* one.
2. **Which bound it answers**, and why that is not zero.

If you would rather not widen the surface, an alternative that solves the same problem is making the throw's message name the two values — today it says the constraint, not the numbers, so a designer's inspector value is not in the log.

Raised from a maintainer review on Ambiguous-Interactive/IshoBoy#666.

### LINKED ISSUE #548: bug(serialization): SerializableDictionary with a collection value type loses every value silently
## The mechanism

`Runtime/Core/DataStructure/Adapters/SerializableDictionary.cs:199-203`

```csharp
[SerializeField] protected internal TKey[] _keys;
[SerializeField] protected internal TValueCache[] _values;
```

For the two-argument `SerializableDictionary<TKey, TValue>`, `TValueCache` **is** `TValue` (stated in `Runtime/Core/Serialization/JsonConverters/SerializableDictionaryConverterFactory.cs:46`).

So declaring

```csharp
[SerializeField] private SerializableDictionary<string, List<Foo>> _map;
```

backs onto `List<Foo>[]` — a nested collection. **Unity's serializer does not serialize nested collections at all.** It emits no error and no warning. The asset records every key and none of the values, and the Inspector still happily accepts edits that vanish on reload.

## Why it is worth fixing upstream rather than in each consumer

A one-to-many authored mapping is an ordinary thing to want, the declaration compiles, the Inspector looks right, and the failure is total and silent. There is nothing about `SerializableDictionary`'s surface that warns you off it — the whole point of the type is "a dictionary Unity can serialize", so the reasonable assumption is that it handles what you put in it.

The package ships no `SerializableList<T>` (`grep -rn "class SerializableList"` across the package returns nothing), so every consumer that hits this writes the same wrapper. We have: a 174-line `SerializableList<T>` with a ProtoBuf contract and a `JsonConverterFactory`, plus a 207-line test, to hold two fields.

## Ask, either half of which closes it

1. **Ship a `SerializableList<T>`** beside the dictionary — a `[Serializable]` class holding one `List<T>`. That is the entire fix, and it is what every consumer writes anyway. Shipping it upstream also means one implementation of the ProtoBuf/JSON round-trip instead of one per project.
2. **Refuse the declaration loudly.** Detect a collection-typed `TValue` in `OnBeforeSerialize` (or a static check on first use) and log an error naming the field. Slower to reach the user, but it converts total silent data loss into a message.

(1) is the better fix; (2) is worth having regardless, because a consumer can always nest one level deeper than whatever wrapper you ship.

Found while sweeping our tree for local workarounds around this package (Ambiguous-Interactive/IshoBoy#666).

### LINKED ISSUE #549: perf: IsDontDestroyOnLoad allocates a string per call, and it reads like a cheap predicate
## The code

`Runtime/Core/Extension/Partials/General.cs:43`

```csharp
public static bool IsDontDestroyOnLoad(this GameObject gameObjectToCheck)
{
    ...
    return string.Equals(gameObjectToCheck.scene.name, "DontDestroyOnLoad", StringComparison.Ordinal);
}
```

Every read of `scene.name` marshals a fresh managed string out of native memory. So this allocates once per call, forever.

## Why that matters more than it looks

The signature reads as a cheap predicate — a `bool` off a `GameObject`, no obvious cost — so it ends up in `Update`/`LateUpdate`. In our project a profiler capture attributed **112 of the 176 bytes per frame** that `BehaviourUpdate` allocates to exactly two call sites of this method. Unity's Boehm GC is non-generational and non-compacting, so a steady drip like this is the shape that actually hurts.

We now carry a local `SceneResidency` type (78 lines plus a test) whose only job is to cache the answer so the allocation happens once instead of per frame.

## Ask

Make it allocation-free. The scene's **handle** is a value type and comparing it costs nothing:

- Resolve the DontDestroyOnLoad scene's handle once, lazily, on first call — a throwaway object passed to `Object.DontDestroyOnLoad` gives it up, after which the object can be destroyed.
- Compare `gameObject.scene.handle` against it.
- Fall back to the existing name comparison only if the handle could not be resolved (edit mode, or before any DDOL object exists), so behaviour is unchanged in every case that works today.

That deletes our local cache and makes the method safe to call from the places its signature invites.

Found while sweeping our tree for local workarounds around this package (Ambiguous-Interactive/IshoBoy#666).

### LINKED ISSUE #550: bug(attributes): WGroup and WShowIf declare AttributeTargets.Property, and the layout builder only reads fields
## The mismatch

`Runtime/Core/Attributes/WGroupAttribute.cs:53`

```csharp
[AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = true, ...)]
```

The sole consumer, `Editor/Utils/WGroup/WGroupLayoutBuilder.cs:98-110`, walks

```csharp
currentType.GetFields(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly)
```

There is **no `GetProperties` call anywhere in the package's editor assembly.**

## The consequence

Putting `[WGroup]` or `[WShowIf]` on a property compiles, reads as correct, and reaches nothing. It looks byte-for-byte like the working thing one line above it on a field. Nothing draws, nothing warns.

We ended up writing a 138-line reflection fixture (`InspectorAttributePlacementTests`) whose entire job is to fail the build when someone puts one of these attributes on a property — enforcing, from a consumer project, a restriction the attribute itself advertises as allowed.

## Ask, either half

1. **Honour properties in `WGroupLayoutBuilder`** — read `GetProperties` alongside `GetFields`. The intent behind the declaration is real: a live derived read-out (a computed count, a resolved path, a current state) is genuinely worth grouping into the inspector next to the fields it derives from, and today there is no way to do that.
2. **Narrow the `AttributeUsage` to `AttributeTargets.Field`** so the compiler refuses the mistake outright.

Either one is fine; what is not fine is the current state, which is a promise nothing keeps and which each consumer has to police for itself. Whichever you pick, please apply it to `WShowIfAttribute` too — it declares the same targets.

Found while sweeping our tree for local workarounds around this package (Ambiguous-Interactive/IshoBoy#666).

wallstop:
**Why:** Four issues from a consumer's sweep of the workarounds they carry around this package, each of which cost them a type to work around.

**What:**

- `IsDontDestroyOnLoad()` and `Animator.ResetTriggers()` no longer allocate per call; both read as free and were not.
- Every ranged draw gains a non-throwing sibling, so an authored min/max pair set equal no longer kills the coroutine it sits in.
- `WUH002` warns when a Unity-serialized field resolves onto a collection of collections, which Unity drops in full and silently.
- Eight inspector attributes now accept fields only, instead of advertising properties that nothing reads.

Fixes #546
Fixes #548
Fixes #549
Fixes #550

---

> [!NOTE]
> **Medium Risk**
> Touches inspector attribute targets (`WGroup` fields-only), drawer condition resolution, and hot-path caches. New analyzer is warning-only; runtime caches need domain-reload reset correctness.
> 
> **Overview**
> Adds **WUH002**, a default-on warning when a Unity-serialized field walks onto a collection of collections (including through adapters like `SerializableDictionary<string, List<T>>`). Nested `[Serializable]` DTOs are included; plain non-Unity types are not.
> 
> **`WShowIf`** now resolves `[field: SerializeField]` auto-properties from serialized backing-field names via new `SerializedMemberNames`, so pending Inspector edits drive visibility. **`WGroup` / `WGroupEnd`** accept fields only (`[field: WGroup]` on auto-properties). Other inspector attributes declare Field|Property explicitly so targets no longer drift with Unity’s `PropertyAttribute`.
> 
> Hot paths: `Animator.ResetTriggers()` caches trigger hashes per controller; `GameObject.IsDontDestroyOnLoad()` caches scene identity instead of marshaling `Scene.name`. Relational **base-class** collection fields use typed `GetComponents`. Eight `Next*InRange` draws return the low bound on an empty authored range instead of throwing.
> 
> Also hardens the `[Preserve]` linter for comments/wrapped attributes and records related agent rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67eeec76281f3253837261f2c91f53c18718b727. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/CustomDrawers/WShowIfPropertyDrawer.cs
```diff
@@ -306,6 +306,40 @@ out shouldShow
...
+        )
+        {
+            SerializedProperty resolved = serializedObject.FindProperty(path);
+            if (resolved != null || string.IsNullOrEmpty(path))
+            {
+                return resolved;
+            }
+
+            int separatorIndex = path.LastIndexOf('.');
+            string leaf = separatorIndex == -1 ? path : path.Substring(separatorIndex + 1);
```
wallstop: Please never hardcode magic sentinels like "-1". You should explicitly check for valid values, like >= 0, here, and throughout the code base.

Runtime/Core/Attributes/WGroupAttribute.cs
```diff
@@ -55,11 +55,10 @@ namespace WallstopStudios.UnityHelpers.Core.Attributes
     /// }
     /// </code>
     /// </example>
-    [AttributeUsage(
-        AttributeTargets.Field | AttributeTargets.Property,
-        AllowMultiple = true,
-        Inherited = true
-    )]
+    // Fields only: the sole consumer, WGroupLayoutBuilder, lays out Unity SerializedProperty paths,
+    // which exist for serialized fields and never for a C# property. Declaring Property let the
```
wallstop: Are you sure? Even properties that have their backing fields serialized? Sounds like a bug.

Runtime/Core/Extension/AnimatorExtensions.cs
```diff
@@ -29,15 +47,54 @@ public static void ResetTriggers(this Animator animator)
...
+            if (!TriggerHashesByController.TryGetValue(controller, out int[] triggerHashes))
+            {
+                triggerHashes = ReadTriggerHashes(animator);
+                TriggerHashesByController.Remove(controller);
+                TriggerHashesByController.Add(controller, triggerHashes);
+            }
+
+            for (int i = 0; i < triggerHashes.Length; ++i)
+            {
+                animator.ResetTrigger(triggerHashes[i]);
```
wallstop: Why not foreach?

Runtime/Core/Extension/AnimatorExtensions.cs
```diff
@@ -29,15 +47,54 @@ public static void ResetTriggers(this Animator animator)
...
+
+            if (triggerCount == 0)
+            {
+                return Array.Empty<int>();
+            }
 
-                animator.ResetTrigger(animatorParameter.name);
+            int[] hashes = new int[triggerCount];
+            int next = 0;
+            for (int i = 0; i < parameters.Length; ++i)
```
wallstop: Why not foreach?

Runtime/Core/Helper/SerializedMemberNames.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            if (string.IsNullOrEmpty(propertyName) || IsBackingField(propertyName))
+            {
+                return propertyName;
+            }
+
+#if !SINGLE_THREADED
+            return BackingFieldNames.GetOrAdd(propertyName, BackingFieldNameFactory);
+#else
+            if (!BackingFieldNames.TryGetValue(propertyName, out string cached))
```
wallstop: Don't we have an extension for this?

Runtime/Core/Helper/SerializedMemberNames.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <returns><c>true</c> when <paramref name="fieldName"/> was a backing field.</returns>
+        public static bool TryGetPropertyName(string fieldName, out string propertyName)
+        {
+            if (!IsBackingField(fieldName))
+            {
+                propertyName = fieldName;
+                return false;
+            }
+
+            propertyName = fieldName.Substring(
```
wallstop, line 108: Are these indices absolutely valid (non-throwing)?

Editor/CustomDrawers/WShowIfPropertyDrawer.cs
wallstop: Fixed, and swept.

The five in this file are gone, and so are the other thirteen: `!= -1` -> `>= 0`, `== -1` -> `< 0`,
across `Runtime/` and `Editor/`.

| file | sites |
| --- | ---: |
| `Editor/CustomDrawers/WShowIfPropertyDrawer.cs` | 5 |
| `Runtime/Core/Extension/Sorting/IListSortPowerPlus.cs` | 6 |
| `Editor/Sprites/SpriteSheetAnimationCreator.cs` | 4 |
| `Runtime/Core/Extension/StringExtensions.cs` | 1 |
| `Runtime/Core/Extension/IReadonlyListExtensions.cs` | 1 |
| `Editor/Utils/WButton/WButtonMetadataCache.cs` | 1 |

`rg '(==|!=)\s*-1\b' Runtime/ Editor/ --glob '*.cs'` now returns nothing outside comments.

The linked-list ones in `IListSortPowerPlus` are the clearest illustration of why you are right
beyond style: `nodeIndex != -1` was asking "is it not the terminator", `nodeIndex >= 0` asks "is it
an index", and only the second is also correct if anything ever produces a different negative.

Recorded as a rule in `.llm/context.md` so the next session does not reintroduce them.

Runtime/Core/Attributes/WGroupAttribute.cs
wallstop: Sure — and you are right that the comment made it sound like a bug, because it said "never for a C#
property" and stopped there. Corrected.

`[field: WGroup(...)]` puts the attribute on the compiler-generated backing field. That is a field,
and it is the member Unity serializes, so it groups exactly like one. `AttributeTargets.Field` is
what makes that spelling work rather than what blocks it.

Proved rather than asserted, because I did not want to answer this from reading:
`WGroupLayoutBuilderTests.GroupingWorksThroughABackingFieldAttribute` mirrors
`InfiniteAutoIncludeTerminatesAtEndAttribute` exactly, one attribute target over — same
`InfiniteAutoInclude`, same `[WGroupEnd]` terminator — against an asset declared entirely with
`[field: SerializeField]` auto-properties. The group comes back holding all three serialized paths:

```text
<Primary>k__BackingField, <Secondary>k__BackingField, <Tertiary>k__BackingField
```

Green in the editor on `6000.4.6f1`, 10/0 for that fixture.

Worth saying that my **first** version of that test failed, and it was the test, not the feature: I
had left off the `InfiniteAutoInclude` the field-based fixture uses, so only the tagged member
joined the group and it looked exactly like the bug you suspected.

What `Field` does refuse is a property Unity does **not** serialize. That one has no
`SerializedProperty` path for the layout to attach to, so it could only ever compile and silently do
nothing — which is what #550 is about.

The six attributes that have Odin drawers are `Field | Property` again, per your other note.

Runtime/Core/Extension/AnimatorExtensions.cs
wallstop: No reason. Both loops are `foreach` now — this one over `int[]`, and the two in `ReadTriggerHashes`
over `AnimatorControllerParameter[]`. Nothing needed the index.

Runtime/Core/Helper/SerializedMemberNames.cs
wallstop: We do, and it is better than I realised: `DictionaryExtensions.GetOrAdd<K, V>(this IDictionary<K, V>,
K, Func<K, V>)` already dispatches to `ConcurrentDictionary`'s own overload when that is what it is
handed.

So the whole `#if SINGLE_THREADED` split in the body collapses to one call site, and only the
field's type still needs the define:

```csharp
return BackingFieldNames.GetOrAdd(propertyName, BackingFieldNameFactory);
```

The factory stays in a `static readonly` field — that is the `WUH001` shape this package's own
analyzer flagged on the first version of this file, which is the other Bugbot comment on this PR.

Runtime/Core/Helper/SerializedMemberNames.cs
wallstop: Yes, and the guarantee now lives next to the arithmetic instead of in my head.

`IsBackingField` has already established three things before the `Substring` runs: the name starts
with `<`, ends with `>k__BackingField`, and is **strictly** longer than the two affixes combined.
The length test is the load-bearing one — without it `"<>k__BackingField"` asks for a zero-length
name and anything shorter asks for a negative one.

Tested rather than argued, with the inputs that get closest to the edge:

```text
<>k__BackingField                     affixes exactly, no property name
<k__BackingField                      prefix running into the suffix
>k__BackingField                      suffix with no opening bracket
<                                     prefix alone
<>                                    brackets alone
k__BackingField                       suffix without brackets
<<A>k__BackingField>k__BackingField   doubly wrapped
<A>k__BackingField<B>k__BackingField  two concatenated
```

Every one asserts `DoesNotThrow` on all three entry points, and any name the guard *accepts* must
yield a non-empty property name. Plus `TheShortestAcceptedNameIsExactlyOneCharacterOfProperty`,
which pins the boundary itself. 18/0 in the editor.

Runtime/Core/Extension/AnimatorExtensions.cs
wallstop: Same — `foreach` in both loops here now. The second one keeps a separate `next` counter for the
write position, which is a real use of an index rather than an artefact of the loop shape.

Generator~/WallstopStudios.UnityHelpers.Analyzers/NestedCollectionAnalyzer.cs
```diff
@@ -0,0 +1,327 @@
...
+                }
+            }
+
+            if (hasSerializeField)
+            {
+                return true;
+            }
+
+            return field.DeclaredAccessibility == Accessibility.Public
+                && DerivesFromUnityObject(field.ContainingType);
```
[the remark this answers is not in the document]
wallstop: Correct, and it is the most valuable comment on this PR — that gate is right at a top-level
declaration and wrong one step in.

A DTO written the ordinary way is exactly what a dictionary value usually is:

```csharp
[Serializable]
public sealed class Dto { public List<List<int>> rows; }   // serialized by Unity, was skipped

[SerializeField] private SerializableDictionary<string, Dto> _byName;
```

The knowledge travels down the walk now, as `containerIsSerialized`: false at a top-level field,
where the containing type must derive from `UnityEngine.Object`; true once the walk has stepped into
a `[Serializable]` type reached from a serialized field, where Unity serializes public fields too.

Five cases added — the DTO directly, and through a dictionary value, the `SerializableList` wrapper,
an array and a list — plus one pinning that the relaxation does **not** leak back out to a public
field on a plain class, which is the false positive the original gate existed to prevent.

Red-green: reverting the single clause turns **5 of 27** red.

Runtime/Core/Extension/Partials/General.cs
```diff
@@ -37,21 +38,75 @@ public static bool IsNoise(this Vector2 inputVector, float threshold = 0.2f)
...
+            if (handle != 0)
+            {
+                if (isDontDestroyOnLoad)
+                {
+                    _dontDestroyOnLoadSceneHandle = handle;
+                }
+                else
+                {
+                    _sceneHandleKnownNotDontDestroyOnLoad = handle;
+                }
```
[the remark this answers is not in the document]
wallstop: Right, and the failure mode you name is the one the change existed to remove, so this needed fixing
properly rather than documenting.

Two changes:

1. **Once the DontDestroyOnLoad handle is known, it answers on its own.** That scene is unique and
   its handle is stable for the session, so anything with a different handle is not in it — no
   negative cache is consulted at all. In a game with any persistent object this is the steady
   state after the first call.
2. **Until then, four handles instead of one.** Before any DontDestroyOnLoad object has been seen —
   a game that has none, or the frames before it is created — a fixed four-entry set of
   known-not-DDOL handles replaces the single slot, so additively loaded scenes stop evicting each
   other. A linear scan of four ints is cheaper than a hash and allocates nothing.

Both cleared at `SubsystemRegistration`, as before, so a project with domain reload disabled cannot
carry a handle into a session that renumbers its scenes.

Runtime/Core/Helper/SerializedMemberNames.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </remarks>
+        public static string BackingFieldFor(string propertyName)
+        {
+            if (string.IsNullOrEmpty(propertyName) || IsBackingField(propertyName))
+            {
+                return propertyName;
+            }
+
+#if !SINGLE_THREADED
+            return BackingFieldNames.GetOrAdd(propertyName, BuildBackingFieldName);
```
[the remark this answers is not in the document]
wallstop: Already fixed — in `033434fc`, one commit after the `0b4f7dc9` this reviewed.

Worth recording that the package's **own** analyzer is what caught it: `WUH001` fired on
`SerializedMemberNames.cs(63,61)` during `npm run typecheck:unity`, which is the first time one of
these rules has reported a defect in code written the same session. It is now held in a
`static readonly Func<string, string>` field.

The same review round also collapsed the `#if SINGLE_THREADED` split around it to one call site,
because `DictionaryExtensions.GetOrAdd` takes `IDictionary` and dispatches to `ConcurrentDictionary`'s
own overload.

Editor/CustomDrawers/WShowIfPropertyDrawer.cs
```diff
@@ -747,10 +794,10 @@ out List<int> indices
                 {
                     name = raw.Substring(0, bracket);
                     int cursor = bracket;
-                    while (cursor < raw.Length && (cursor = raw.IndexOf('[', cursor)) != -1)
+                    while (cursor < raw.Length && (cursor = raw.IndexOf('[', cursor)) >= 0)
```
wallstop, line 797: We should order all comparisons, whenever possible, left-to-right, as in, only use "<" and "<=" symbols, for simplicity, here and throughout the code base.

Runtime/Core/Extension/Partials/General.cs
```diff
@@ -37,21 +38,108 @@ public static bool IsNoise(this Vector2 inputVector, float threshold = 0.2f)
...
+            }
+
+            if (isDontDestroyOnLoad)
+            {
+                _dontDestroyOnLoadSceneHandle = handle;
+            }
+            else
+            {
+                SceneHandlesKnownNotDontDestroyOnLoad[_nextKnownSceneHandleSlot] = handle;
+                _nextKnownSceneHandleSlot =
```
wallstop: I think we have a positive mod function for this, or just wrapped increment, or similar - please utilize our math helpers throughout.

wallstop: We do — `WallMath.WrappedAdd(ref _nextKnownSceneSlot, 1, KnownSceneCapacity)` now, instead of the
open-coded `% KnownSceneHandleCapacity`.

Worth noting it is not only a tidiness win: `WrappedAdd` widens to `long` before wrapping, because
adding near `int.MaxValue` overflows to a negative that is not congruent to the real sum, and it
routes through `PositiveMod` so a negative increment lands in `[0, max)` rather than staying
negative. `%` does neither. My version happened to be safe because the increment is 1 and the
capacity is 4 — which is exactly the reasoning that stops being true the next time someone edits it.

Swept the file while I was in it; this was the only open-coded wrap in the change.

Editor/CustomDrawers/WShowIfPropertyDrawer.cs
wallstop: Applied to everything this PR authored — including the eighteen sentinel conversions from the round
before, which now read `0 <= index` rather than `index >= 0`. So `0 <= nodeIndex`,
`0 <= left.prev`, `BackingFieldPrefix.Length + BackingFieldSuffix.Length < fieldName.Length`,
`MaxSerializationDepth <= depth`, `!(low < high)`, and the rest.

Recorded in `.llm/context.md` so new code lands in that form.

**The repo-wide half is filed as #554 rather than done here**, and I want to be explicit about why
rather than quietly scoping it out. I measured it: **~2,900 candidate lines across 281 files**, the
worst being `SpriteSheetAlgorithms.cs` at 145 and `Buffers.cs` at 105. Three reasons that is its own
PR:

1. **Swapping operands changes evaluation order.** `Next() > Peek()` and `Peek() < Next()` call them
   in the opposite order. Anything with a side effect, an `out`, or an assignment inside the
   comparison — `while (0 <= (index = input.IndexOf(...)))`, which is in this diff — has to be read
   rather than rewritten.
2. **`>` is not always a comparison.** Generic closers, lambda arrows, shifts, and pattern syntax
   (`{ Count: > 0 }`) all carry the token. My first count over-reported for exactly this reason.
3. **A relational pattern cannot be flipped at all.** `c is >= 'a' and <= 'z'` has no left operand to
   move, and there are a few dozen.

#554 asks for a Roslyn-based linter first — it can tell a comparison from a generic closer and can
refuse to swap operands that are not provably side-effect-free — then the bulk conversion driven by
its output, so the two cannot disagree. That is also what keeps the rule from decaying, the same way
the `-1` rule needs one. Happy to take either half.

Issue #556, opened 2026-08-24
ci: three gates could pass without checking anything, and only one had a way to notice


wallstop:
## What happened

Session 221 hit **three** gates that report success while checking nothing, in one session, by
accident. None of them was found by running it — all three are silent by construction, because the
tree they scan is clean, and "found nothing" and "looked at nothing" produce identical output.

1. **`lint-tests.ps1 -Paths a b c` linted only `a`.** `pwsh -File` binds the first token to `-Paths`
   and drops the rest unless a `ValueFromRemainingArguments` sibling catches them. Six scripts were
   missing it. `.llm/context.md` documented that exact multi-file invocation. Measured: `-Paths`
   with all 1,118 test files reported **0** advisories where the same files report **72**.
   Fixed in #555, enforced by a new `PWS005`.
2. **A pre-filter in `test-shell-portability.sh` disabled all six sections it was added to.**
   `$( )` strips trailing newlines, so the last candidate never matched the membership test — and
   the probe sorted last for every token. Caught only by injecting a violation per section and
   diffing against the unmodified script. Fixed in #555, with a self-check in the helper.
3. **A `--fix` in the new comparison-direction linter corrupted an interpolated string.** This one
   was caught, but only because the result did not compile. Had it compiled, nothing would have
   said so.

## Why this is its own ticket

Each fix is local, and the class is not. The property that matters is: **can this gate go red?** For
a scanner over a clean corpus, a green run is not evidence that it can. Three of our gates could not,
and we have on the order of sixty.

Prior art in this repository is the same shape and was written down at the time:
`test-shell-portability.sh` already carried a comment saying each replaced pattern "was verified to
select exactly the same lines as the grep it replaces, over the real corpus PLUS a synthetic line
per pattern -- the corpus contains no violations, so agreement on it alone would only prove the
absence of false positives, not that the checks still fire." That is exactly the discipline that was
then not applied to the next change to the same file.

## Ask

1. **Inventory the gates whose corpus is clean by construction** — every linter in `lint:repo` and
   `validate:tests` that scans the tree and reports violations. For each, does a self-test inject a
   violation and assert it is reported, or does it only assert a clean run passes?
2. **Add the missing red halves.** The pattern already exists in several self-tests
   (`test-lint-csharp-naming.ps1`, `test-lint-comparison-direction.js`, the new `PWS005` probe):
   write a fixture that violates, assert non-zero exit and the file/line in the message.
[...]

Issue #543, opened 2026-08-23
chore: Make local gates extremely cheap


wallstop:
They should run in <30s, ideally <10s. Right now they take 10+minutes.

wallstop:
## `validate-lint-error-codes` halved, and `shell-portability` profiled (session 219)

### The check that was already diagnosed: 126.6 s -> 62.6 s

#540 attributed `validate-lint-error-codes` to "nine pwsh start-ups plus nine cspell start-ups...
Neither start-up is work." **Half of that is wrong, and it is the half that looks obvious:**

| | idle machine |
| --- | ---: |
| `pwsh -NoProfile -Command exit` | **0.16 s** |
| `cspell --version` | **8.13 s** |
| `cspell stdin` over two words | **8.13 s** |

Nine pwsh start-ups are 1.4 s. And `--version` costs exactly what real work costs, because all of it
is process start-up and none of it is spelling.

`scripts/validate-lint-error-codes.ps1` spawned cspell **twice** -- once for a `--version`
reachability probe, once to do the work. The probe proved nothing new: `run-node-bin.js` already
exits 127 with an actionable message when the package or its bin entry is missing. So it was a full
8.13 s to learn nothing, on every run, and the contract test runs the validator nine times.

Interleaved `A,B,A,B` on an idle machine:

| | before | after |
| --- | ---: | ---: |
| validator alone | 17.17 / 17.12 / 17.12 s | 8.96 / 9.93 / 10.50 s |
| whole check | 133.5 / 126.6 s | 85.6 / **62.6 s** |

Nine of nine assertions pass in every run, so coverage is identical and only the time moved. The
133.5 s reproduces #540's recorded 134.5 s, which is what makes the after-number believable.

Swept as a class: `scripts/install-hooks.sh` launched every tool twice, once in an `if` to test it and
once in the message to print its version, so hook installation paid 16 s of cspell start-up for one
line of output. Shipped in #545.

### `shell-portability`: profiled, and it is neither of the two candidates

#540 asked "whether the cost is the walk or a per-file subprocess". **It is neither. It is a
subprocess per _line_.**

First, the recorded number needs correcting: 122.9 s came from a `--jobs 8` run. On an idle machine
(`loadavg` 1.48) the check is **82.31 s**. That is #540's own scepticism about contended per-check
numbers, confirmed.

`scripts/tests/test-shell-portability.sh` walks every tracked shell file line by line in bash, and
[...]

Issue #509, opened 2026-08-20
random: three generator improvements blocked on a seeded-stream decision (weak-half output, no NextUlong override, RomuDuo divergence)


wallstop:
Three separate improvements to the PRNG roster share one blocker: **each changes the sequence a given
seed produces.** `docs/features/utilities/random-generators.md` promises "reproducible results for
replays and testing" and tells users to "seed generators explicitly when reproducibility matters",
so shipping any of them in a minor release would silently change a shipped game's procedural world
or replay. They need a deliberate decision — a major version, or an opt-in define — which is why
#508 fixed only the documentation half of #285 and left these.

Evidence is in the #285 thread; the summary is below.

## 1. `XoroShiroRandom` returns the weak half of its own output

The generator is `xoroshiro128+` (confirmed against `xoroshiro128plus.c`: `s0 + s1`, `rotl 24`,
`s1 << 16`, `rotl 37`). Its `NextUint()` ends `return (uint)result;` — the **low** 32 bits. The
reference implementation's own header says the generator "passes all tests we are aware of except for
the four lower bits, which might fail linearity tests", and recommends the upper bits.

`AbstractRandom.NextBool()` then reads bit 0 of that value — the single weakest bit — and the
benchmark table records `XoroShiroRandom` at 761M `NextBool`/s, so that path is well exercised.

Fix: `return (uint)(result >> 32);`. Costs nothing.

## 2. No generator overrides `NextUlong()`

Verified across all 19 files: every one overrides `uint NextUint()`, none overrides `ulong NextUlong()`.
The base implementation is:

```csharp
public virtual ulong NextUlong()
{
    uint upper = NextUint();
    uint lower = NextUint();
    return ((ulong)upper << 32) | lower;
}
```

Five generators — `WyRandom`, `XoroShiroRandom`, `RomuDuo`, `SplitMix64`, `BlastCircuitRandom` —
compute a full 64-bit word internally, **discard 32 bits**, and are then asked for two of them to
rebuild one. So every `NextUlong`, `NextLong`, `NextDouble` and `NextUlong(max)` costs two generator
steps where one would do.

The package's own `docs/performance/random-performance.md` shows the shape. `NextDouble` ops/s:

| generator | native word | NextDouble ops/s |
| --- | --- | ---: |
| `SplitMix64` | 64 | 297,700,000 |
| `LinearCongruentialGenerator` | 32 | 296,300,000 |
[...]

PR #557, opened 2026-08-24
Bind the enabled relational component

wallstop:
**Why:** A single relational field bound the disabled component when `IncludeInactive = false`, and eight issues had measurable work waiting.

**What:**

- Fix `[SiblingComponent]`/`[ChildComponent]`/`[ParentComponent]` binding a disabled candidate ahead of the enabled one (#529).
- Let Unity's own type query resolve interfaces: 1.35x-2.49x, and three dead reflection type-tests removed (#529).
- Stop `ProtoSerialize(value, ref buffer)` allocating a second full payload, plus three per-call costs (#504).
- Group a `DisjointSet` in one pass: 2.4x-3.2x at the call site (#309).
- Implement published romuDuo, with the shipped algorithm run through the same 8GB battery (#509, #286).
- Make ten gates able to fail, with the fixtures that prove it (#556, #543, #435).

---

> [!NOTE]
> **Medium Risk**
> Touches relational assignment, protobuf/JSON serialization, RNG stream compatibility (RomuDuo), and array-pool semantics across runtime hot paths; tooling changes are broad but mostly fail-closed with new regressions tests.
> 
> **Overview**
> Fixes **`[SiblingComponent]` / `[ChildComponent]`** with `IncludeInactive = false` picking a **disabled** candidate when an enabled one exists behind it ([#529](https://github.com/Ambiguous-Interactive/unity-helpers/issues/529)). Resolution now flows through **`TryFirstMatchingComponent`** (bool + `out`) so results are not dropped via Unity’s implicit `bool` conversion, and interface/base-type fields use Unity’s typed **`GetComponents`** instead of scanning every component with reflection.
> 
> **Serialization and allocation:** **`ProtoSerialize(..., ref buffer)`** for serializable collections writes into the caller buffer without an extra full copy; wrapper types/constructors are cached per `T`. JSON paths drop per-value **`WGuid`** string allocations, **`AnimationCurve`** keyframe boxing, and fix **`SphericalHarmonicsL2Converter`** using **`stackalloc`** after a pooled-buffer lifetime bug. **`DisjointSet.TryGetAllSets()`** builds result lists in one pass (dense root index, **`SystemArrayPool`**). Several hot paths switch from exact-size **`WallstopArrayPool`** to **`SystemArrayPool`** where APIs accept “at least” length; **`GetWorldCorners`** iteration is bounded to four elements.
> 
> **`RomuDuo`** now matches **published romuDuo** (seed streams change vs 3.5.1; PractRand manifest/docs updated).
> 
> **Tooling ([#556](https://github.com/Ambiguous-Interactive/unity-helpers/issues/556)):** **`PWS005`** requires both **`ValueFromRemainingArguments`** and **`[CmdletBinding(PositionalBinding = $false)]`** on scripts with array parameters; Unity runner scripts refuse stray CLI args. Repo-wide linters and **`extract-code-samples`** **fail** on empty/missing corpus instead of exiting 0; contract tests (**`test-empty-corpus-gates`**, **`test-extract-code-samples`**, expanded PWS005 fixtures) and **`build-practrand.sh`** self-test against a known-failing generator. Agent docs add **`unity-api-costs`** and devcontainer fixture-selection guidance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62f2d98f64a08cec0d2b56b8d720436620620427. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Attributes/BaseRelationalComponentAttribute.cs
```diff
@@ -974,44 +906,28 @@ out List<Component> components
                 return true;
             }
 
+            // The first candidate failed a filter (or there was none), so every candidate has to be
+            // considered. The typed query returns them all -- including interface implementers, which
+            // Unity resolves natively -- so an empty result here means no candidate exists and the
+            // former sweep of every component on the object could only have confirmed that.
             if (scratch != null)
             {
                 scratch.Clear();
```
wallstop: Do we need to clear? Will GetComponents clear for us? Here and elsewhere (and anywhere we explicitly clear buffers before handing them off)?

Runtime/Core/DataStructure/DisjointSet.cs
```diff
@@ -244,52 +245,58 @@ public List<List<int>> TryGetAllSets(List<List<int>> results)
...
+                return results;
             }
 
-            foreach (List<int> scratch in setMap.Values)
+            // A root is an index into _parent, so a dense array indexed by root answers "which
+            // result list is this set's" without hashing -- and the element goes straight into its
+            // destination, where the previous shape gathered every element into a per-root scratch
+            // list and then copied all of them a second time. The slot holds resultIndex + 1 so a
+            // zero-filled array already means "unseen" and no sentinel fill pass is needed.
+            int[] rootToResult = ArrayPool<int>.Shared.Rent(elementCount);
```
wallstop: Why aren't we using our own buffers? We have a system array pool with a handle concept, no try/finally needed. We should avoid using ArrayPool (from system) unless absolutely necessary, and instead, lean on our own helpers/buffers.

Runtime/Core/Attributes/BaseRelationalComponentAttribute.cs
```diff
@@ -974,44 +906,28 @@ out List<Component> components
...
-
-                return FirstMatchingComponent(
+                singleComponent = FirstMatchingComponent(
                     scratch,
                     filters,
-                    elementType,
-                    isInterface: true,
                     filterDisabledComponents
                 );
+                return singleComponent != null;
```
wallstop: Is it possible that we can return a boolean from here somehow and not null check? Here and in any other component paths? I believe unity null is expensive.

Runtime/Core/Attributes/BaseRelationalComponentAttribute.cs
wallstop: Yes — Unity clears it, and I'd assumed rather than checked. Measured on `6000.4.6f1`, pre-filling each buffer with sentinels first:

```text
GetComponents(Type, list):                prefilled 3 -> 1
GetComponents<T>(list):                   prefilled 2 -> 3   (the object's 3 components)
GetComponentsInChildren<T>(bool, list):   prefilled 3 -> 2
GetComponentsInParent<T>(bool, list):     prefilled 3 -> 2
ZERO-MATCH GetComponents(Type, list):     prefilled 2 -> 0
ZERO-MATCH GetComponentsInChildren<T>:    prefilled 2 -> 0
```

Every overload clears, and — the case that actually matters — a zero-match query clears too, so a stale result can't survive.

Swept `Runtime/`, `Editor/` and `Tests/` for a `.Clear()` within 20 lines before a `Get*Components` filling the same buffer. Five candidates, and they split three ways:

- **Redundant, removed:** this one, and `SiblingComponentFastInvoker.Collect`.
- **Load-bearing, kept with the reason at the site:** `GetComponentsOfType`'s clear guards the `isInterface && !allowInterfaces` path, which returns the buffer *without* querying; `Helpers.cs:661` guards a `target` that matches neither `switch` case. Both would be silent bugs if removed, so each now says which it is rather than looking like an oversight the next sweep should "fix".
- **False positive:** `CollisionSenses.cs` — the scan matched a `Clear()` at the end of a different method.

Recorded in `.llm/context.md` so the next sweep doesn't re-derive it.

Runtime/Core/DataStructure/DisjointSet.cs
wallstop: Fixed, and swept — it wasn't just this site. Six scoped scratch rents had the same `ArrayPool<int>.Shared.Rent` + `try`/`finally` + `Return(clearArray: true)` shape:

`QuadTree2D` (x2), `OctTree3D`, `KDTree3D`, `KDTree2D`, `DisjointSet`. All six now use `WallstopArrayPool<int>.Get(elementCount, out int[] scratch)` with a `using`.

Three concrete reasons beyond the handle, worth recording:

1. Ours returns an array of **exactly** the requested length. System's is power-of-two bucketed, so `.Length` on a rented array lies. None of the six read `.Length`, so nothing was broken — but it was one edit away.
2. Ours **clears on release**, so a rent is already zeroed. That's what let `DisjointSet` drop its `Array.Clear`, and it's what makes the `resultIndex + 1` encoding free rather than paid for with a fill pass.
3. `using` instead of `try`/`finally`.

I left `Serializer.cs` on System's pool deliberately: `PooledBufferStream` and `PooledArrayBufferWriter` return their buffer **mid-life** while growing, not at scope end, so the handle doesn't model their lifetime. Flagged rather than converted.

`.llm/context.md` now states the preference and names the one exception, so this isn't re-litigated per site.

Spatial-tree and `DisjointSet` fixtures through the MCP editor after the change: 100 fixtures, 2574 pass / 1 fail — the failure is `CacheCapacityDiagnosticTests`, an NRE from the reflection loop not running `[OneTimeSetUp]`, unrelated to these files.

Runtime/Core/Attributes/BaseRelationalComponentAttribute.cs
wallstop: Your belief is right, and by more than I expected. Measured on `6000.4.6f1`, 20M iterations, best-of-three, managed winning every round:

```text
UnityEngine.Object operator!= : 3.380 ns/op
managed reference compare     : 0.578 ns/op   -> 5.84x
```

So `FirstMatchingComponent` is now `TryFirstMatchingComponent(..., out Component match)` returning `bool`, and neither caller null-tests the answer. `TryAssignChildSingleFallback` was doing the same thing at the end of its ancestor walk — it now carries the `bool` that `TryResolveSingleComponent` already handed it instead of re-asking Unity.

There's a second reason to prefer the `Try` shape here beyond the nanoseconds: returning a `Component` from a `bool`-shaped position is *exactly* what caused this PR's original bug. `return FirstMatchingComponent(...)` from a `bool` method compiles because of `implicit operator bool`, and the found component was silently discarded. The `out` makes that unrepresentable.

Swept `Runtime/` and `Editor/` for the remaining shape — a Unity null test on a value a `Try*` call had already answered for — and found no others.

**One thing this turned up.** Adding the `[ParentComponent]` case the fix was missing made the new test *fail*, and the test was wrong, not the code: the parent paths pass `filterDisabledComponents: false`, so `IncludeInactive` gates only `activeInHierarchy` there (`ParentComponentTests` already documents this). Every parent filter is therefore per-GameObject, so all components on one ancestor pass or fail together and the discard bug **cannot** arise for parents. My CHANGELOG named all three shapes — it now names the two that are affected, and the `IncludeInactive` XML doc records the asymmetry rather than promising uniform per-component filtering.

Runtime/Core/DataStructure/DisjointSet.cs
```diff
@@ -244,52 +244,56 @@ public List<List<int>> TryGetAllSets(List<List<int>> results)
...
+            // zero-filled array already means "unseen" and no sentinel fill pass is needed.
+            //
+            // WallstopArrayPool rather than System's: it hands back an array of EXACTLY
+            // elementCount, clears on release so the rent is already zeroed -- which is what makes
+            // the resultIndex + 1 encoding free -- and disposes through the PooledArray handle, so
+            // no try/finally is needed.
+            using PooledArray<int> rootLease = WallstopArrayPool<int>.Get(
+                elementCount,
+                out int[] rootToResult
+            );
```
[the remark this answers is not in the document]
wallstop: Correct, and the pool's own class doc says it in as many words — I read the `Get` overload's remarks and the `PooledArray` disposal semantics, and not the `<remarks>` on the class directly above them:

> **⚠️ MEMORY LEAK WARNING:** This pool creates a separate pool bucket for EVERY unique size requested.
> **UNSAFE uses (will leak memory):** `Get(collection.Count)` — Every unique collection size leaks memory
> **Rule of thumb:** If you cannot enumerate ALL possible sizes at compile time, use `SystemArrayPool<T>` instead.

`elementCount` is `collection.Count`. Confirmed the mechanism too: `PoolBucketMap` is a `ConcurrentDictionary<int, PoolBucket<T>>` with no eviction.

All six sites now use `SystemArrayPool<int>.Get(elementCount, clearArray, out int[] scratch)` — still one of this package's helpers, still a `PooledArray` handle, so the review point that started this (lean on our own buffers, no `try`/`finally`) is unaffected. It's also already the established idiom here: `RTree2D`, `RTree3D`, `SparseSet` and `BitSet` all rent runtime sizes that way.

`clearArray` differs by site, deliberately:
- **`DisjointSet`: `true`.** Its slot is *read* before it is written — that read is what "unseen" means in the `resultIndex + 1` encoding — and the shared pool hands back dirty arrays.
- **The four trees: `false`.** They write every slot before reading it, which is exactly what the previous `ArrayPool.Shared.Rent` already relied on.

**Swept the rest of the repo** for the same misuse. Every other `Wallstop(Fast)ArrayPool` call site is safe and stays: `ElementByteSize` in three PRNGs, `4` in `Bounds`, `27` in the spherical-harmonics converter — all constants; and `IListSortIps4o`'s `bucketCount`, which `DetermineIps4oBucketCount` clamps to `[4, 32]` — the documented "algorithm-bounded sizes with small fixed upper limits" case.

One pre-existing site **is** the leak shape and I did not change it: `SpriteSheetExtractor.cs:6547` rents `WallstopFastArrayPool<Color32>` by `pixelCount`, with a comment stating that `GetPixels32` requires the exact length, which rules out `SystemArrayPool`'s power-of-two buckets. Editor-only and a deliberate trade-off, so it's flagged rather than silently broken.

`.llm/context.md` had a rule I added one commit earlier recommending `WallstopArrayPool` over System's — that advice was wrong and is now the bounded-versus-runtime split, naming this trap and the lesson: read a pool's **class** remarks, not just the overload you're calling.

Spatial and `DisjointSet` fixtures through the MCP editor after the change: 14 fixtures, 323 pass / 0 fail. `typecheck:unity` green in all four configurations. Re-measured `TryGetAllSets` as well — 10.05 / 40.01 / 168.34 µs, unchanged from the previous pool.

Runtime/Core/DataStructure/DisjointSet.cs
wallstop: No, I meant using our SystemArrayPool, which is a warpper. Only use WallstopArrayPool if we need precise length.

wallstop: Understood, and applied as stated — your rule is stricter than the one I'd encoded, which was "bounded sizes are fine on the exact-size pool". That was wrong twice over: it let a `collection.Count` through in the first place, and it kept several sites on `WallstopArrayPool` that had no need of a precise length.

The six spatial/`DisjointSet` sites already moved to `SystemArrayPool` when Bugbot flagged the leak. This pass applies your rule to **every remaining site**:

| site | why it moves |
| --- | --- |
| `StormDropRandom`, `PhotonSpinRandom`, `FlurryBurstRandom` | payload is bounded by an `ArraySegment`/span slice, and `BuildState` copies it out |
| `SphericalHarmonicsL2Converter` | indexes `[0,27)` and validates the count separately |
| `IListSortIps4o` (×3) | nothing reads `.Length`; it already rents its element buffer from `SystemArrayPool` two lines below |
| `Bounds.GetWorldCorners` | see below |

That leaves **exactly one** exact-size rent in the package.

### "Precise" is worth measuring, and both of my guesses were wrong

On `6000.4.6f1`:

```text
Texture2D.SetPixels32(oversized)      -> ArgumentException: size of data to be written is
                                         outside the target buffer bounds       (needs EXACT)
RectTransform.GetWorldCorners(len 8)  -> accepted                               (needs 4 OR MORE)
```

So `SpriteSheetExtractor`'s `WallstopFastArrayPool<Color32>` rent is the one justified exact-size use — it's now correct *by* your rule rather than an exception I was flagging against it. And `Bounds`, which reads as obviously needing exactly four, does not.

`clearArray: false` at every converted site: each writes before it reads. Worth noting one consequence — `IListSortIps4o` clears `bucketCounts` explicitly on the next line, which was redundant under the clear-on-release pool and is now load-bearing.

### Knowledge

The rule is now in a new [unity-api-costs](https://github.com/Ambiguous-Interactive/unity-helpers/blob/dev/wallstop/session-222/.llm/skills/unity-api-costs.md) skill, alongside the other two measured findings from this review round. `.llm/context.md` had reached 483 of its 500 lines almost entirely from this session — and `agent-preflight` fails on the CRITICAL band rather than merely warning — so the long-form rules moved into the skill and `context.md` keeps a one-line pointer. 461 lines now.

15 fixtures over the converted sites: 339 pass / 0 fail. `typecheck:unity` green in all four configurations.

Runtime/Core/Extension/Partials/Bounds.cs
```diff
@@ -36,8 +36,12 @@ public static Rect Rect(this in Bounds bounds)
 
         public static Rect GetWorldRect(this RectTransform transform)
         {
-            using PooledArray<Vector3> fourCornersResource = WallstopFastArrayPool<Vector3>.Get(
+            // SystemArrayPool: GetWorldCorners requires at least four elements, not exactly
+            // four -- measured on 6000.4.6f1, an eight-element array is accepted -- so this does not
+            // need the exact-size pool. clearArray is false because the call fills all four.
+            using PooledArray<Vector3> fourCornersResource = SystemArrayPool<Vector3>.Get(
                 4,
+                clearArray: false,
```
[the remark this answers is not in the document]
wallstop: Correct, and already fixed — this review is against `00d36500`; the fix landed in `62f2d98f` about forty minutes later.

`GetWorldRect` no longer iterates the whole array:

```csharp
const int CornerCount = 4;
using PooledArray<Vector3> fourCornersResource = SystemArrayPool<Vector3>.Get(
    CornerCount, clearArray: false, out Vector3[] fourCorners);
transform.GetWorldCorners(fourCorners);
...
// Indexed to CornerCount, NOT a foreach: this pool may hand back a LONGER array than
// requested, and folding its uninitialized tail into the bounds is exactly the bug that
// reddened every playmode leg once this stopped using an exact-size pool.
for (int i = 0; i < CornerCount; ++i)
```

Worth recording that you and CI found this independently: the four playmode legs went red on `00d36500`, and sweeping every pooled rent in `Runtime/` and `Editor/` for the shape — a `foreach` over, or a `.Length` of, a rented array — turned up exactly two real instances, this one and the spherical-harmonics converter you flagged alongside it. The other five hits were `Array.Clear(x, 0, x.Length)`, which over-clears harmlessly.

Unity Tests on `62f2d98f`: 18/18 green.

Runtime/Core/Serialization/JsonConverters/SphericalHarmonicsL2Converter.cs
```diff
@@ -59,8 +59,12 @@ JsonSerializerOptions options
                             throw new JsonException("coefficients must be an array");
                         }
                         using (
-                            PooledArray<float> lease = WallstopFastArrayPool<float>.Get(
+                            // SystemArrayPool: every one of the 27 slots is written before it
+                            // is read, and the count is validated separately, so the exact-size
+                            // pool buys nothing here.
+                            PooledArray<float> lease = SystemArrayPool<float>.Get(
                                 27,
+                                clearArray: false,
```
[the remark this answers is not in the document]
wallstop: Correct, and already fixed — this review is against `00d36500`; the fix landed in `62f2d98f` about forty minutes later. This was the one that actually reddened CI: all four playmode legs, failing in `Tests.Serialization`.

Your diagnosis is exact — `coeffs.Length != 27` silently required a pool returning the exact requested size. Fixing it surfaced a **second, older bug in the same block**: `coeffs = tmp` assigned the rented array to a local that outlived its `using`, so the coefficients were read *after* the array had gone back to the pool. That survived only because the exact-size pool it used does not clear on release.

Both go away together with a `stackalloc`:

```csharp
const int CoefficientCount = 27;
Span<float> coeffs = stackalloc float[CoefficientCount];
bool haveCoefficients = false;
```

108 bytes, no pool, no escape, and the count tracked in a local instead of inferred from an array's length.

Serialization fixtures through the MCP editor: 52 fixtures, 541 pass / 0 fail. Unity Tests on `62f2d98f`: 18/18 green.

The verification gap that let this reach CI is also recorded now: I ran the fixtures whose *names* matched the converted types and got a clean 339/0, but the regression lives in `JsonConverterTests`, which covers this converter without being named after it. `.llm/skills/unity-devcontainer-testing.md` now says to select fixtures by the namespace of the file you changed, never by the names of the types you touched.

PR #571, opened 2026-08-25
Keep a buffed attribute's value across a save

### LINKED ISSUE #567: perf: a static authoring mistake is warned about on every effect application, JSON-serializing the whole effect
## Summary

`EffectHandler.InternalApplyEffect` warns when an effect is `Instant` but carries periodic or
behaviour data:

```csharp
private void InternalApplyEffect(AttributeEffect effect)
{
    if (effect.durationType == ModifierDurationType.Instant && RequiresHandle(effect))
    {
        this.LogWarn(
            $"Effect {effect:json} defines periodic or behaviour data but is Instant. ..."
        );
    }
```

The condition depends only on `effect`. It is a **static property of the asset**, reported on the
**per-application** path -- so a single mis-authored effect warns on every hit, every tick, every
time it is applied for the whole session.

## Why it costs more than a normal warning

Two costs, both paid per application:

1. **`{effect:json}` is `GenericToString()`**, which serializes every public field and property of
   the effect to JSON. It is inside the interpolated string, so it runs whenever the call runs.
2. **Unity captures a managed stack trace for the log.** Measured on `6000.4.6f1` in
   [#564](https://github.com/Ambiguous-Interactive/unity-helpers/issues/564): 178.4 us with a stack
   trace against 13.3 us without.

Neither is stripped in the editor or a development build. `WallstopStudiosLogger.LogWarn` is
`[Conditional]` on `ENABLE_UBERLOGGING` / `WARN_LOGGING` / `DEVELOPMENT_BUILD` / `DEBUG` /
`UNITY_EDITOR`, so a non-development player pays nothing -- but the editor, where this is authored
and profiled, pays all of it.

## Relationship to #564

#564 was the same shape one layer down: a diagnostic on a routine path whose triggering condition is
a normal state. That one was fixed by `stackTrace: false`, which is now available on all three
severities. This site is different in one way that makes it worth its own ticket: the *condition* is
static, so the right fix is probably not to make the log cheaper but to move it off the hot path
entirely.

A sweep in session 224 found 14 `LogError` / `LogWarn` calls in `Runtime/` inside a loop or a
per-frame method. Thirteen fire on genuine misconfiguration that a correct scene never reaches (a
null `CosmeticEffectData`, for instance) and were deliberately left alone. This is the one that
does not fit that description.

## Options, in rough order of preference

1. **Validate on the asset.** `AttributeEffect.OnValidate` knows everything this condition needs, and
   an authoring mistake belongs in the Inspector, not in the play-mode console. The apply path then
   carries no diagnostic at all.
2. **Report once per effect**, keyed on the asset, so the message survives but the repetition does
   not.
3. **`stackTrace: false` plus dropping `:json`** to the effect's name. Cheapest change, keeps the
   behaviour, and still pays per application.

## Acceptance

1. A mis-authored `Instant` effect still tells the author something actionable.
2. Applying that effect N times does not cost N JSON serializations of the effect.
3. A measurement of the apply path before and after, with a control that moved.

### LINKED ISSUE #569: bug: an Attribute's buffed value does not survive a JSON round trip on 6000.4.6f1, an editor CI does not run
## Summary

Two EditMode tests in `Tests/Runtime/Tags/AttributeSerializationTests.cs` fail on **`6000.4.6f1`**,
which is not one of the four editors CI runs:

```text
JsonRoundTripKeepsTheModifiedValueRatherThanRecalculating:  Expected: 125.0f  But was: 100.0f
ClearCacheAfterJsonRoundTripFallsBackToTheBaseValue:        Expected: 125.0f  But was: 100.0f
```

`.github/unity-versions.json` is `["2021.3.45f1", "2022.3.45f1", "6000.3.16f1", "6000.5.2f1"]`.
`6000.4.6f1` sits between the last two and nothing covers it, so this is invisible to the matrix.

## What the tests assert

```csharp
Attribute attribute = new(100f);
attribute.Add(25f);                                   // CurrentValue == 125

string json = Serializer.JsonStringify(attribute);
Attribute deserialized = Serializer.JsonDeserialize<Attribute>(json);

Assert.AreEqual(100f, deserialized.BaseValue);        // passes
Assert.AreEqual(125f, deserialized.CurrentValue);     // fails: 100
```

The comment in the test says why the cached value matters: *"Modifications are not serialized.
Recalculating on load would report 100 for an attribute that was written while buffed, so the
written value is kept instead."* On `6000.4.6f1` the written value is **not** kept, so a save taken
while a character is buffed loads with the buff silently gone.

## It is pre-existing, and that was measured rather than assumed

Found while running the Tags fixtures for
[#567](https://github.com/Ambiguous-Interactive/unity-helpers/issues/567) through the MCP bridge.
Three hypotheses were tested and the first two were refuted:

| hypothesis | result |
| --- | --- |
| the reflection harness skips `[OneTimeSetUp]` (a known trap) | **refuted** -- invoking `CommonTestBase.CommonOneTimeSetUp` base-first changed nothing |
| an earlier serializing probe poisoned the domain | **refuted** -- still fails after `EditorUtility.RequestScriptReload()` in a fresh domain |
| it is pre-existing rather than caused by the branch | **confirmed** -- `git checkout main -- Runtime/ Tests/`, recompile, and both fail identically, with a gate asserting the loaded assembly did NOT carry the branch's edit |

So: not a harness artifact, not branch-caused, and not visible to CI.

## Why it matters beyond the tests

Whatever makes the cached `CurrentValue` fail to survive a JSON round trip on this editor is a
save-data correctness question, not a test question. The failure mode is quiet: the attribute loads
with a plausible value that is simply the wrong one.

## Acceptance

1. The cause is identified -- most likely a serializer behaviour that differs between `6000.3`/`6000.5`
   and `6000.4`, since the same sources pass on the four tested editors.
2. Either the code is fixed, or the version-specific behaviour is documented and the test states it.
3. **A decision on the matrix.** Either `6000.4.x` joins `.github/unity-versions.json`, or the gap is
   recorded deliberately -- the same class as
   [#553](https://github.com/Ambiguous-Interactive/unity-helpers/issues/553), where an answer was
   confidently wrong for every editor that ships.

## Note for whoever picks this up

`CosmeticEffectComponentTests.OnDestroyInvokesRemoveEffectForTrackedTargets` also fails under the MCP
harness, but that one IS an artifact: the harness runs in EditMode where `Destroy` is not
`DestroyImmediate`, and the run logged `Destroy may not be called from edit mode!`. It is not part of
this issue.

wallstop:
**Why:** An attribute deserialized while buffed reported its base value in the editor, and a single effect-authoring mistake was re-reported on every application.

**What:**

- `Attribute.CurrentValue` keys its cache on the base value it was calculated from, instead of discarding the cache whenever `Application.isPlaying` is false.
- An `AttributeEffect` authoring mistake is reported in the Inspector and once per effect, not once per application — three sites, each of which rendered the whole effect to JSON at 20.5 us.
- New EditMode fixtures live in an editor-only assembly, which is the only kind Unity's EditMode runner executes.
- `validate-lint-error-codes` starts its nine scenarios together: 81.3 s -> 16.2 s.
- `lint-tests` caches the directories it walks looking for an asmdef: 92.0 s -> 25.7 s, byte-identical output.

Fixes #569
Fixes #567

---

> [!NOTE]
> **Medium Risk**
> Changes core attribute caching and effect-application logging paths used in editor and runtime; behavior is covered by new EditMode tests but affects saved/deserialized attribute values.
> 
> **Overview**
> Fixes **`Attribute.CurrentValue`** so edit mode no longer throws away the cache whenever `Application.isPlaying` is false. The getter now reuses the cache only when `_calculatedFromBaseValue` still matches `_baseValue`, so JSON round-trips keep a buffed value and direct `_baseValue` writes (Inspector / prefab / undo) trigger recalculation. EditMode tests were added because Unity’s EditMode runner only runs editor-only assemblies.
> 
> **`AttributeEffect` / `EffectHandler`** misconfiguration is reported once per asset instead of on every apply: Instant effects with periodic/behavior data and unassigned cosmetic slots warn in **`OnValidate`** and at most once at runtime, using effect **names** instead of `{effect:json}` (and the duplicate Instant warning in `InternalApplyEffect` is removed).
> 
> Supporting changes: LLM skills/docs now say **`foreach` boxing comes from interface-typed iteration**, not concrete `List<T>`; **`test-validate-lint-error-codes.ps1`** runs nine scenarios in parallel via **`ProcessStartInfo.ArgumentList`** with space-bearing temp paths; **`lint-tests.ps1`** caches asmdef walks and uses **`EnumerateFiles`** for speed; effects docs and CHANGELOG entries for #567 / #569.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 683adfa443886b11424ffdc21aa3aa20b129db5a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Runtime/Tags/AttributeEffect.cs
+++ b/Runtime/Tags/AttributeEffect.cs
@@ -183,6 +183,9 @@ namespace WallstopStudios.UnityHelpers.Tags
         [NonSerialized]
         private bool _instantWithHandleDataReported;
 
+        [NonSerialized]
+        private bool _unassignedCosmeticReported;
+
         /// <summary>
         /// Gets whether this effect is <see cref="ModifierDurationType.Instant"/> yet carries
         /// periodic or behaviour data. Instant effects return no handle, so neither can ever run.
@@ -212,13 +215,61 @@ namespace WallstopStudios.UnityHelpers.Tags
             return true;
         }
 
+        /// <summary>
+        /// Gets whether <see cref="cosmeticEffects"/> holds an unassigned entry. Those entries
+        /// cannot be instanced, so they are skipped every time the effect is applied.
+        /// </summary>
+        internal bool HasUnassignedCosmeticEffect
+        {
+            get
+            {
+                for (int index = 0; index < cosmeticEffects.Count; ++index)
+                {
+                    if (cosmeticEffects[index] == null)
+                    {
+                        return true;
+                    }
+                }
+
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Returns <c>true</c> the first time it is called on an effect holding an unassigned
+        /// cosmetic entry, and <c>false</c> forever after.
+        /// </summary>
+        /// <remarks>
+        /// Same shape as <see cref="ShouldReportInstantWithHandleData"/>: a static property of the
+        /// asset, previously reported once per unassigned entry per application.
+        /// </remarks>
+        internal bool ShouldReportUnassignedCosmeticEffect()
+        {
+            if (_unassignedCosmeticReported || !HasUnassignedCosmeticEffect)
+            {
+                return false;
+            }
+
+            _unassignedCosmeticReported = true;
+            return true;
+        }
+
         private void OnValidate()
         {
             _instantWithHandleDataReported = false;
+            _unassignedCosmeticReported = false;
             if (IsInstantWithHandleData)
             {
                 this.LogWarn(
                     $"Effect {name} defines periodic or behaviour data but is Instant. These features require a Duration or Infinite effect.",
+                    stackTrace: false
+                );
+            }
+
+            if (HasUnassignedCosmeticEffect)
+            {
+                this.LogWarn(
+                    $"Effect {name} has an unassigned CosmeticEffectData entry, which cannot be instanced and is skipped.",
                     stackTrace: false
                 );
             }
--- a/Runtime/Tags/AttributeEffect.cs
+++ b/Runtime/Tags/AttributeEffect.cs
@@ -223,9 +223,9 @@ namespace WallstopStudios.UnityHelpers.Tags
         {
             get
             {
-                for (int index = 0; index < cosmeticEffects.Count; ++index)
+                foreach (CosmeticEffectData cosmeticEffect in cosmeticEffects)
                 {
-                    if (cosmeticEffects[index] == null)
+                    if (cosmeticEffect == null)
                     {
                         return true;
                     }
--- a/Runtime/Tags/AttributeEffect.cs
+++ b/Runtime/Tags/AttributeEffect.cs
@@ -180,6 +180,50 @@ namespace WallstopStudios.UnityHelpers.Tags
         [JsonIgnore]
         public List<EffectBehavior> behaviors = new();
 
+        [NonSerialized]
+        private bool _instantWithHandleDataReported;
+
+        /// <summary>
+        /// Gets whether this effect is <see cref="ModifierDurationType.Instant"/> yet carries
+        /// periodic or behaviour data. Instant effects return no handle, so neither can ever run.
+        /// </summary>
+        internal bool IsInstantWithHandleData =>
+            durationType == ModifierDurationType.Instant
+            && ((periodicEffects is { Count: > 0 }) || (behaviors is { Count: > 0 }));
+
+        /// <summary>
+        /// Returns <c>true</c> the first time it is called on a misconfigured effect, and
+        /// <c>false</c> forever after.
+        /// </summary>
+        /// <remarks>
+        /// The condition is a static property of the asset, so reporting it on the per-application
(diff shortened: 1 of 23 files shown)


Runtime/Tags/AttributeEffect.cs
```diff
@@ -180,6 +180,101 @@ public sealed class AttributeEffect : AttributeEffectBase, IEquatable<AttributeE
...
+
+        /// <summary>
+        /// Gets whether <see cref="cosmeticEffects"/> holds an unassigned entry. Those entries
+        /// cannot be instanced, so they are skipped every time the effect is applied.
+        /// </summary>
+        internal bool HasUnassignedCosmeticEffect
+        {
+            get
+            {
+                for (int index = 0; index < cosmeticEffects.Count; ++index)
```
wallstop: Why can't we use a foreach loop?

Issue #505, opened 2026-08-20
[Bug]: pre-push validation takes 9+min


wallstop:
### Description

This is way too long. It needs to be <30s, ideally <10s

### Steps to Reproduce

Run pre-push + "fast repo contract suite" + e2e fast suite

### Expected Behavior

<30s, even in a devcontainer

### Actual Behavior

10+min

### Severity

Critical (crash, data loss)

### Unity Version

Unity 6.5 (6000.5)

### Package Version

3.5.1

### Platform

- [ ] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Regression

- [ ] Yes, this is a regression (worked in a previous version)

wallstop:
Session 207 measurement update, linked to PR #507: https://github.com/Ambiguous-Interactive/unity-helpers/pull/507

On this devcontainer's 9p workspace:
- npm run validate:prepush: about 1.5 s;
- npm run lint:tests: about 94 s;
- EOL validation: about 12 s.

An ad hoc eight-way profile made the slowest checks about 140 s and caused the sync-script contract test to race, though it passed sequentially. That rules out naive “parallelize everything.”

The next implementation needs to:
1. name the exact fast repo-contract/e2e command whose under-30-second budget is being enforced;
2. profile that intended command sequentially on the 9p mount;
3. preserve every exhaustive check in CI before moving it out of local pre-push;
4. group only checks proven independent; and
5. keep deterministic per-check diagnostics.

Moving a slow check without verified CI coverage would game the target rather than improve it. PR #507 records the measured classification but makes no speculative hook change.

wallstop:
## Measured progress, and an explicit statement that the target is not met (session 216)

#535 takes 5.1 s off any serial pass over the repo lint suite (`lint-license-headers` 4925 -> 3014 ms, `lint-duplicate-usings` 10318 -> 8658 ms, `lint-doc-links` 10002 -> 8438 ms, all byte-identical output).

That does not bring pre-push near the `<30 s` this issue asks for, and it is not claimed to. Same finding as on #425: after the I/O conversions the dominant remaining cost is the interpreted PowerShell scanners themselves — `lint-duplicate-usings` still spends 8.6 s walking 21.6 M characters line by line. Getting to the target is a rewrite of those scanners or removing them from the pre-push path, not further tuning of how they enumerate files.

wallstop:
[...]

Issue #431, opened 2026-08-14
docs: Weekly / release-scheduled perf/benchmark runs


wallstop:
A few things:

1. We should utilize similar techniques to https://github.com/Ambiguous-Interactive/DxMessaging/commit/e18776ec4cedb4b19cb2a8f5dcd4a010334d8a17 to ensure that our benchmarks, where relevant, are measuring from a clean place - like not polluting the heap
2. We should have either a periodic run (weekly) or a run that is triggered on our release-prepare that refreshes all of our benchmark and performance data. It needs to be ran on ELI-MACHINE
3. Our performance data and similar documentation needs to be more user-focused. There are lots of "operator" (me) intended phrasing, like what to do when a missing upstream repo comes back online.

wallstop:
## The EditMode benchmark legs were running zero tests

Both scheduled runs of `unity-benchmarks.yml` failed (28938221812, 28514563014), and every leg
failed at `Run Unity Test Runner` + `Verify tests actually ran`. For the four **editmode** legs
that is now explained.

`unity-benchmarks.yml` gave them `Tests.Runtime.Performance` and `Tests.Runtime.Random`. Both
asmdefs are platform-neutral (`includePlatforms: []`), and Unity's EditMode runner takes only
assemblies flagged `EditorAssembly` — see
[#570](https://github.com/Ambiguous-Interactive/unity-helpers/issues/570). So those legs were handed
two assembly names and ran nothing, which is exactly what `Verify tests actually ran` is for.

Confirmed from CI logs rather than inferred: in the `2021.3.45f1` **editmode** leg of the last
`unity-tests.yml` run on `main`, `Tests.Runtime.Random` appears **3** times — the assembly list
being echoed — against **2,375** times in the playmode leg, and `PoolLifecycleHooksTests` runs
**94** times in playmode and **0** in editmode.

#574 makes the benchmark matrix playmode-only and carries the thorough Random configuration
(`Performance;Stress;Fast`, `UH_RANDOM_SAMPLE_COUNT=12750000`, the 6600 s watchdog) onto the leg
that can execute it. That removes four licensed Unity legs per scheduled run.

**Still open here:** the four **playmode** legs also failed, in July, and their job logs have since
expired, so that half is unexplained. The next scheduled run will produce fresh evidence.

wallstop:
## Items 1 and 3 shipped; item 2 already exists and is switched off elsewhere

PR #576.

### Item 1 -- measuring from a clean place

[...]

PR #574, opened 2026-08-26
Stop naming assemblies Unity will not run in EditMode

### LINKED ISSUE #338: chore/research: Investigate uv for python
uv is great, it might help optimize python operations locally (pre-commit?) and in CI

### LINKED ISSUE #378: research: max Warnings-as-errors, static analyzers
See if we can enable warnings as errors, crank to maximum warnings, and enable as many useful static analyzers as possible for all code surfaces

### LINKED ISSUE #415: ci: every job on ELI-MACHINE fails at 'Set up runner' — the watchdog pre-job hook cannot open its global mutex
## Every job that lands on ELI-MACHINE fails before checkout

Seven of the eight Unity legs on [run 31542993311](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31542993311)
(PR #413) failed in about twelve seconds at **Set up runner**, before `Checkout`. Every one of them
was assigned to `ELI-MACHINE`; the single leg that went to `DAD-MACHINE` ran and passed. Nothing in
the repository was reached, so this is not a code failure and re-running only helps if the job is
reassigned.

## The cause, from the log

The failure is the runner administrator's **job started hook**, not the workflow:

```text
A job started hook has been configured by the self-hosted runner administrator
##[group]Run 'C:\ProgramData\DockerRunnerWatchdog\hooks\pre-job.ps1'
Unhandled exception. System.UnauthorizedAccessException: Access to the path 'Global\DockerRunnerWatchdog.Log' is denied.
   at System.Threading.MutexAcl.Create(Boolean initiallyOwned, String name, Boolean& createdNew, MutexSecurity mutexSecurity)
   at DockerRunnerWatchdog.JsonLineLog..ctor(String path, SecurityIdentifier[] authorizedSids)
   at DockerRunnerWatchdog.Program.Main(String[] args)
##[error]Process completed with exit code 1.
```

`DockerRunnerWatchdog` cannot create or open its **global** named mutex `Global\DockerRunnerWatchdog.Log`.
Creating one in the `Global\` namespace needs the `SeCreateGlobalPrivilege`, and opening an existing
one needs the ACL to grant the caller — so the usual causes are the runner service account losing
that privilege, or the mutex having been created earlier by a different account (an interactive
session, or a service running as a different user) with an ACL that excludes it.

## Why it matters

It takes out the whole matrix on any PR whose legs happen to be scheduled there, and it reads as a
red X on the pull request rather than as a machine problem. `main` was green about two hours before
this run, so the machine degraded in between.

## Suggested checks on the host

1. Which account the runner service runs as, and whether it still holds **Create global objects**
   (`SeCreateGlobalPrivilege`) in the local security policy.
2. Whether a stale `Global\DockerRunnerWatchdog.Log` mutex is held by another session — restarting
   the watchdog service (or the host) releases it.
3. Whether the hook should **fail open**. A watchdog that cannot write its own log currently blocks
   every job on the machine; logging the failure and exiting zero would degrade to "no watchdog"
   rather than "no CI". That is a decision for whoever owns the hook, and it is the difference
   between one machine being unmonitored and one machine being unusable.

## Evidence

| Job | Runner | Result |
| --- | --- | --- |
| Unity 6000.3.16f1 editmode | DAD-MACHINE | success |
| Unity 2021.3.45f1 editmode | ELI-MACHINE | failure, Set up runner |
| Unity 2021.3.45f1 playmode | ELI-MACHINE | failure, Set up runner |
| Unity 2022.3.45f1 editmode | ELI-MACHINE | failure, Set up runner |
| Unity 2022.3.45f1 playmode | ELI-MACHINE | failure, Set up runner |
| Unity 6000.3.16f1 playmode | ELI-MACHINE | failure, Set up runner |
| Unity 6000.5.2f1 editmode | ELI-MACHINE | failure, Set up runner |
| Unity 6000.5.2f1 playmode | ELI-MACHINE | failure, Set up runner |

### LINKED ISSUE #425: chore: Significantly optimize githooks
Githooks should be <10s, ideally <5s. Right now they take multiple minutes. We need to heavily optimize them - remove redundancies, parallelize, investigate alternative, faster mechanisms. We need to be able to perform git operations quickly and reliably.

### LINKED ISSUE #468: ci: il2cpp.exe crashes with AccessViolationException in GenericsCollectionParallel, failing the gated standalone leg on an identical tree
### Description

`Unity standalone (gated)` on `6000.5.2f1` failed on `main` at `3ce50e2c`
([run 31963114467](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31963114467)).
It is not our code. `il2cpp.exe` crashed:

```text
C:\Unity\Editors\6000.5.2f1\Editor\Data\il2cpp\build\deploy\il2cpp.exe @Library\Bee\artifacts\rsp\15556249709063864864.rsp
Fatal error. System.AccessViolationException: Attempted to read or write protected memory.
   at Unity.IL2CPP.Common.ReadOnlyHashSet`1[...].CopyTo(System.__Canon[], Int32)
   at System.Collections.Generic.List`1[...]..ctor(System.Collections.Generic.IEnumerable`1<...>)
   at Unity.IL2CPP.GenericsCollection.ReadOnlyInflatedCollectionCollector..ctor(...)
   at Unity.IL2CPP.AssemblyConversion.PrimaryCollection.Steps.Global.GenericsCollectionParallel`1[NeverScanCollectionSettings].C...
   at Unity.IL2CPP.Contexts.Scheduling.PhaseWorkScheduler`1[...].WorkerLoop(System.Object)
```

Everything downstream is the same failure wearing three different masks, which is why the job reports
three failed steps rather than one:

1. `*** Tundra build failed` -> `Player build failed` -> `TestLaunchFailedException`
2. `Editor build produced invalid unity-helpers test player output ... (missing GameAssembly.dll (IL2CPP native code))` (`run-ci-tests.ps1:3898`)
3. `No NUnit results.xml (with a <test-run> element) ... -- tests did not run` and
   `Unity cleanup was not confirmed ... cleanup-reason=unity-return-400006`

### Why this is upstream and nondeterministic, not a regression

**The same tree passed and then failed.** `cba9c775` (session 196 branch head,
[run 31960194754](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31960194754), **success**)
and `3ce50e2c` (its squash-merge on `main`, **failure**) resolve to the byte-identical tree
`565c65175d3c3fac75e08ce698e76275d1530d2f`:

```console
$ git rev-parse cba9c775^{tree} 3ce50e2c^{tree}
565c65175d3c3fac75e08ce698e76275d1530d2f
565c65175d3c3fac75e08ce698e76275d1530d2f
$ git diff --stat cba9c775 3ce50e2c
(empty)
```

The stack is also self-describing: an `AccessViolationException` out of `ReadOnlyHashSet.CopyTo`,
reached from a `List<T>` copy constructor, running inside `GenericsCollectionParallel` on a
`PhaseWorkScheduler` **worker thread**. A hash set faulting mid-copy is what concurrent mutation
during enumeration looks like — a data race in Unity's own generics collector, scheduled differently
on two runs of identical input.

Re-running the failed jobs was enough to clear it.

### Why no retry was added yet

A build-tool `AccessViolationException` is the textbook case for one retry, and
`run-ci-tests.ps1` could wrap the player build. But a retry costs a second IL2CPP build (the
expensive part of the gated leg) on every genuine failure too, and this is the **first** occurrence
recorded with this signature. Filed with the exact signature so a second occurrence is recognized
immediately and the retry is justified by data rather than guessed at.

### What to do on a recurrence

Grep the failed job log for `Unity.IL2CPP.GenericsCollection` and `AccessViolationException`. If it
matches, it is this — re-run failed jobs. If it recurs more than once or twice, add a single retry
around the player build in `run-ci-tests.ps1` and report it to Unity against `6000.5.2f1`.

Found while driving `main` green in session 197.

### LINKED ISSUE #515: ci: mirror DxMessaging's two-tier perf publishing -- our benchmark auto-commit is a direct push to a protected branch
`.github/workflows/unity-benchmarks.yml` refreshes the perf data and then commits it with
`stefanzweifel/git-auto-commit-action` using the built-in `GITHUB_TOKEN`. Its own TODO says what
happens next:

> if the default branch is protected such that the built-in `GITHUB_TOKEN` cannot push (branch
> protection / required reviews), this direct commit will be **REJECTED**. DxMessaging's
> `perf-numbers.yml` hit exactly this and switched to a GitHub App installation token that opens an
> auto-merge PR instead.

**That precondition is now met.** `GET /repos/Ambiguous-Interactive/unity-helpers/branches/main`
reports `"protected": true`. (The classic `/protection` endpoint returns 404 for the token I checked
with, so the exact rule set is unverified -- but the branch is protected, which is the condition the
TODO names.)

The failure mode is the bad kind: the benchmark job does all the expensive work on ELI-MACHINE, the
push is rejected, and **the refreshed numbers are silently discarded**. Nothing is left behind to
show a human that measurements were taken and thrown away. The next run repeats it.

## What DxMessaging does, and what we should mirror

`Ambiguous-Interactive/DxMessaging` `.github/workflows/perf-numbers.yml` publishes in **two tiers**,
so the numbers are never lost:

1. **Direct push** with a GitHub App installation token (`actions/create-github-app-token`). The
   built-in `GITHUB_TOKEN` cannot bypass branch protection because `github-actions[bot]` is a system
   account and is not a selectable bypass actor; an App installation can be.
2. **Fallback auto-merge pull request** when the direct push is rejected with `GH006`. The same
   commit goes to a `ci/perf-auto-update-*` branch and is delivered as a PR opened with the App
   token, with squash auto-merge requested.

> The refreshed numbers are therefore NEVER lost -- worst case they wait in a visible PR for one
> human click.

Supporting pieces worth mirroring:

- **Recursion guard.** App-token pushes *do* re-trigger workflows (unlike `GITHUB_TOKEN` pushes), so
  DxMessaging's `push:` trigger carries `paths-ignore:` for the generated doc and baseline CSV. Our
  benchmark workflow is `schedule` + `workflow_dispatch` only today, so this is not a live problem --
  but it becomes one the moment a `push:` trigger is added, and the guard should land with the App
  token rather than after the first infinite loop.
- **Graceful degradation.** If the App secrets are absent, DxMessaging's commit job degrades to a
  logged no-op instead of failing the run.
- **Freshness guard.** If the branch advances while the long benchmark runs, it warns and skips the
  stale artifacts rather than overwriting newer numbers.
- **A runbook.** `docs/runbooks/perf-numbers-auto-commit.md` records the App setup and the exact
  click-paths for the bypass. We have `docs/runbooks/` but no perf equivalent.

## Why this matters more as of this week

Two things in flight make the publish path load-bearing:

- `docs/performance/random-performance.md` is generated from `[RandomGeneratorMetadata]` by
  `RandomBenchmarkMarkdownBuilder`, and it is **only** rewritten by a benchmark run. #514 had to
  hand-resync nine rows because the ratings changed and the generated table still asserted the old
  ones. If the publish path is silently dropping commits, that table drifts from the code
  indefinitely and every rating correction needs a manual patch.
- Two generators added in #514 (`Xoshiro128StarStar`, `Xoshiro256StarStar`) are in the benchmark
  roster but absent from the published tables until a run lands. That is the gap made concrete.

## Scope

- [ ] Provision the GitHub App (Contents: read/write, Pull requests: read/write) and add
      `AUTO_COMMIT_APP_ID` / `AUTO_COMMIT_APP_PRIVATE_KEY`; allow it to bypass `main` protection.
- [ ] Port the two-tier publish (direct push, then `GH006` fallback auto-merge PR) into
      `unity-benchmarks.yml`, replacing the single `git-auto-commit-action` step and its TODO.
- [ ] Degrade to a logged no-op when the App secrets are absent.
- [ ] Add the freshness guard against a branch that advanced mid-run.
- [ ] Add `paths-ignore` for the generated perf paths, so a future `push:` trigger cannot recurse.
- [ ] Write `docs/runbooks/perf-auto-publish.md` mirroring DxMessaging's runbook.
- [ ] Verify a real run lands the refreshed numbers, since none of this is testable locally.

## Relationship to #431

#431 covers *when* benchmarks run (weekly or release-triggered, on ELI-MACHINE) and the
user-focused tone of the perf docs. This issue is the *delivery mechanism* for the numbers those
runs produce -- the half that is currently a rejected push. They are complementary; #431 makes the
data fresh, this makes it actually land.

### LINKED ISSUE #540: perf: three checks are 93% of the parallel contract suite, so #505's 30 s target needs them, not more cores
## Where the wall clock went

`validate:tests:fast` is a concurrent registry now (#505), and the measurement that unblocked it is
also the measurement that bounds it. Devcontainer, eight cores, `scripts/run-contract-tests.js`:

| run | wall clock | serial sum |
| --- | ---: | ---: |
| `--jobs 1` | **553.0 s** | 552.9 s |
| `--jobs 8` | **144.0 s** | 695.7 s |
| `--jobs 8` again | 152.9 s | 756.4 s |

3.0x-3.8x, and #505 asks for **under 30 s**. Concurrency cannot get there, because the wall clock is
no longer scheduling. It is three checks:

| check | seconds (`--jobs 8`) |
| --- | ---: |
| `validate-lint-error-codes` | **134.5** |
| `shell-portability` | **122.9** |
| `lint-tests` | **104.9** |

The longest one alone is 93% of the whole parallel run. Adding cores changes nothing until these
three change; fixing one only exposes the next, so all three have to move.

## What is actually slow in each

- **`validate-lint-error-codes` (134.5 s).** `scripts/tests/test-validate-lint-error-codes.ps1`
  builds nine fixture repositories and runs the validator in each as a **separate `pwsh` process**,
  and the validator itself shells out to cspell through `scripts/run-node-bin.js`. So the suite pays
  nine pwsh start-ups plus nine cspell start-ups to make nine assertions. Neither start-up is work.
- **`shell-portability` (122.9 s).** Not profiled. It walks every tracked shell script; the question
  is whether the cost is the walk or a per-file subprocess, and that is one `-VerboseOutput` run away.
- **`lint-tests` (104.9 s).** Not profiled. `scripts/lint-tests.ps1` is the interpreted-scan shape
  that #486 already measured on three other linters, where the ceiling was the character loop rather
  than I/O.

## Suggested order

1. Profile `shell-portability` and `lint-tests` before touching them -- #486 and #216 both recorded
   cases where the obvious cause was not the measured one, and #216's own note is that an
   interpreted loop over millions of characters is an honest ceiling, not a bug.
2. `validate-lint-error-codes` is the one whose cause is already known. Nine processes exist because
   the validator is invoked as a script; invoking it in-process, or collapsing the fixtures that do
   not need process isolation, is the obvious first cut.

## Note on the serial sum

The serial sum RISES with concurrency (552.9 s -> 695.7 s -> 756.4 s), so eight workers on eight
cores are contending. That is a reason to be sceptical of any per-check number taken from a `--jobs 8`
run in isolation -- profile a candidate with `--only <id>` on an otherwise idle machine.

### LINKED ISSUE #556: ci: three gates could pass without checking anything, and only one had a way to notice

### LINKED ISSUE #570: ci: seven test assemblies never run in EditMode, so an edit-mode branch of runtime code has no coverage
## What was measured

`UnityEditor.Compilation.CompilationPipeline.GetAssemblies(AssembliesType.Editor)` on `6000.4.6f1`,
through the MCP bridge:

```text
WallstopStudios.UnityHelpers.Tests.Runtime      EditorAssembly=False
WallstopStudios.UnityHelpers.Tests.Core         EditorAssembly=False
WallstopStudios.UnityHelpers.Tests.Editor.Tags  EditorAssembly=True
test assemblies total=33  editorFlagged=26
```

Unity's EditMode runner takes only assemblies flagged `EditorAssembly` — that is, an asmdef with
`"includePlatforms": ["Editor"]`. **Seven of our 33 test assemblies are not**, so their fixtures run
in **playmode only**, on every editor version, regardless of what `.github/unity-versions.json`
contains and regardless of what the editmode leg is handed.

That last part is worth stating explicitly, because the workflow reads as though it covers them.
`scripts/unity/lib/asmdef-discovery.js` resolves `editmode_integrations` to 33 assemblies including
`WallstopStudios.UnityHelpers.Tests.Runtime`, and `unity-tests.yml` passes that whole list to the
editmode legs. Unity then silently drops the seven. The list is a superset; the runner is the filter.

## Why it matters

A `Runtime/` code path that only executes with `Application.isPlaying` **false** has **no CI
coverage at all** unless a `Tests/Editor/**` fixture happens to reach it.

That is not hypothetical — it shipped a bug.
[#569](https://github.com/Ambiguous-Interactive/unity-helpers/issues/569): `Attribute.CurrentValue`
carried

```csharp
#if UNITY_EDITOR
    if (Application.isPlaying)
#endif
    {
        if (_currentValueCalculated) { return _currentValue; }
    }
```

so outside play mode the cache was discarded and a deserialized buff was silently dropped. Two
fixtures asserted the correct behaviour, passed on all four editors for sixteen days, and never once
executed the branch that was wrong. It was found by a harness that runs fixtures by reflection in
EditMode, not by the matrix.

## What is uncovered

`Runtime/` currently has ~20 `Application.isPlaying` sites. Most are lifecycle guards
(`UnityMainThreadDispatcher`, `SceneHelper`, `RuntimeSingleton`) where the edit-mode branch is a
deliberate no-op. The dangerous shape is the one #569 had: a branch that changes a **returned
value**. A sweep for that specific shape — a cache flag beside a `[SerializeField]` — found
`Attribute` and nothing else, so this issue is about the standing gap rather than a known second
instance.

## Possible responses, roughly in order of cost

1. **Document it and rely on placement.** Done as part of #569: the limitation is now recorded in
   `.llm/skills/unity-devcontainer-testing.md`, with the rule that a fixture for an edit-mode branch
   of runtime code belongs in an editor-only assembly. Cheapest, and depends on everyone reading it.
2. **A gate.** A check that flags a `Runtime/**` branch on `Application.isPlaying` whose two sides
   return different values, unless an editor-only test assembly names the member. This is the
   `empty-corpus-gates` shape from
   [#556](https://github.com/Ambiguous-Interactive/unity-helpers/issues/556) — the property is "can
   the uncovered branch be observed at all", and it needs a red half like everything else.
3. **An editor-only companion assembly** that re-runs a chosen subset of `Tests.Runtime`'s
   non-PlayMode fixtures in EditMode. Real coverage, but it doubles the run for those fixtures and
   needs a rule for what belongs in it — most `Tests.Runtime` fixtures are `[UnityTest]` coroutines
   that cannot run in EditMode at all.
4. **Make the discovery honest.** `defaultIncludeAssemblies(target: "editmode")` returning
   assemblies Unity will not run makes the workflow's `EDITMODE_INTEGRATION_ASSEMBLIES` misleading
   to read. Filtering it to editor-only assemblies would not change what runs, but it would stop the
   list implying coverage that does not exist — and it is the change that would have made the gap
   visible.

## Acceptance

1. A decision on which of the above (possibly only 1, which is already done).
2. If a gate: it can go red, with the red half demonstrated per #556.
3. The seven unflagged assemblies are named somewhere durable, so the next reader does not have to
   re-measure them.

### LINKED ISSUE #572: lint: Start-Process -ArgumentList joins without quoting, and the failure looks like a real one
## What happened

Bugbot found this on [PR #571](https://github.com/Ambiguous-Interactive/unity-helpers/pull/571), in a
gate that had just been rewritten *for*
[#556](https://github.com/Ambiguous-Interactive/unity-helpers/issues/556).

`Start-Process -FilePath pwsh -ArgumentList @('-NoProfile', '-File', $path)` concatenates the array
with spaces and **adds no quoting**. A `$path` containing a space is split into separate arguments.

Reproduced with a fixture under `.../dir with space/probe.ps1`:

```text
exit   = 64
stderr = The argument '/tmp/.../scratchpad/dir' is not recognized as the name of a script file.
stdout = Usage: pwsh[.exe] [-Login] [[-File] <filePath> [args]] ...
```

`ProcessStartInfo.ArgumentList` with the same inputs: `exit = 0`, `PROBE RAN OK`.

## Why it deserves a rule rather than three fixes

**The failure is a non-zero exit with a usage banner**, which is indistinguishable from "the child
ran and failed". Any caller asserting "the child failed" is satisfied by it.

Measured, not argued: reverting `test-validate-lint-error-codes.ps1` to `Start-Process` with its
fixtures under a spaced root reddens **six of nine** scenarios — and the other **three stay green**,
because they only assert a non-zero exit or the absence of a crash marker. That is exactly #556's
property ("can this gate go red?"), failing in a gate written to enforce it.

## Current state of the repository

Three call sites, all `Start-Process ... -ArgumentList <array>`:

| site | reachable? | action taken on #571 |
| --- | --- | --- |
| `scripts/tests/test-validate-lint-error-codes.ps1` | yes — temp fixture roots | converted to `ProcessStartInfo.ArgumentList`; fixture root now deliberately contains a space so the regression cannot return quietly |
| `scripts/run-unity-tests.ps1` | yes — `-projectPath`, `-testResults`, `-logFile` are all paths | converted |
| `scripts/unity/bootstrap-windows-runner.ps1` | **no** — every caller passes fixed switches (`/q`, `/install`, `/quiet`, `/norestart`) | left alone, with a comment; `Start-Process -Wait` also waits on installer descendants in a way `Process.WaitForExit()` does not, and it is Windows-only so it cannot be tested from this container |

The rule is documented in `.llm/skills/bash-pwsh-invocation.md`. **It is not enforced.**

## Ask

Add `PWS006` to `scripts/lint-pwsh-invocations.ps1`:

1. **Detect** via the AST, the way `PWS005` already does — a `CommandAst` naming `Start-Process`
   with an `-ArgumentList` (or `-Args`) parameter whose value is an `ArrayLiteralAst`,
   `ArrayExpressionAst` or `VariableExpressionAst`. A regex over the file would repeat #556's
   CSS-gate mistake (matching two things independently); the AST loop is already there.
2. **Message** naming `ProcessStartInfo.ArgumentList` as the fix, and mentioning that
   `Start-Process -Wait` is not `Process.WaitForExit()` so the conversion is not always mechanical.
3. **A per-site opt-out marker**, following the existing `PWS003` marker pattern — the
   `bootstrap-windows-runner.ps1` site needs one and is the test case for it.
4. **A self-test with a red half**, per #556: a fixture using the array form must be reported, and
   one using `ProcessStartInfo` must not.

`PWS` is already registered in `cspell.json`'s root `words`, so `PWS006` needs no dictionary change
(`npm run validate:lint-error-codes` is the contract that would say otherwise).

## Worth considering at the same time

The same "joins without quoting" shape exists for `Invoke-Expression` and for bash `${array[@]}`
expansion without quotes. Both are out of scope here, but if the rule is being written it is worth
deciding explicitly whether `PWS006` is about `Start-Process` specifically or about unquoted
argument joining generally.

wallstop:
**Why:** Unity's EditMode runner takes only assemblies flagged `EditorAssembly`, and discovery was handing the editmode legs seven it would silently drop — which had already cost a whole workflow.

**What:**

- Editmode discovery returns editor-only assemblies; the seven platform-neutral ones are named in a contract test that fails if an eighth appears.
- `unity-benchmarks.yml` is playmode-only: its four EditMode legs ran zero tests and failed `Verify tests actually ran` on every scheduled run.
- Every compiler warning is an error under `Generator~`, which found three assertions that could not fail and one symbol comparison that could be wrong.
- An attribute with only additive modifications recalculates 2.85x faster (0.4563 us -> 0.1600 us), and the effect stacking loop no longer reads a pooled list it has already returned.
- `PWS006` reports `Start-Process -ArgumentList <array>`, which joins without quoting.
- The export smoke gate stages off the bind mount (47 s -> 36 s) and its `rm -rf` guard is driven rather than grepped.
- Docs workflows install through `uv` (5.20 s -> 0.06 s).

Fixes #572
Fixes #570
Fixes #540
Fixes #556
Fixes #425
Fixes #378
Fixes #338
Fixes #515
Fixes #468
Fixes #415

Also advanced with code or measurement: #529, #543, #505, #435, #431. Filed: #575.

---

> [!NOTE]
> **Medium Risk**
> Changes affect gameplay-adjacent runtime (`Attribute`, `EffectHandler`) and CI matrix shape, though behavior is heavily tested; export path guards reduce accidental destructive deletes.
> 
> **Overview**
> **EditMode test discovery and weekly benchmarks** no longer list platform-neutral assemblies Unity silently skips in EditMode. `asmdef-discovery.js` treats EditMode as **editor-only** asmdefs; a new contract test pins the seven playmode-only test assemblies, and **`unity-benchmarks.yml` is playmode-only** with the full Random sweep and timeouts applied on every leg.
> 
> **Runtime behavior:** `Attribute.CalculateCurrentValue` does one addition pass that records whether multiplication/override are needed, skipping empty passes (~2.85× faster for additive-only cases). **`EffectHandler` stack trimming** re-fetches stack handles after each `RemoveEffect` so pooled lists and `OnEffectRemoved` callbacks cannot corrupt the loop.
> 
> **Tooling and CI:** Docs workflows install requirements via **`uv`** after `pip install uv`. **`Generator~/Directory.Build.props`** turns warnings into errors (test analyzer projects opt out). **`PWS006`** lints unsafe `Start-Process -ArgumentList` array usage; **`export-unitypackage.sh`** tightens which `--project-dir` values are allowed before `rm -rf`, with driven refusal tests and staging under temp in portability tests. **`WProtoGenerator`** uses `ReferenceEquals` for backing-field symbol checks (RS1024).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 166eb4532fa13d216214b555623a8c39c339efbb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Runtime/Tags/Attribute.cs
+++ b/Runtime/Tags/Attribute.cs
@@ -134,9 +134,40 @@ namespace WallstopStudios.UnityHelpers.Tags
             float calculatedValue = _baseValue;
             if (0 < _modifications.Count)
             {
-                ApplyModificationsInOrder(ModificationAction.Addition, ref calculatedValue);
-                ApplyModificationsInOrder(ModificationAction.Multiplication, ref calculatedValue);
-                ApplyModificationsInOrder(ModificationAction.Override, ref calculatedValue);
+                // The Addition pass has to visit every modification anyway, so it reports which
+                // other actions are present and the other two passes are skipped when they would
+                // find nothing. Most effects use one action, so this is usually one traversal
+                // rather than three: measured 0.464 us -> 0.171 us for three handles of two
+                // additions, on 6000.4.6f1 (#529).
+                //
+                // The passes stay separate and in this order. Addition, Multiplication and
+                // Override are not interchangeable, and a single pass accumulating them would
+                // also change the ORDER of the float additions, which changes their result.
+                ApplyModificationsInOrder(
+                    ModificationAction.Addition,
+                    ref calculatedValue,
+                    out bool hasMultiplication,
+                    out bool hasOverride
+                );
+                if (hasMultiplication)
+                {
+                    ApplyModificationsInOrder(
+                        ModificationAction.Multiplication,
+                        ref calculatedValue,
+                        out _,
+                        out _
+                    );
+                }
+
+                if (hasOverride)
+                {
+                    ApplyModificationsInOrder(
+                        ModificationAction.Override,
+                        ref calculatedValue,
+                        out _,
+                        out _
+                    );
+                }
             }
 
             _currentValue = calculatedValue;
@@ -283,8 +314,15 @@ namespace WallstopStudios.UnityHelpers.Tags
             _currentValueCalculated = false;
         }
 
-        private void ApplyModificationsInOrder(ModificationAction action, ref float value)
+        private void ApplyModificationsInOrder(
+            ModificationAction action,
+            ref float value,
+            out bool hasMultiplication,
+            out bool hasOverride
+        )
         {
+            hasMultiplication = false;
+            hasOverride = false;
             foreach (
                 KeyValuePair<EffectHandle, List<AttributeModification>> entry in _modifications
             )
@@ -293,9 +331,25 @@ namespace WallstopStudios.UnityHelpers.Tags
                 for (int index = 0; index < modifications.Count; index++)
                 {
                     AttributeModification modification = modifications[index];
-                    if (modification.action == action)
+                    ModificationAction modificationAction = modification.action;
+                    if (modificationAction == action)
                     {
                         ApplyAttributeModification(modification, ref value);
+                        continue;
+                    }
+
+                    switch (modificationAction)
+                    {
+                        case ModificationAction.Multiplication:
+                        {
+                            hasMultiplication = true;
+                            break;
+                        }
+                        case ModificationAction.Override:
+                        {
+                            hasOverride = true;
+                            break;
+                        }
                     }
                 }
             }
--- a/Runtime/Tags/Attribute.cs
+++ b/Runtime/Tags/Attribute.cs
@@ -305,21 +305,6 @@ namespace WallstopStudios.UnityHelpers.Tags
             _currentValueCalculated = false;
         }
 
-        // Returned rather than reported through `out` parameters: these accumulate across the
-        // whole traversal, and an `out` assigned anywhere but immediately before the return is the
-        // shape that lets a later path forget to write it.
-        private readonly struct RemainingActions
-        {
-            public readonly bool hasMultiplication;
-            public readonly bool hasOverride;
-
-            public RemainingActions(bool hasMultiplication, bool hasOverride)
-            {
-                this.hasMultiplication = hasMultiplication;
-                this.hasOverride = hasOverride;
-            }
-        }
-
         private RemainingActions ApplyModificationsInOrder(
             ModificationAction action,
             ref float value
@@ -581,6 +566,21 @@ namespace WallstopStudios.UnityHelpers.Tags
         public override string ToString()
         {
             return ((float)this).ToString(CultureInfo.InvariantCulture);
+        }
+
+        // Returned rather than reported through `out` parameters: these accumulate across the
+        // whole traversal, and an `out` assigned anywhere but immediately before the return is the
+        // shape that lets a later path forget to write it.
+        private readonly struct RemainingActions
+        {
+            public readonly bool hasMultiplication;
+            public readonly bool hasOverride;
+
+            public RemainingActions(bool hasMultiplication, bool hasOverride)
+            {
+                this.hasMultiplication = hasMultiplication;
+                this.hasOverride = hasOverride;
+            }
         }
     }
 }
--- a/Runtime/Tags/Attribute.cs
+++ b/Runtime/Tags/Attribute.cs
@@ -143,30 +143,21 @@ namespace WallstopStudios.UnityHelpers.Tags
                 // The passes stay separate and in this order. Addition, Multiplication and
                 // Override are not interchangeable, and a single pass accumulating them would
                 // also change the ORDER of the float additions, which changes their result.
-                ApplyModificationsInOrder(
+                RemainingActions remaining = ApplyModificationsInOrder(
                     ModificationAction.Addition,
-                    ref calculatedValue,
-                    out bool hasMultiplication,
-                    out bool hasOverride
+                    ref calculatedValue
                 );
-                if (hasMultiplication)
+                if (remaining.hasMultiplication)
                 {
-                    ApplyModificationsInOrder(
+                    _ = ApplyModificationsInOrder(
                         ModificationAction.Multiplication,
-                        ref calculatedValue,
-                        out _,
-                        out _
+                        ref calculatedValue
                     );
                 }
 
-                if (hasOverride)
+                if (remaining.hasOverride)
                 {
-                    ApplyModificationsInOrder(
-                        ModificationAction.Override,
-                        ref calculatedValue,
-                        out _,
-                        out _
-                    );
+                    _ = ApplyModificationsInOrder(ModificationAction.Override, ref calculatedValue);
                 }
             }
 
@@ -314,15 +305,28 @@ namespace WallstopStudios.UnityHelpers.Tags
             _currentValueCalculated = false;
         }
 
-        private void ApplyModificationsInOrder(
+        // Returned rather than reported through `out` parameters: these accumulate across the
+        // whole traversal, and an `out` assigned anywhere but immediately before the return is the
+        // shape that lets a later path forget to write it.
+        private readonly struct RemainingActions
+        {
+            public readonly bool hasMultiplication;
+            public readonly bool hasOverride;
+
+            public RemainingActions(bool hasMultiplication, bool hasOverride)
+            {
+                this.hasMultiplication = hasMultiplication;
+                this.hasOverride = hasOverride;
+            }
+        }
+
+        private RemainingActions ApplyModificationsInOrder(
             ModificationAction action,
-            ref float value,
-            out bool hasMultiplication,
-            out bool hasOverride
+            ref float value
         )
         {
-            hasMultiplication = false;
-            hasOverride = false;
+            bool hasMultiplication = false;
+            bool hasOverride = false;
             foreach (
                 KeyValuePair<EffectHandle, List<AttributeModification>> entry in _modifications
             )
@@ -353,6 +357,8 @@ namespace WallstopStudios.UnityHelpers.Tags
                     }
                 }
             }
+
+            return new RemainingActions(hasMultiplication, hasOverride);
         }
 
         private static void ValidateInput(float value, [CallerMemberName] string caller = null)
--- a/scripts/unity/export-unitypackage.sh
+++ b/scripts/unity/export-unitypackage.sh
@@ -51,16 +51,26 @@ fi
 # --stage-only run (#540).
 ARTIFACTS_ROOT="$(realpath -m "${REPO_ROOT}/.artifacts")"
 TEMP_ROOT="$(realpath -m "${TMPDIR:-/tmp}")"
+RESOLVED_REPO_ROOT="$(realpath -m "${REPO_ROOT}")"
 PROJECT_DIR="$(realpath -m "${PROJECT_DIR}")"
+
+# .artifacts is unconditionally fine -- it IS the repository's scratch directory. The temp root is
+# fine only outside the checkout, because a checkout can live under it: a CI runner working
+# directory, or a container whose workspace is a temp mount. Without the second test, --project-dir
+# pointing at the repository root would be accepted there and then deleted (Bugbot, PR #574).
 PROJECT_DIR_ALLOWED=0
-for allowed_root in "${ARTIFACTS_ROOT}" "${TEMP_ROOT}"; do
-    if [[ "${PROJECT_DIR}" != "${allowed_root}" && "${PROJECT_DIR}" == "${allowed_root}/"* ]]; then
+if [[ "${PROJECT_DIR}" != "${ARTIFACTS_ROOT}" && "${PROJECT_DIR}" == "${ARTIFACTS_ROOT}/"* ]]; then
+    PROJECT_DIR_ALLOWED=1
+elif [[ "${PROJECT_DIR}" != "${TEMP_ROOT}" && "${PROJECT_DIR}" == "${TEMP_ROOT}/"* ]]; then
+    # Neither the checkout, nor anything inside it, nor anything containing it.
+    if [[ "${PROJECT_DIR}" != "${RESOLVED_REPO_ROOT}" &&
+          "${PROJECT_DIR}" != "${RESOLVED_REPO_ROOT}/"* &&
+          "${RESOLVED_REPO_ROOT}" != "${PROJECT_DIR}/"* ]]; then
         PROJECT_DIR_ALLOWED=1
-        break
     fi
-done
+fi
 if (( PROJECT_DIR_ALLOWED == 0 )); then
-    echo "ERROR: Refusing to create the export project unless it is a subdirectory under ${ARTIFACTS_ROOT} or ${TEMP_ROOT}: ${PROJECT_DIR}" >&2
+    echo "ERROR: Refusing to create the export project unless it is a subdirectory under ${ARTIFACTS_ROOT}, or under ${TEMP_ROOT} and outside ${RESOLVED_REPO_ROOT}: ${PROJECT_DIR}" >&2
     exit 1
 fi
 
--- a/scripts/unity/export-unitypackage.sh
+++ b/scripts/unity/export-unitypackage.sh
@@ -41,10 +41,26 @@ if [[ -z "${OUTPUT_PATH}" ]]; then
     OUTPUT_PATH="${REPO_ROOT}/.artifacts/release/${PACKAGE_NAME}-${PACKAGE_VERSION}.unitypackage"
 fi
 
+# The project directory is deleted with `rm -rf` below, so it has to be somewhere a mistake is
+# survivable. A strict subdirectory of .artifacts or of the system temp directory qualifies; the
+# roots themselves do not, and neither does anything else.
+#
+# The temp root is not a convenience. Staging copies ~1,900 files and 82 MB, and on a bind-mounted
+# workspace that is the whole cost of the export smoke gate: measured 4.25 s to copy Runtime and
+# Editor into .artifacts against 0.022 s into /tmp, and 15.0 s against ~1 s for the whole
+# --stage-only run (#540).
 ARTIFACTS_ROOT="$(realpath -m "${REPO_ROOT}/.artifacts")"
+TEMP_ROOT="$(realpath -m "${TMPDIR:-/tmp}")"
 PROJECT_DIR="$(realpath -m "${PROJECT_DIR}")"
-if [[ "${PROJECT_DIR}" == "${ARTIFACTS_ROOT}" || "${PROJECT_DIR}" != "${ARTIFACTS_ROOT}/"* ]]; then
-    echo "ERROR: Refusing to create the export project unless it is a subdirectory under ${ARTIFACTS_ROOT}: ${PROJECT_DIR}" >&2
+PROJECT_DIR_ALLOWED=0
+for allowed_root in "${ARTIFACTS_ROOT}" "${TEMP_ROOT}"; do
+    if [[ "${PROJECT_DIR}" != "${allowed_root}" && "${PROJECT_DIR}" == "${allowed_root}/"* ]]; then
+        PROJECT_DIR_ALLOWED=1
+        break
+    fi
+done
+if (( PROJECT_DIR_ALLOWED == 0 )); then
+    echo "ERROR: Refusing to create the export project unless it is a subdirectory under ${ARTIFACTS_ROOT} or ${TEMP_ROOT}: ${PROJECT_DIR}" >&2
     exit 1
 fi
 
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -44,8 +44,16 @@ jobs:
         id: pages
         uses: actions/configure-pages@45bfe0192ca1faeb007ade9deae92b16b8254a0d
 
+      # uv resolves and installs by hard-linking from its cache instead of unpacking wheels.
+      # Measured on this repository's requirements-docs.txt into a fresh environment, both with a
+      # warm cache: pip 5.20 s best of three, uv 0.06 s -- and this step is 11 s of a 31 s job
+      # (#338). pip installs uv itself in ~2 s, which keeps setup-python's cache doing the
+      # downloading and adds no new action to pin.
       - name: Install dependencies
-        run: pip install -r requirements-docs.txt
+        run: |
+          set -euo pipefail
+          pip install uv==0.12.6
+          uv pip install --system -r requirements-docs.txt
 
       - name: Build with MkDocs
         run: mkdocs build --strict
--- a/.github/workflows/unity-benchmarks.yml
+++ b/.github/workflows/unity-benchmarks.yml
@@ -70,30 +70,32 @@ jobs:
           const performanceAssembly = "WallstopStudios.UnityHelpers.Tests.Runtime.Performance";
           const randomAssembly = "WallstopStudios.UnityHelpers.Tests.Runtime.Random";
 
+          // PlayMode only. Both benchmark assemblies are platform-neutral, and Unity's EditMode
+          // runner takes only assemblies flagged EditorAssembly -- so the four editmode legs
+          // were handed two assembly names, ran zero tests, and failed "Verify tests actually
+          // ran" on every scheduled run. Measured: the 2021.3.45f1 editmode leg's log mentions
+          // Tests.Runtime.Random three times (the echoed list) against 2,375 in playmode
+          // (#570). The thorough Random sweep the editmode leg was configured for moves here,
+          // which is the only mode that can execute it.
           function benchmarkProfile(target) {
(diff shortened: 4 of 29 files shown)


Runtime/Tags/Attribute.cs
```diff
@@ -283,8 +305,28 @@ public void ClearCache()
             _currentValueCalculated = false;
         }
 
-        private void ApplyModificationsInOrder(ModificationAction action, ref float value)
+        // Returned rather than reported through `out` parameters: these accumulate across the
+        // whole traversal, and an `out` assigned anywhere but immediately before the return is the
+        // shape that lets a later path forget to write it.
+        private readonly struct RemainingActions
         {
+            public readonly bool hasMultiplication;
```
wallstop: Please place class/struct definitions at the end of the file, or in their own file, do not intersperse with code.

scripts/unity/export-unitypackage.sh
```diff
@@ -41,10 +41,26 @@ if [[ -z "${OUTPUT_PATH}" ]]; then
...
 PROJECT_DIR="$(realpath -m "${PROJECT_DIR}")"
-if [[ "${PROJECT_DIR}" == "${ARTIFACTS_ROOT}" || "${PROJECT_DIR}" != "${ARTIFACTS_ROOT}/"* ]]; then
-    echo "ERROR: Refusing to create the export project unless it is a subdirectory under ${ARTIFACTS_ROOT}: ${PROJECT_DIR}" >&2
+PROJECT_DIR_ALLOWED=0
+for allowed_root in "${ARTIFACTS_ROOT}" "${TEMP_ROOT}"; do
+    if [[ "${PROJECT_DIR}" != "${allowed_root}" && "${PROJECT_DIR}" == "${allowed_root}/"* ]]; then
+        PROJECT_DIR_ALLOWED=1
+        break
+    fi
+done
```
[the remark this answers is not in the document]
wallstop: Confirmed and fixed in 166eb453 — and the reachable case is worse than the description: a checkout
can *live* under the temp root (a CI runner working directory, a container whose workspace is a temp
mount), so `--project-dir <repo>` would have been accepted there and then `rm -rf`'d.

The temp branch now also refuses the checkout, anything inside it, and anything containing it:

```bash
elif [[ "${PROJECT_DIR}" != "${TEMP_ROOT}" && "${PROJECT_DIR}" == "${TEMP_ROOT}/"* ]]; then
    if [[ "${PROJECT_DIR}" != "${RESOLVED_REPO_ROOT}" &&
          "${PROJECT_DIR}" != "${RESOLVED_REPO_ROOT}/"* &&
          "${RESOLVED_REPO_ROOT}" != "${PROJECT_DIR}/"* ]]; then
```

You were also right that the new contract test could not have caught it — it passed the *real*
repository root, which is not under the temp root in this container, so the assertion held for the
wrong reason. The test now drives the script from a copy of itself inside a **fake checkout under
the temp root**, which is the layout the finding is about, and asserts that the fake repo root, a
directory inside it, and its parent are all refused. Removing the new condition reddens it.

Runtime/Tags/Attribute.cs
wallstop: Moved to the end of the class in 166eb453.

Recorded the rule so it does not come back: `context.md` rule 6 and `create-csharp-file` rule 5b.

Then swept for the same shape — a nested type declaration followed by more members in the same type
body, across `Runtime/`, `Editor/` and `Tests/`: **249 sites in 178 files**. Two verified by hand as
true positives (`WGroupAttribute.cs:70`, `SerializableList.cs:72`), so the count is not a heuristic
artefact. Filed as [#575](https://github.com/Ambiguous-Interactive/unity-helpers/issues/575) with
the inventory, the worst offenders, and the ratchet-baseline shape the linter should take (the same
one `test-out-parameter-discipline.js` uses, so new violations fail immediately while the backlog is
worked down).

Left it out of this PR because 178 mechanical file moves would bury the rest of the diff — say the
word if you would rather have the sweep here.

Issue #578, opened 2026-08-26
[Feature]: an int-keyed open-addressing map, measured at 1.26x-2.19x over Dictionary


wallstop:
### What the measurement found

[#309](https://github.com/Ambiguous-Interactive/unity-helpers/issues/309) asked whether the Java
hash-map benchmarks transfer to Unity and whether a custom table is worth building. The research is
done and the answer is yes, for one specific shape: **an int-keyed, lookup-dominated map.**

Measured on Unity **6000.4.6f1**, Editor Mono, counterbalanced `ABBABAAB` with a settled heap per
slot, geometric-mean ratio over four adjacent pairs. `TryGetValue`/`TryGet` on both sides, so the
API shapes match. The probe map starts at capacity 16 and grows, and a tenth of its entries are
removed and re-added first so it carries real tombstones rather than a pristine table:

| entries | miss rate | open addressing vs `Dictionary<int,int>` | spread (subject / reference) |
| ---: | ---: | ---: | --- |
| 1,000 | 0% | **1.97x** | 4% / 2% |
| 1,000 | 50% | **1.26x** | 8% / 5% |
| 10,000 | 0% | **2.19x** | 2% / 2% |
| 10,000 | 50% | **1.65x** | 3% / 7% |

It never loses, and it degrades exactly where linear probing is expected to: a miss must walk to an
empty slot, so the margin halves at a 50% miss rate.

Two adjacent findings from the same session, both on #309:

- **A dense key range should not use a hash table at all.** An `int[]` indexed by the key is ~5x a
  `Dictionary` for an 8-member enum.
- **Do not hand-write an `IEqualityComparer<TEnum>`** -- it is **11% slower** than the default,
  because Unity's Mono already ships `EnumEqualityComparer<T>` and `Dictionary` fast-paths the
  default comparer while a supplied one is an interface call it cannot skip. Filed as
  [#577](https://github.com/Ambiguous-Interactive/unity-helpers/issues/577).

### What to build

A read-mostly, value-key map. The measured shape is the specification:

- `int` (and probably `long`) keys, open addressing with linear probing, power-of-two capacity, a
  single-multiply mixer, `key + 1` stored so `0` reads as empty and `-1` as a tombstone.
- `TryGet`, `Set`, `Remove`, `Count`, `Clear`, enumeration, and growth at a 0.5 load factor.
- No `IEqualityComparer` indirection: that indirection is a real part of what is being beaten.

### Acceptance criteria

[...]

Issue #399, opened 2026-08-10
WallstopProto: Arbitrary-Dimension Arrays, Jagged Arrays, all stdlib types, all wallstop + Ambiguous-Interactive Types


wallstop:
We should support:
- Arbitrary dimension Arrays
- Collections like List<int[][]>, List<List<int[][]>[][]>, etc
- int[3][100]
- Every serializable stdlib type
- All possible proto-types in projects in the github user wallstop (me) projects, especially UniverseTowerDefense
- Similar all proto types in ambiguous-interactive repos
- research into public projects that are Unity, use protobuf-net, and their use cases - do we support the majority, if not all of them? Can we?

wallstop:
## Measured: protobuf-net supports none of the nested shapes

Before designing anything here, I checked what the oracle does. protobuf-net 3.2.56 **refuses all
four** nested forms outright:

```
int[][]          => NotSupportedException: Nested or jagged lists, arrays and maps are not supported
int[,]           => NotSupportedException: Repeated data of type System.Int32[,] is not supported
List<int[]>      => NotSupportedException: Nested or jagged lists, arrays and maps are not supported
List<List<int>>  => NotSupportedException: Nested or jagged lists, arrays and maps are not supported
```

This is not a protobuf-net limitation so much as a protobuf one: **`repeated repeated` is not
expressible in the wire format.** A repeated field is a field number repeated N times; there is no
bracket to say where one inner list ends and the next begins. The standard protobuf idiom is a
**wrapper message per inner list** — `repeated Inner rows` where `message Inner { repeated int32
values = 1; }`.

### Why that reframes this issue

WallstopProto currently refuses these with `WPROTO003`, which **exactly matches protobuf-net**. So
supporting them is not a compatibility fix, it is new capability — and the usual constraint does not
bind:

> the requirement is that protobuf-net can read what we write

**For these shapes there is nothing on the other side to interoperate with.** protobuf-net cannot
produce such a payload, so no existing data has this form and no protobuf-net reader will ever be
handed one. We are free to choose the encoding, and the obvious choice is the wrapper-message idiom
above, because it is what every other protobuf implementation would emit for the equivalent schema
and what `protoc` generates.

Two consequences worth deciding deliberately:

1. **A wrapper message costs bytes** — a key and a length prefix per inner list. For `int[][]` that
   is unavoidable; it is the price of the shape being expressible at all.
2. **Round-tripping through protobuf-net becomes impossible for these members**, by construction.
[...]

PR #581, opened 2026-08-26
RuntimeSingleton safety, WProto shapes, IntMap, JSON allocs

### LINKED ISSUE #504: [Bug]: perf: Optimize JSON/Protobuf Ser/Deser
### Description

It looks like there are some opportunities to utilize buffers, pooling, and other techniques, like in our custom Json ReadArray (double allocates) as well as others.

### Steps to Reproduce

Scan through impl

### Expected Behavior

Low alloc + fast

### Actual Behavior

More allocs than necessary

### Severity

Critical (crash, data loss)

### Unity Version

Unity 6.5 (6000.5)

### Package Version

3.5.1

### Platform

- [ ] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Regression

### Additional Context

_No response_

### LINKED ISSUE #577: docs: a hand-written enum comparer is 11% slower than the default on Unity's Mono
## The advice

Every Unity performance article says the same thing about enum-keyed dictionaries: `Dictionary<TEnum, TValue>` boxes its keys, so supply your own `IEqualityComparer<TEnum>` that casts to the underlying integer.

## It is false on the runtime this package ships against, and following it costs 11%

Measured on **6000.4.6f1**, Editor Mono, counterbalanced `ABBABAAB` with a settled heap per slot and
the raw per-cycle spread retained (the protocol added for
[#573](https://github.com/Ambiguous-Interactive/unity-helpers/issues/573)). Eight-member enum,
hit-only lookups, 4M per slot:

| | Mops/s | ratio | spread |
| --- | ---: | ---: | ---: |
| `Dictionary<TEnum,int>`, default comparer | 168.7 | -- | 3.1% |
| the same, hand-written struct comparer | 150.2 | **0.8902x** | 1.3% |

Two independent runs agree: 0.8940x, then 0.8902x.

The cause is one line:

```text
EqualityComparer<ProbeKey>.Default.GetType().Name  ->  EnumEqualityComparer`1
```

Unity's Mono BCL already ships a specialized enum comparer, so there is no boxing to remove.
`Dictionary` also fast-paths the default comparer, while a supplied one is an interface call it
cannot skip -- which is where the 11% goes.

## What to do

This package's documentation is where a Unity developer looks for this kind of thing, and it
currently says nothing. Proposed, in rough order of value:

1. A short section in the performance docs stating the measurement and the conclusion: **do not
   supply a comparer for an enum key**. One paragraph, one table, with the runtime and editor
   version named, because this is exactly the kind of claim that goes stale.
2. Re-measure under **IL2CPP** before publishing. All three numbers above are Editor Mono. If AOT
   behaves differently the doc has to say which is which, and that is the version of this note worth
   having.
3. Sweep this repository for the same pattern -- a hand-written comparer supplied for a key type
   whose default is already specialized. `rg 'IEqualityComparer' Runtime/` currently returns
   `Serializer`, `ProtoEqualityExtensions`, `IReadonlyListExtensions`, `DictionaryExtensions`,
   `ColorQuantization`, `SpatialHash3D`, `SpatialHash2D`, `SparseSet` and `DisjointSet`; none is an
   enum comparer today, so this is a guard against adding one rather than a cleanup.

## Where the measurement came from

[#309](https://github.com/Ambiguous-Interactive/unity-helpers/issues/309), which also measured the
two shapes that DO win: a dense `int[]` indexed by the enum is roughly **5x** a `Dictionary`, and
linear-probing open addressing is roughly **2x** on 1,000 sparse int keys. Those are structures to
build; this one is a sentence to write, which is why it is filed separately.

### LINKED ISSUE #582: RuntimeSingleton: entering Play Mode destroys scene-authored instances
`RuntimeSingletonRegistry` clears its instances from a
`[RuntimeInitializeOnLoadMethod(BeforeSceneLoad)]`, and `RuntimeSingleton<T>.ClearInstance` destroys
**every live instance's GameObject**, not just the one it cached:

```csharp
T[] liveInstances = UnityObjectExtensions.FindObjectsOfTypeShim<T>(true);
for (int i = 0; i < liveInstances.Length; i++)
{
    liveInstances[i].StopAllCoroutines();
    liveInstances[i].gameObject.Destroy();
}
```

In a player that is harmless: nothing is alive when `BeforeSceneLoad` runs. **In the Editor, pressing
Play on a scene that already carries a `RuntimeSingleton` is not harmless** — the open scene's
objects are alive at that moment, so the sweep destroys objects it did not create, and the scene runs
without the component it authored.

## Measured

Unity 6000.5.2f1, `m_EnterPlayModeOptionsEnabled: 1`, `m_EnterPlayModeOptions: 0`.

A level scene with eleven root GameObjects, two of which carry a `RuntimeSingleton` subclass. In Play
mode **exactly those two roots are absent and the other nine are present**. Nothing else in the scene
selects on "is a RuntimeSingleton", and `FindObjectsByType<T>(FindObjectsInactive.Include)` returns
zero for both types for the whole session.

The consequence is not an absence, because `Instance` fabricates one:

| moment | the `LevelBounds` the game sees |
| --- | --- |
| first frame | none |
| after any consumer asks `Instance` | `(x:0, y:0, width:0, height:0)` |

The authored value is `(0, 0, 76.8, 43.2)`. The consumer downstream clamps a camera to the centre of
that rectangle, so the camera left the level and held on the world origin — 62.56 world units in one
move, in a take where nothing else moved.

## Why this is worth fixing rather than documenting

Every symptom is silent. `Instance` cannot fail, the fabricated object answers every call, and the
authored values it replaces are defaults that look plausible. A consumer written the obvious way has
no way to tell which one it is holding.

## Suggested shapes

- Have `ClearInstance` drop the cached reference and destroy only instances the getter created —
  track those, rather than sweeping by type.
- Or skip the sweep when `Application.isPlaying` is already true at `BeforeSceneLoad`, which is the
  Editor-only state that makes open-scene objects visible to it.

The clear-the-cache half is what the callback needs; the destroy half is what a test fixture wants,
and the two look worth separating.

Found from Ambiguous-Interactive/IshoBoy#694, where it put both marketing captures of a level on the
world origin for most of their take with every gate still green.

wallstop:
**Why:** A pass over the open-issue backlog: preserve scene-authored runtime singleton state, finish the measured WallstopProto standard-library sweep, add the int-keyed map the benchmarks asked for, and fix the named JSON allocation class.

**What:**

- Reset `RuntimeSingleton` caches before scene load without destroying live scene-authored components; retain explicit destructive teardown for test isolation (fixes #582)
- Keep build-lock pins at the last known-good v1.12.1 while runner-capacity quarantine remains tracked in #584
- Recover warm Unity projects when package source paths are added or removed by clearing only compilation outputs; commit the inventory marker only after passing NUnit output
- Serve `char` and `Uri` in WallstopProto against both protobuf-net oracles; refuse `IntPtr`/`UIntPtr`/`Type` with recorded reasons (finishes the #399 stdlib-shape sweep)
- Add `IntMap<TValue>`, an int-keyed open-addressing map with raw-int comparison, oracle-tested (part of #578)
- Stop JSON array reads allocating twice; property names compare without allocating (fixes #504, #543, #505 measurements)
- Document the 11% enum-comparer cost on Unity's Mono with runtime and protocol named (closes #577)

Fixes #504, fixes #577, fixes #582

---

> [!NOTE]
> **Medium Risk**
> Play-mode singleton behavior and JSON/protobuf deserialization paths change in production-facing code; CI cache invalidation logic affects build reliability but not game runtime.
> 
> **Overview**
> This PR bundles backlog fixes across singleton lifecycle, protobuf BCL types, JSON deserialization, a new int map, and Unity CI warm-cache rules.
> 
> **Runtime singletons** no longer destroy scene-authored `RuntimeSingleton` components when entering Play Mode: `BeforeSceneLoad` only **resets static caches** via `ResetAllRegisteredCaches`, while `ClearInstance` / `ClearAllRegisteredInstances` still tear down live objects for tests.
> 
> **WallstopProto** adds **`char`** (varint code unit, repeated elements never packed) and **`Uri`** (UTF-8 of `OriginalString`, strict decode). Generator `Shape` gains `NeverPacked`; runtime mirrors that with `IWProtoNeverPacked`. **`IntPtr`**, **`UIntPtr`**, and **`Type`** stay **WPROTO003** refusals with documented reasons. Differential tests and docs are updated.
> 
> **JSON** array reads use **`JsonArrayAccumulator<T>`** with pooled growth so steady state avoids list backing-array churn plus `ToArray`; dictionary/set converters match property names with **`ValueTextEquals`** instead of allocating strings.
> 
> **`IntMap<TValue>`** is a new open-addressing int-key map (raw int keys, power-of-two table, tombstones) aimed at hit-heavy lookups vs `Dictionary<int,int>`.
> 
> **Unity CI** invalidates Bee/ScriptAssemblies caches when the **repo root** or **package source path inventory** changes (content-only edits keep the warm cache); the source inventory marker is written only after **passing NUnit** results.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40379b42e2553183d8dd3ec13287d39f0c2d277f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/DataStructure/IntMap.cs
```diff
@@ -0,0 +1,476 @@
...
+    /// </para>
+    /// <para>
+    /// The table stores keys verbatim beside their values, and the two lowest key values name the
+    /// slot states instead of callers' data; see <see cref="MinimumAllowedKey"/>. A workload whose
+    /// misses outnumber its hits should stay on <see cref="Dictionary{TKey, TValue}"/>.
+    /// </para>
+    /// </remarks>
+    public sealed class IntMap<TValue>
+        : IEnumerable<KeyValuePair<int, TValue>>,
+            IReadOnlyCollection<KeyValuePair<int, TValue>>
```
wallstop: Should this implement IDictionary<int, TValue> / IReadonlyDictionary<int, TValue>?

Runtime/Core/DataStructure/IntMap.cs
```diff
@@ -0,0 +1,476 @@
...
+                }
+
+                return false;
+            }
+
+            /// <inheritdoc />
+            public void Reset()
+            {
+                _slot = 0;
+                _current = default(KeyValuePair<int, TValue>);
```
wallstop, line 532: Do we care about resetting this value here?

wallstop: Yes. `Reset()` now restores `_current` to its default as well as rewinding the slot, matching the pair, key, and value enumerators' before-first state.

Runtime/Core/DataStructure/IntMap.cs
wallstop: Implemented `IReadOnlyDictionary<int, TValue>` so callers get the standard read contract without adding the mutating `IDictionary` surface. The concrete `TryGet`/`TrySet` API remains the zero-indirection hot path.

scripts/tests/test-unity-workflow-matrix-contract.ps1
```diff
@@ -1284,6 +1305,51 @@ function Test-UnityCompilationCacheBehavior {
...
+        Clear-StaleUnityCompilationCache -Project $matchingMarkerFixture.Project -RepoRoot $matchingMarkerRoot
+        if (-not (Test-CompilationCacheSentinelsPresent -Fixture $matchingMarkerFixture)) {
+            return 'content-only source changes must preserve compilation caches to retain normal warm-build performance'
+        }
+
+        # A newly added source path is different: Unity 2022 can begin Bee compilation before a
+        # warm persistent project's AssetDatabase includes the new package file, leaving sibling
+        # sources to compile against a missing type. Invalidate only compilation outputs so the
+        # source graph is rebuilt while PackageCache remains warm.
+        Set-Content -LiteralPath (Join-Path $matchingMarkerRoot 'Runtime\Added.cs') -Value 'class Added {}' -Encoding utf8
```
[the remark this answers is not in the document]
wallstop: Fixed in cc2186e8. The fixture now creates `Runtime` once and joins `Existing.cs` / `Added.cs` as separate path segments, so the source inventory is exercised on Windows, Linux, and macOS. The focused contract passed locally under Linux after the change.

Issue #478, opened 2026-08-17
[Docs]: un-llm-ify the documentation


wallstop:
### Page URL or Section

https://github.com/Ambiguous-Interactive/unity-helpers

### Description of Issue

The original docs were mostly llm-written. We need to seriously cut all of the llm stuff.

Documentation should be:
- Short
- User-focused
- Simple
- To-the-point
- Zero fluff

We should have zero em-dashes, or even "--", common llm-isms, emojis, etc.

Documentation should primarily showcase features of the toolkit, why the exist, and how to use them correctly, so that users fall into a pit of success.

### Suggested Fix

Complete repo sweep.

wallstop:
The agent-facing half of this shipped in #532, for pull request text specifically.

**The data.** The rule already said "under about 60 characters". Of the twelve titles before this
change, **nine were over 60** — median 85, longest 136:

```text
136  Hand an immutable contract the seed its constructor built, pin the NUnit Unity ships, and warn when ...
132  Construct a JSON converter where its closure lives, declare a hook where every reader runs it, and ...
112  Take large struct receivers by in, give UI Toolkit its extensions, and make a passing PRNG rating ...
106  Make a grid cell cost its distance from the origin, and stop one refused surrogate taking the rest ...
 51  perf: optimize serialization and random restoration
```

A soft rule that loses nine times in twelve is not a rule. It is now mechanical:

- **50 characters, hard**, with a `printf '%s' "$TITLE" | wc -c` check to run before sending.
- A copy-paste body template: one `**Why:**` sentence, two to five one-line `**What:**` bullets,
  `Fixes #123`, nothing else.
- A limits table, and an explicit never-include list — root causes, measurements, run IDs, session
  numbers, CI results, validation lists, byte traces, the diff, the file list.
- Short sentences, active voice, common words. A number only when the number is the point.

The guidance itself was rewritten in the style it asks for, and is about half its former length.
This PR's own title is 39 characters.

**This does not close the issue.** The ask here is the user-facing documentation under `docs/`, which
is untouched. Leaving open for that.

wallstop:
## The backlog, counted

Measured over `docs/**/*.md` plus `README.md`:

| thing | count | files |
| --- | ---: | ---: |
| em-dash `U+2014` | **613** | 61 |
| ` -- ` | 30 | 6 |
| emoji | -- | 10+ |

Worst files: `docs/features/serialization/serialization.md` (97), `README.md` (77),
[...]

Issue #462, opened 2026-08-16
[Feature]: mutation tests for serialization tech


wallstop:
### Feature Description

Once WallstopProto and our JSON serializers are solidified, we should throw as much mutation/fuzz/randomized testing as possible at them to ensure correctness, doing research into modern best practices, techniques, and approaches for testing serialization technologies.

### Use Case / Motivation

Ensure robustness and correctness.

### Unity Version

None

### Package Version

None

wallstop:
Session 207 update — PR #507 adds deterministic legal-equivalence JSON metamorphic coverage: https://github.com/Ambiguous-Interactive/unity-helpers/pull/507

The repository already had substantial fixed-seed malformed/random coverage: protobuf byte mutation and hostile wire claims, public-facade random bytes, JSON property/node mutation and truncation, accepted-payload canonical fixed points, and protobuf-net differential random valid values. Repeating “malformed input fails safely” would not add the requested technique.

The new converter test transforms every readable current/legacy seed by:
- recursively reversing unique object-property order;
- Unicode-escaping every property-name UTF-16 code unit;
- adding insignificant whitespace;
- preserving array order and scalar spellings;
- independently proving the parsed structure is exactly reversed;
- independently scanning raw UTF-8 property-name tokens to prove they contain only \uNNNN units; and
- requiring decode/re-encode canonical equality with the unmodified seed.

The transformer conservatively rejects case-insensitive duplicate property names, and coverage counters prevent a no-op transform from passing.

Remaining protobuf metamorphic work is intentionally not faked. It needs a schema-aware tokenizer, known-tag exclusion for unknown-field injection, repeated/map occurrence-order preservation, packed-run splitting that preserves element order, include/oneof constraints, and an explicit protobuf-net v2/v3 known-divergence policy. PR #507 advances #462 without claiming that broader battery complete.

wallstop:
CI found and sharpened the #462 oracle after the initial PR head: WGuid's `_low/_high` alternate is a well-formed but deliberately invalid version-4 probe, not an accepted legacy value.

The follow-up commit `c355111e` preserves both metamorphic contracts:
- accepted seeds must preserve canonical value under reversal/Unicode escapes/whitespace;
- rejected seeds must preserve `JsonException` rejection under the same equivalent spelling.

[...]

Issue #441, opened 2026-08-14
[Docs]: Improve Doc Simplicity, Examples, readability


wallstop:
### Page URL or Section

https://unity-helpers.ambiguous-interactive.github.io

### Description of Issue

Lots of docs are very verbose, with poor examples, and more "wallstop the developer" focused instead of "user of Unity Helpers" focused. Every single example and documentation area should be extremely user-friendly, with short, simple, "why does this exist, how do you use it, here is some sample code" concepts. The sample code should be REAL examples, not contrived. This is critical.

### Suggested Fix

Sweep the entire documentation and examples surface, ensure it is extremely high quality.

wallstop:
The agent-facing half of this shipped in #532, for pull request text specifically.

**The data.** The rule already said "under about 60 characters". Of the twelve titles before this
change, **nine were over 60** — median 85, longest 136:

```text
136  Hand an immutable contract the seed its constructor built, pin the NUnit Unity ships, and warn when ...
132  Construct a JSON converter where its closure lives, declare a hook where every reader runs it, and ...
112  Take large struct receivers by in, give UI Toolkit its extensions, and make a passing PRNG rating ...
106  Make a grid cell cost its distance from the origin, and stop one refused surrogate taking the rest ...
 51  perf: optimize serialization and random restoration
```

A soft rule that loses nine times in twelve is not a rule. It is now mechanical:

- **50 characters, hard**, with a `printf '%s' "$TITLE" | wc -c` check to run before sending.
- A copy-paste body template: one `**Why:**` sentence, two to five one-line `**What:**` bullets,
  `Fixes #123`, nothing else.
- A limits table, and an explicit never-include list — root causes, measurements, run IDs, session
  numbers, CI results, validation lists, byte traces, the diff, the file list.
- Short sentences, active voice, common words. A number only when the number is the point.

The guidance itself was rewritten in the style it asks for, and is about half its former length.
This PR's own title is 39 characters.

**This does not close the issue.** The ask here is the user-facing documentation under `docs/`, which
is untouched. Leaving open for that.

wallstop:
## Five user-facing defects fixed in https://github.com/Ambiguous-Interactive/unity-helpers/pull/591, and the gates that let them happen

Working the user-facing half. Each of these was a documentation claim or link that nothing verified:

[...]

PR #586, opened 2026-08-27
Strict UTF-8 decode, test-flake fix, docs sweep

### LINKED ISSUE #393: test: a Zenject scene test fails on Unity's temp-allocator lifetime warning, nondeterministically
## Symptom

`RelationalComponentsZenjectTests.SceneInitializerUsesMultiPassWhenConfigured` fails in **TearDown**,
never in the body:

```text
TearDown : Unhandled log message: '[Warning] Internal: deleting an allocation that is older than
its permitted lifetime of 4 frames (age = 6)'. Use UnityEngine.TestTools.LogAssert.
```

That warning comes from Unity's **native temp allocator**, not from any code in this repository. It
fires when a temp allocation outlives four frames, which depends on how many frames elapse while the
scene initializes — i.e. on machine load, not on logic.

## Why it is a fragile check rather than a defect

Observed on run [`31390509089`](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/31390509089),
`6000.5.2f1 standalone` on DAD-MACHINE. In the **same run**, the same test passed on
`2022.3 standalone`, `6000.3 standalone` and `6000.5 playmode`, each with zero leg failures. Same
commit, same code, same editor family — so nothing about the test's subject changed between the leg
that failed and the three that passed.

The test does not assert anything about allocator lifetimes. It fails because Unity's test framework
treats *any* unexpected log message as a failure, and this particular warning is emitted
nondeterministically by the engine during scene setup.

## Options

1. **Expect it where it can occur.** `LogAssert.ignoreFailingMessages = true` for the duration, or a
   targeted `LogAssert.Expect(LogType.Warning, new Regex("permitted lifetime"))`. Targeted is better
   — blanket ignoring would also swallow a real warning this test should catch.
2. **Reduce the frame count during initialization**, if the multi-pass path is holding a temp
   allocation across more frames than it needs to. Worth a look before suppressing: the warning is
   pointing at something real even if it is not a test failure.
3. **Shared handling.** If other scene-initialization tests can hit the same warning, the expectation
   belongs in the common fixture base rather than in one test.

Option 2 first — the warning names a real lifetime, and suppressing it without looking would be
treating the symptom. If the lifetime turns out to be inherent to how Unity schedules scene
initialization, then option 1, targeted.

## Not to be confused with

[#392](https://github.com/Ambiguous-Interactive/unity-helpers/issues/392) — that one fails **before
any test runs** and uploads no `results.xml`. This one has a full `results.xml` with 8,967 cases and
exactly one failure, which is how the two are told apart in ten seconds: **if the artifact has a
`results.xml`, repository code ran and it is not an infrastructure failure.**

### LINKED ISSUE #580: chore: sweep every decode path for strict fallback and refused-conversion discipline
## Summary

While implementing the `Uri` formatter for WallstopProto this week, two different silent-decode failure modes were hit, both inside one method boundary. Both would recur anywhere else in production code that decodes bytes into values. This asks for a scan and implementation pass over `Runtime/` (and a grep-only pass over `Editor/`) to make strict decoding deliberate everywhere.

## The two classes found

**1. Replacement-fallback decoding corrupts instead of refusing.**

`Encoding.UTF8.GetString(invalidBytes)` replaces invalid sequences with U+FFFD rather than reporting them. A decoder wired that way turns hostile or truncated input into a *different, valid-looking value* — for our case, a wrong Uri — with no error surface:

```csharp
// Today's fix: the only decoder that refuses
private static readonly UTF8Encoding StrictUtf8 = new UTF8Encoding(false, true);
```

Scope of suspicion (`rg -n "Encoding.UTF8.GetString|GetString\(" Runtime/`): JSON converters (`WGuidConverter`, `Hash128Converter`), binary helpers reading UTF-8 prefixes, compression paths, anything taking network/file bytes. Sites where the bytes provably came from our own writer may legitimately keep the forgiving decoder, but that needs to be a per-site decision, not an accident of which API was nearest.

**2. Parser laxness on text -> value conversion.**

The malformed-corpus probe also discovered `new Uri("http://example.com/%AAA")` parses happily while `%AAA` in authority position throws — "looks like it should fail" is not acceptance logic. Anywhere we convert decoded strings to typed values (Uri, Type.GetType-style lookups, enum parsing, version strings) should answer refuse-vs-accept from a stated rule, and refuses must be observable (TryX pattern returning false) rather than defaulted.

## What "done" means

- [ ] An inventory of every byte/character decode site under `Runtime/`, each marked strict / forgiving / ours-only, with the owning rationale.
- [ ] Forgive-mode sites carry a comment naming why corruption is impossible there.
- [ ] Untrusted-input decode sites use throwing decoders (`throwOnInvalidBytes`) or pre-validation, converting failures into the package's standard graceful-refusal (no throws crossing public APIs).
- [ ] Malformed-payload tests exist for each newly-strict path (the `BclDifferentialTests` malformed section is the pattern).
- [ ] `Editor/` gets a grep-level audit recorded here as evidence, not full parity; editor-only decoders feeding runtime saves are the exception worth checking.

## Non-goals

No new linter/diagnostic unless the inventory shows a recurring shape a human keeps re-mistaking — measure first.

### LINKED ISSUE #583: skill/ci: Unity_RunCommand entry-point contract drifted from the fixture-runner skill; executions fail closed with "No logs available"
`Unity_RunCommand`'s contract changed under this server build (unity-mcp-server 1.0.0 / protocol 2025-11-25), and `.llm/skills/unity-mcp-fixture-runner.md` no longer matches it.

## What was measured today

- Configuration + attach work exactly as documented: `npm run unity:mcp:configure`, `npm run unity:mcp:probe`, session handshake over streamable HTTP, `tools/list` enumerates all 50 tools, `Unity_ManageEditor GetState` answers.
- `Code` is compiled **as types-only** at namespace scope inside `Unity.AI.Assistant.Agent.Dynamic.Extension.Editor`: statement-level scripts fail CS8805 ("top-level statements must be an executable"), and the server wraps submissions, renaming them (`localFixedCode` shows the generated wrapper).
- The entry point is `IRunCommand.Execute(ExecutionResult)` (discovered via CS0535 when probing). Any class less accessible than its parameter must be `internal`.
- **Every execution of such a class fails closed with `UNEXPECTED_ERROR: Execution failed:\nNo logs available`** -- empty body, `result.Log(...)`, `result.LogError(...)`, and even throwing mid-body produce identical output, so output plumbing between the sandbox and the relay appears broken or renamed for legacy `IRunCommand` implementations.
- Compilation diagnostics still flow perfectly, which makes the compile-check use case fully functional today: submitting our changed package sources and reading back per-file/per-line errors works, and the host editor compiles the shared working tree.

## Ask

1. Update [the fixture-runner skill](./../../.llm/skills/unity-mcp-fixture-runner.md) with this server version's entry-point shape once someone discovers where execution output moved (suspects: a new interface inheriting `IRunCommand`, an attribute-based discovery, or `ExecutionResult` members renamed).
2. Until then, record that MCP-side validation supports **compile-level checks** only, and EditMode fixture sweeps belong to CI legs.

wallstop:
**Why:** A pass over the open-issue backlog.

**What:**
- Wire strings that are not valid UTF-8 refuse as malformed instead of decoding to replacement text, and `string.FromBase64()` refuses non-UTF-8 bytes the same way. Fixes #580
- PlayMode teardowns no longer fail on Unity's native temp-allocator lifetime warning. Fixes #393
- The Unity MCP RunCommand contract is measured and documented, including the invisible stale-assembly trap. Fixes #583
- The WallstopProto Uri formatter writes with the same encoder as every string field, so a lone surrogate cannot throw from a save (#399)
- `SerializationMutationTests` mutates valid payloads (bit flips, truncations) with a mutation-kill assertion (#462)
- Em-dashes and LLM vocabulary swept out of the documentation (46 files) (#478, #441)

---

> [!NOTE]
> **Medium Risk**
> Strict UTF-8 refusal changes how corrupt wire/base64 payloads behave (fail instead of silent replacement), which is correct for proto3 but may surface previously hidden bad data; devcontainer and doc edits do not affect shipped game code.
> 
> **Overview**
> **WallstopProto and base64** now treat invalid UTF-8 as malformed: shared `WProtoText.StrictUtf8` drives `WProtoReader.TryReadString` and strict `Uri` reads (latching `Malformed` / refusing the message), while `FromBase64()` stops inventing replacement-character text on corrupt bytes. **Uri writes** align with other string fields—forgiving encode so lone surrogates become U+FFFD on the wire rather than throwing during save.
> 
> **Tests and stability:** New `SerializationMutationTests` bit-flip/truncate valid proto and JSON payloads with a kill check that flipped Id bytes never decode as the original; PlayMode teardown tolerates Unity’s temp-allocator lifetime warning in `CommonTestBase` (#393). **Uri** differential test covers lone-surrogate write encoding.
> 
> **Dev container:** Replaces Codex-only install with `install-agent-clis.sh` (codex, opencode, nanocoder), user-global npm prefix without sudo, image-layer CLI baseline, `post-start` volume ownership repair, host `initialize-host.ps1` to remove zombie devcontainers after unclean Docker restarts, and post-create Unity MCP config sync from `.env.local`. **Docs:** Large punctuation/voice sweep across guides; `unity-mcp-fixture-runner.md` documents the measured RunCommand contract and stale-DLL diagnostics (#583).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af9ebe64bdb355984591571ce637fe316bc9dcac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Extension/StringExtensions.cs
```diff
@@ -1520,8 +1531,19 @@ private static bool TryDecodeBase64Utf8(string s, out string result)
                 }
             }
 
-            result = Encoding.UTF8.GetString(buffer, 0, outputLen);
-            return true;
+            try
+            {
+                result = StrictUtf8.GetString(buffer.AsSpan(0, outputLen));
```
wallstop, line 1536: Is there a TryGet* style variant that is non-throwing (for speed?)

Issue #323, opened 2026-07-30
ci: restore licensed compatibility, benchmarks, and release export after central cleanup support


wallstop:
## Context

Organization enrollment remediation in #322 consolidates paid Unity validation onto one canonical Windows lifecycle and retires benchmark and release activation paths that cannot currently satisfy the central cleanup contract. This issue tracks restoring that useful coverage without weakening fail-closed enrollment.

Central prerequisite: Ambiguous-Interactive/ambiguous-organization-build-lock#153.

## Required capabilities

- A reviewed container-aware return and cleanup-evidence contract for the exact editor instance that activates during `.unitypackage` export.
- Any required Darwin-native return support before reintroducing macOS paid activation.
- A representable per-leg fallback identity, or a reviewed typed aggregate for multiple literal holder identities.
- Immutable approved action pins, current-head guards, hosted runner preflight, central return/classification/release/gate ordering, and `cancel-in-progress: false` for every licensed path.

## Restoration scope

1. Restore the supported-version compatibility matrix without concurrent paid activation and without replacing literal holder identities with ambiguous dynamic cleanup.
2. Restore scheduled/manual licensed benchmarks with a complete lifecycle, hosted fallback, and truthful always-reporting aggregate.
3. Restore container-based `.unitypackage` export, then unblock dependent npm and GitHub Release publication only after cleanup is confirmed.
4. Add mutation-based static contracts for trust, cancellation, holder identity, return ownership, fallback, and aggregation.
5. Update the runner runbook and contributing guidance with the restored operating model.

## Acceptance criteria

- Central organization enrollment audit reports zero findings at the exact candidate commit.
- Exact-head hosted checks and each restored real Unity lifecycle pass.
- Every acquired path ends in confirmed cleanup or fail-closed quarantine/incident evidence.
- No raw activation/return evidence or credential material is uploaded or printed.
- Review confirms that the cleanup action observes the same editor/container context that consumed the license.

wallstop:
## Two of the retired workflows are still `disabled_manually`, and one of them is a scheduled refresh nobody is getting

Found while checking whether
[#431](https://github.com/Ambiguous-Interactive/unity-helpers/issues/431)'s scheduled benchmark half
was delivered. Queried against the repository's registered workflows:

```text
disabled_manually   Unity Benchmarks    .github/workflows/unity-benchmarks.yml
disabled_manually   Release Publish     .github/workflows/release.yml
```

[...]

Issue #322, opened 2026-07-29
ci: remediate organization Unity enrollment drift


wallstop:
## Problem

The central source-free organization audit reports **74 policy findings** for this repository at exact default-branch commit `3d4bdd1cf7e7d36fa63fa622bee67272796bab03`. This is a real fail-closed enrollment alert, not a Unity test run. Central tracker: [build-lock #113](https://github.com/Ambiguous-Interactive/ambiguous-organization-build-lock/issues/113). [Retained sanitized artifact](https://github.com/Ambiguous-Interactive/ambiguous-organization-build-lock/actions/runs/30439340249/artifacts/8718779776).

Finding summary: 8 each: missing classifier/gate/typed release/aggregate/return; 7 ineligible triggers; 6 each classifier/gate/release not always and missing acquire; 3 missing preflights.

The artifact contains repository, commit, workflow, job, classification, and reason codes only. Do not attach raw Unity logs, credential values, serials, tokens, PEMs, or matched source lines here.

## Required remediation

- Ensure every paid-serial job uses an approved immutable central lock pin and acquires before activation.
- Keep Unity/writer credentials out of workflow/job scopes that make unrelated jobs license-capable; scope them only to the exact steps that require them.
- Preserve literal `cancel-in-progress: false` and `fail-fast: false` on licensed paths.
- Add fail-closed runner preflight, Unity return, central cleanup classification, typed always-run release, and the always-run final cleanup gate in the same physical runner lifecycle.
- Emit one truthful always-reporting Unity aggregate; runner absence, cancellation, missing outputs, or partial execution must remain red.
- Remove, disable without credential reachability, or fully harden latent Unity workflows. Keep supported triggers limited to reviewed trusted refs/events.
- Run repository static contracts, the affected real Unity lifecycle, and the central organization audit against the merged default-branch commit.

## Acceptance criteria

- The central audit is complete and reports **zero findings for this repository** at its exact current default-branch commit.
- Every acquired path ends with confirmed cleanup or fail-closed quarantine/incident evidence; no skipped or ambiguous path reports green.
- Required aggregates and immutable pins are covered by data-driven contract tests.
- Paid Unity CI and default-branch CI are green on the final reviewed head, with no unresolved review threads.

Related: Ambiguous-Interactive/ambiguous-organization-build-lock#42, #44, #113.

wallstop:
## Current count, and a named root cause for 16 of the findings

[...]

PR #652, opened 2026-08-31
Catch seven Unity footguns, and the 111 sites they found

### LINKED ISSUE #616: No local gate type-checks Tests/Editor/** with the shipped analyzers
Found by a `WPROTO044` error that reached the Unity matrix from a file no local check project compiles.

## The hole

Three check projects exist, and between them they miss one tree:

| Project | Compiles |
| --- | --- |
| `TypeCheck` | `Runtime/**` |
| `EditorCheck` | `Runtime/**` + `Editor/**` |
| `TestCheck` | `Runtime/**` + `Tests/Core/**` + `Tests/Runtime/**` |

**Nothing compiles `Tests/Editor/**`.** `EditorCheck` builds the editor *sources*; `TestCheck`'s globs stop at `Tests/Runtime/**`. So an editor test fixture is type-checked by exactly one thing — a Unity matrix run, roughly fifty minutes away.

```console
$ grep -n 'Tests/Editor' Generator~/WallstopStudios.UnityHelpers.TestCheck/*.csproj
$        # nothing
```

That tree is not small: `Tests/Editor/` holds 20+ feature-scoped assemblies (`…Tests.Editor.WGroup`, `…Tests.Editor.CustomDrawers`, `…Tests.Editor.Validation`, and so on).

## What it cost

`WPROTO044` (#613) reported `FloatListCache` in `Tests/Editor/CustomDrawers/TestTypes/NestedCollectionSerializationTypes.cs`. Every other instance of that diagnostic in the repository was caught by `TestCheck` once its `obj/` was deleted; this one could only be caught by Unity, and it took its own CI round.

The same gap applies to every `WPROTO###` and `WUH###` the package ships, and to ordinary `CS####` in editor fixtures.

## Why it is not a one-line fix

Adding `Tests/Editor/**` to `TestCheck` does not work as-is:

1. **It needs `UnityEditor`.** `TestCheck` references only the community `UnityEngine.Modules`; the editor reference assembly lives in `EditorCheck` (`Unity3D.SDK` 2021.1.14). So the editor tests belong with `EditorCheck`, or in a fourth project.
2. **`EditorCheck` excludes seven `Editor/**` files** for reference-assembly reasons (two permanent uGUI ones, three pin-driven, two cascades). A test fixture binding one of those excluded types would fail for a reason that is not the author's — the same trap the exclusions were written to avoid.
3. **`overrideReferences` fidelity.** `lint-typecheck-asmdef-references` exists because a typecheck project compiling many asmdefs into one assembly is more permissive about references than Unity is (#598). Adding 20+ editor test asmdefs widens that surface, so the new project has to be governed by the same linter, and its probe leg extended.

## Acceptance criteria

- A local command type-checks `Tests/Editor/**` against the Unity editor reference assembly with the shipped analyzers loaded.
- It is governed by `lint-typecheck-asmdef-references` like the other three, and `test-lint-typecheck-asmdef-references` knows about it.
- Any fixture it cannot compile is excluded **one per line with its reason**, the way `EditorCheck` does it, rather than through a packed `Exclude`.
- Reintroducing a `WPROTO###` into an editor fixture is caught by it, proven by doing so.

## Related

While confirming this, `npm run typecheck:tests` was found to report a pass on a tree with four `WPROTO044` errors: the `.cs` files had not changed, so MSBuild considered the compile up to date and the freshly built analyzer never ran. That is written up in `validate-before-commit` — a rebuilt `Runtime/Analyzers/*.dll` is a changed input no check project honours without `obj/` being deleted. The two together are why a new diagnostic needs `rm -rf Generator~/*/obj` **and** still cannot see one of the four source trees.

### LINKED ISSUE #621: [Feature]: analyzer for ?. and ?? on UnityEngine.Object, and the Assert.IsNotNull that passes over a destroyed one
`UnityEngine.Object` overloads `==` to report a destroyed object as null. The C# null-conditional and
null-coalescing operators do **not** use that overload -- they test CLR null. So on a destroyed
object `obj?.Foo` runs the member access and `obj ?? fallback` returns the destroyed object, both
silently and both at exactly the moment the guard was written for.

The package ships `Objects.NotNull` and friends for this, so the replacement already exists; nothing
catches the mistake.

## Three shapes, and the second is the worst

**1. `?.` / `??` / `??=` on a `UnityEngine.Object`.** The ordinary case. Three sites in the project
measured, and they were a level's win condition and, twice, the id a save keys its counts on.

**2. `Assert.IsNotNull(unityObject)` -- and this one fails the *other* way.** `Assert.IsNotNull`
compares to CLR null, so it **passes** over a destroyed object: a fixture that is green about a thing
that is gone. **79 assertions** were in that state in one project. The house form is
`Assert.IsTrue(x != null)`, which goes through the overload.

**3. `== null` on the way *out* of a collection that outlives its entries.** A destroyed object
answers `true` to `== null` while its reference is still a perfectly good dictionary key, so a
removal guarded that way refuses exactly the entries nothing else can reach. `ReferenceEquals(x,
null)` asks "was anything passed"; `x == null` asks "is it gone". They are different questions and
the wrong one leaks.

## Why an analyzer and not a linter

**The signal is the type, not the operator.** `Vector2? p; p?.x` is correct and common -- a nullable
value type is exactly what `?.` is for -- so a regex over `?.` reports mostly false positives, and one
over "does the receiver look like a component" misses every alias and every generic. It needs the
class hierarchy: is the receiver's type assignable to `UnityEngine.Object`? That is a semantic model
question, which is what the package's analyzer pipeline already answers for its other diagnostics.

## Recommended implementation

Three diagnostics, so each can be suppressed on its own:

- **`WUNITY001`** -- `?.`, `??` or `??=` whose receiver (or left operand) is a `UnityEngine.Object`.
  Fix: an explicit `!= null` check, or `Objects.NotNull`.
- **`WUNITY002`** -- `Assert.IsNotNull` / `Assert.IsNull` / `Assert.AreEqual(null, …)` on a
  `UnityEngine.Object`. Fix: `Assert.IsTrue(x != null)`. Test assemblies only.
- **`WUNITY003`** -- `== null` / `!= null` on a value that arrived as a **key or element of a
  collection** being iterated for removal. Narrower and more speculative than the first two; if it
  cannot be made precise, ship the first two and document this one.

A source-only linter was tried first in the project measured and had to be rewritten to resolve the
class hierarchy across 41 root types before it was usable, which is the argument for doing it in the
compiler instead.

### LINKED ISSUE #622: [Feature]: analyzer for UnityEngine.Random, whose state no test can set or read
The package ships ~20 generators behind `IRandom`, all **seedable and serializable**, plus
`PRNG.Instance`. `UnityEngine.Random` is a process-global whose state a test can neither set nor read
without disturbing every other caller.

So a project that reaches for `UnityEngine.Random` loses replayability, and nothing says so.

## Use case

The reason is not taste. Anything built on an `IRandom` can be **re-run and asserted**: a spawn table,
a scatter, a shuffle, a procedural layout. The same code on `UnityEngine.Random` produces a bug report
that says "sometimes the fruit lands inside the wall" and no way to reproduce it. Swapping afterwards
is a behaviour change to every call site at once, which is why it wants catching at the first one.

`UnityRandom` (the package's own `IRandom` wrapper) is not an escape either -- #521 records that it is
the one generator whose snapshot cannot resume its stream -- so the diagnostic should point at
`PRNG.Instance`, not at it.

## Recommended implementation

**`WUNITY010`** -- any member access on `UnityEngine.Random`. Semantic, so an alias or a
`using static` cannot dodge it.

- `System.Random` is a **different** mistake and is out of scope; conflating them makes the
  diagnostic unactionable.
- Editor-only code is a legitimate exemption (a one-off authoring scatter has nothing to replay), so
  the analyzer should be suppressible per-assembly rather than absolute.
- The fix message should name `PRNG.Instance` and mention that a field of type `IRandom` is what makes
  it seedable from a test.

## The companion trap, worth putting in the same diagnostic's docs

`NextFloat(min, max)` throws when `max <= min`. An authored spread that may legitimately be `0` --
"no scatter" -- is the obvious thing a designer types, and it raises on every call. #546 added the
non-throwing draw; the analyzer's fix text is the natural place to point at it, because a project
converting off `UnityEngine.Random` (which clamps instead of throwing) will hit this on the way.

## Cheapest possible moment

The project that motivated this had **zero** `UnityEngine.Random` call sites when the rule was
written. A rule adopted at zero costs nothing and never has to be burned down, which is the argument
for shipping it before anyone needs it.

### LINKED ISSUE #623: [Feature]: analyzer for a discarded EffectHandle, the only way to remove an infinite effect
`TagHandler.ApplyEffect` hands back an `EffectHandle`. Where every `AttributeEffect` in a project is
`ModifierDurationType.Infinite`, **nothing expires on its own and that handle is the only thing that
can take the effect off.** Discarding it is unrecoverable, and it compiles clean.

## The failure, from a real one

An immobilise applied to the **player** for the duration of a cutscene. The handle was discarded, the
cutscene ended by a route that did not run the release, and the player spent the rest of the run
unable to move, act or collide. No error, no log, nothing red -- the effect was applied exactly as
asked and simply never removed.

**The object carrying the effect routinely outlives whatever applied it.** That is the shape: a
summoner, a trigger volume or a cutscene director applies a hold to the player and then goes away.
Whatever holds the handle has to outlive the effect, not the applier.

## Use case

Any consumer of the Tags system with an infinite-duration effect -- which is the default a designer
lands on, because a duration is a number somebody has to choose. The package hands back the only
means of release and does not mind if it is dropped.

## Recommended implementation

**`WTAGS001`** -- an `ApplyEffect` call whose returned `EffectHandle` is discarded: an expression
statement, a `_ =` discard, or a value that is never stored.

- `ForceApplyEffect` is the deliberate no-handle overload for instant effects and must be out of
  scope; the doc comment on `ApplyEffect` already points at it, so the diagnostic's fix text should
  too -- "if this effect is instant, call `ForceApplyEffect`" is the one-line resolution for most
  hits.
- **Do not gate on the duration type.** It is authored data, not visible to the compiler, and an
  effect can be re-authored to `Infinite` long after the call site was written -- which is precisely
  how the failure above arrives.
- A site that genuinely means "apply forever, never release" should say so at the site with a
  suppression carrying a reason, so the decision is visible where it is made.

A source-level linter for this exists in the project measured and found one offender the day it was
written -- the one described above.

### LINKED ISSUE #626: [Feature]: analyzer for a discarded coroutine handle, which is the only thing that can stop the work
#359 fixed the *body* -- a periodic job's action is guarded now, so one exception no longer stops it
permanently. The **handle** is the other half and nothing checks it.

`StartCoroutine` returns a `Coroutine`. That handle is the only thing that can stop the work, and the
only thing that can answer "is this already running". Drop it and neither question has an answer.

## Three shapes, all silent

**1. The handle is never held.** `StopAllCoroutines` is the only remaining lever, and it also stops
the coroutine doing the stopping. `private IEnumerator Start()` hands back no handle at all, so a
`Start` that loops cannot be stopped by anything but destruction.

**2. The handle is held and overwritten.** *Stop what it points at before assigning*, or the field
stops describing what is running and the orphan keeps mutating shared state. This is the one that
produces "it got faster every time I re-triggered it".

**3. One owner starts many.** A single field cannot describe it, so overwriting is all it can do. The
answer is a `List<Coroutine>` the owner clears where its state ends -- not a bigger guard. In the
project measured this was the only genuine exception to rule 2, and the list removed it rather than
excusing it.

## Why the obvious matcher misses most of it

**`StartCoroutine` is usually not the starter a project uses.** In the tree measured, the package's
own `StartPeriodicJob`-style helpers and a delay extension each **outnumbered** raw `StartCoroutine`,
so a rule matching that one name saw **9 of 44** call sites. Any diagnostic here has to be told the
full set of starters, which is an argument for shipping it beside the starters -- i.e. here.

## Use case

Any component that starts recurring or long-running work: a blink, a patrol, a poll, a fade. The
symptom is state that keeps changing after the thing driving it should have stopped, which is
attributed to almost anything else.

## Recommended implementation

**`WUNITY020`** -- a coroutine-starting call whose returned handle is discarded.

- The starter set must be configurable, or at minimum cover the package's own helpers as well as
  `MonoBehaviour.StartCoroutine`.
- A site that must outlive its starter is legitimate; it should say so at the site with a suppression
  carrying a reason, and the reason should be printed by whatever reports the diagnostics so the
  exemptions stay readable rather than accumulating.
- **Shape 2 is deliberately out of scope for a first version.** Whether a stop reached a handle
  before it was reassigned is a dataflow question, and the four correct reassignments in the project
  measured each spelled their guard differently. Ship shape 1, document shape 2.

Where many one-shot delays share an owner, a tracked-delay collection that does not accumulate dead
handles is the natural companion API, and the package is the right home for it.

### LINKED ISSUE #629: [Feature]: analyzer for reading a TryXxx out parameter when the call returned false
`TryGetValue` and friends write `default` to their `out` on a miss. That is what makes this look
harmless:

```csharp
_ = map.TryGetValue(key, out Thing thing);
thing.DoSomething();      // compiles, and reads a default nobody authored
```

The BCL happens to write `default` on failure. **Nothing obliges anyone else's `TryXxx` to**, and the
package ships plenty of them -- so the same pattern over a project's own API reads whatever the callee
left in the slot.

## Why it is worth a diagnostic

It is the mirror of the rule most codebases already hold about `out` parameters: assign them
immediately before each `return`, never once at the top, because an up-front assignment **disables the
compiler's definite-assignment check**. That rule exists precisely so a `TryXxx` can be honest about
what it wrote. Reading the `out` after `false` throws that away from the caller's side.

The failure is quiet in the worst way: a `default` struct or a `0` count is a plausible value, so the
symptom is wrong behaviour rather than a crash. In the project measured this class sat alongside four
components that `+=` an untrusted count and then compare it to a bound -- where a `0` silently
un-does progress a player really made.

## Use case

Any `TryXxx` on the package's own surface, and any consumer writing one. The diagnostic also makes
the package's `TryXxx` contract enforceable rather than conventional.

## Recommended implementation

**`WTRY001`** -- a read of a `TryXxx` `out` variable on a path where the call's `bool` result was not
tested.

Scope, from building this as a source linter:

- **`out _` is fine.** So is a discarded call that reads nothing afterwards -- `_ = set.TryAdd(x);` is
  a legitimate "add if absent".
- The signal is a **read of the out variable**, not the discard. Matching `_ =` alone reports every
  fire-and-forget `TryAdd` in the tree.
- Treat `if (!TryX(out v)) { return; }` and `if (TryX(out v)) { … }` as tested; that is the overwhelming
  majority and must not be reported, or the rule is unusable.

### LINKED ISSUE #630: [Feature]: analyzer for a teardown whose base call runs before the body that still needs it
A teardown that calls its `base` first can take away the thing its own body still needs.

```csharp
protected override void OnDestroy()
{
    base.OnDestroy();          // deregisters the singleton, tears down the messaging token
    Announce(...);             // now has nothing to announce through
    SomeSingleton.Instance...  // now fabricates or throws
}
```

The base call is where a `RuntimeSingleton` drops its registration and where a messaging component
releases its token. Anything the derived body does **after** it is operating on a half-dismantled
object, and the failure is a beat that quietly does not happen at the end of a level or a scene --
the moment nobody is watching a console.

## Why the package is the right home

Both halves of the hazard are things the package (or its immediate neighbours) provide: the singleton
registration, and the base `OnEnable` / `OnDisable` / `OnDestroy` a consumer is told to override and
chain. The rule is a property of *that contract*, so it belongs beside it rather than being
rediscovered per project.

There is a real asymmetry to encode, and it is why "always call base first" is wrong advice:

- **Setup chains base-first.** The base has to have registered before the body uses what it
  registered.
- **Teardown chains base-last.** The body has to finish using it before the base takes it away.

## Use case

Any consumer overriding a lifecycle method on a package base class -- which is the normal way to use
`RuntimeSingleton` and the messaging components. The mistake is natural, because base-first is
correct everywhere else in the same file.

## Recommended implementation

**`WUNITY040`** -- in an override of `OnDisable` / `OnDestroy` (and any package teardown hook), a
`base.<Method>()` call that is **not the last statement**, where statements follow it.

- Report the statements that follow, not just the call, so the fix is visible.
- A body that only logs after the base call is harmless but not worth carving out: moving one line is
  cheaper than a suppression, and an exception list here would be read as permission.
- The mirror rule for setup (`base` first in `Awake` / `OnEnable`) is worth the same diagnostic id
  family, but it is a separate check and a separate default -- do not fold them into one "call base
  in the right place" rule, because the right place is opposite.

### LINKED ISSUE #632: Extend IListExtensions to Span<T>: a shuffle you can use on a stack buffer, without changing what the RNG draws
## What is missing

`IListExtensions` is `IList<T>`-only, so a caller holding a stack buffer or a slice cannot use any of it. In practice that means the allocation-free path is unavailable exactly where it would pay most: a small, fixed-size, shuffled working set.

The concrete shape, from a downstream game (DoxReloaded):

```csharp
// once per generator call, ~1,266 calls to build 112 levels
ButtonDirection[] directions = ButtonDirectionExtensions.NavigableDirections.ToArray();
directions.Shuffle(random);
```

`NavigableDirections` is a `static readonly T[]` of **four** elements. It is copied only because `Shuffle` mutates in place and the shared static must not be mutated. The natural fix is a stack buffer:

```csharp
Span<ButtonDirection> directions = stackalloc ButtonDirection[4];
NavigableDirections.CopyTo(directions);
directions.Shuffle(random);   // ← does not compile: Shuffle<T>(IList<T>, IRandom)
```

That pattern appears **7 times** in one downstream module. Measured there: the button generators account for **65 % of generation allocation — 21.4 MB over 1,266 calls, ~17.3 KB per call** — on a path that runs while a player waits for a level.

## Why this is close to free to add

`ShuffleArray` is **already span-shaped** — it only ever does indexed reads, writes and swaps over a contiguous buffer:

```csharp
private static void ShuffleArray<T>(T[] array, int count, IRandom random)
{
    for (int i = 0; i < count - 1; ++i)
    {
        int nextIndex = random.Next(i, count);
        if (nextIndex == i) { continue; }
        (array[i], array[nextIndex]) = (array[nextIndex], array[i]);
    }
}
```

Changing that parameter to `Span<T>` and adding `public static void Shuffle<T>(this Span<T> span, IRandom random = null)` costs nothing at the call sites that exist: the `T[]` fast path in `Shuffle<T>(IList<T>, …)` already reaches this body and an array converts implicitly.

🚨 **And it gives the property that actually blocks adoption: draw-for-draw identity.** A shuffle consumes exactly `count - 1` calls to `random.Next(i, count)`, in that order. Any *reimplementation* of Fisher-Yates for spans could consume the RNG differently, and for a downstream user with **seeded, reproducible generation** that silently changes every generated artifact. Sharing the one body makes "switching to the span overload cannot change my output" a fact rather than a hope — and that guarantee is worth stating in the doc comment, because it is the reason a caller can adopt it at all. (This is not hypothetical: it is why the downstream change was abandoned rather than written locally.)

## The rest of the surface, ranked by how much we would use it

1. **`CopyShuffled<T>(ReadOnlySpan<T> source, Span<T> destination, IRandom random = null)`** — the single most useful addition, because it *is* the pattern above in one call: copy a shared read-only sequence into caller-provided storage and shuffle it, allocating nothing. Every one of the 7 sites collapses to this.
2. **`Shuffle<T>(this Span<T>)`** — as above.
3. **`GetRandomElement<T>(this ReadOnlySpan<T>, IRandom)`** — the read-only companion; picking from a static table without copying it.
4. **`Swap<T>`, `Fill<T>`, `Reverse<T>`, `RotateLeft/Right<T>`, `Shift<T>` for `Span<T>`** — all are index-only bodies today, so all are mechanical.
5. **`IndexOf<T>(this ReadOnlySpan<T>, Func<T, bool>)` / `LastIndexOf`** — predicate search without materialising a list.
6. **Destination-taking variants of the two that allocate today**: `FindAll<T>` returns a `List<T>`, and `Partition<T>` returns *two* `List<T>`. `TryFindAll(ReadOnlySpan<T> source, Span<T> destination, …, out int written)` and a `Partition` writing into two caller spans would make both usable in a hot loop.

⚠ **One caveat worth designing around rather than discovering:** `Span<T>` cannot be captured by a lambda or used in an `async`/iterator method, so the predicate-taking overloads need to stay non-capturing (or take a `TState` parameter, `Func<TState, T, bool>`) to be usable from the places that most want them.

## Suggested scope

Items 1–2 alone unblock the downstream change and are a small, self-contained PR with an exact-equivalence argument. 3–6 are the natural follow-on and can land separately.

Happy to send a PR for 1–2 if that is welcome.

wallstop:
**Why:** every rule here was written against a shape the package documents and tells consumers to avoid — and had violated itself in 171 places, including six destroyed-object guards that did nothing.

**What:**

- Add `WUH003`-`WUH009`: `?.`/`??` on a `UnityEngine.Object`, an NUnit null assertion that passes over a destroyed one, `UnityEngine.Random`, a discarded `EffectHandle` or coroutine handle, an untested `TryXxx` `out`, and a teardown whose `base` call is not last.
- Add `WUH010` for a hard dictionary indexer read, opt-in — the family's first off-by-default member.
- Sweep all 171 sites; ten were real defects, not restatements.
- Add `Span<T>` shuffling that draws identically to the `IList` path, because it is the same body.
- Add `EditorTestCheck`, which type-checks `Tests/Editor/**` — the one tree nothing compiled.

**External checks, per [ship-changes Step 10](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/.llm/skills/ship-changes.md):** `Running Copilot Code Review` fails with *"the user who requested the review has reached their quota limit"* — an account entitlement, not the diff. It has failed on every run since session 238 and re-pushing cannot clear it. `Unity Tests` is queued on licensed runners, as it is on `main` (#322, #323).

Fixes #616
Fixes #621
Fixes #622
Fixes #623
Fixes #626
Fixes #629
Fixes #630
Fixes #632


Editor/Sprites/AnimationCreatorWindow.cs
```diff
@@ -1994,31 +1994,29 @@ private static string StripDensitySuffix(string name)
         private static bool TryExtractBaseAndIndex(string name, out string baseName, out int index)
         {
             Match m = s_ParenIndexRegex.Match(name);
-            if (m.Success)
+            if (m.Success && int.TryParse(m.Groups["index"].Value, out int parenIndex))
```
wallstop: Should this be using TryGetValue? I want to avoid hard-indexing on dictionaries and force Try* style APIs throughout, (and potentially opt-in warn on consumers).

wallstop: Taking it, and the general form too.

On this line specifically you are right for a reason beyond style: `GroupCollection`'s string indexer never throws — it returns a **non-participating** `Group` whose `Value` is `""`. So if someone renames a group in the regex above, `Groups["index"]` keeps compiling, `int.TryParse("")` fails, and the method silently stops recognising indices. That is the same silent-fall-through class as the rest of this PR, and `TryGetValue` turns it into a visible miss. `GroupCollection` implements `IReadOnlyDictionary<string, Group>`, so it is literally a dictionary read.

Doing three things:

1. Converting this site and sweeping `Runtime/` and `Editor/` for hard indexer **reads** on dictionaries. Writes (`map[key] = value`) stay — that is add-or-update and there is no better `Try` form.
2. Adding `WUH010` for the shape, so consumers get it too. **Opt-in (`isEnabledByDefault: false`)**, which is a deliberate deviation from the `WUH###` family contract and I will record it: a dictionary read where the key is known present is correct and ubiquitous, so on-by-default would bury the other nine on a consumer's first build. Enabling it is one line in `Assets/Default.ruleset`. The package will hold itself to it via a ruleset on the four local check projects, where warnings are errors.
3. Updating the agentic guidance so a future session finds inline review threads at all — `ship-changes` tells an agent how to answer review feedback but never says where to look for it, and inline threads are behind a different API from PR comments, so a session polling only `/issues/{n}/comments` sees nothing and reports "no feedback".

Will report the sweep count here before landing it, in case the shape is more common than it looks.

wallstop: Done, all three. Pushed as `03d5a8d5`. Reporting the count as promised, because it decided the scope.

**Enabled across all four local gates, `WUH010` reported 346 sites:**

| tree       | sites |
| ---------- | ----: |
| `Runtime/` |     9 |
| `Editor/`  |    51 |
| `Tests/`   |   286 |

The test number is not 286 defects. It is dominated by fixtures whose subject **is** the indexer — `DictionaryExtensionTests` 54, `SerializableDictionaryTests` 31, `SerializableDictionaryJsonTests` 16, `EditorCacheHelperBoundedCacheTests` 16, and ten more. Rewriting `Assert.AreEqual(1, map["a"])` in a dictionary's own test deletes what it tests.

So: `Generator~/CheckProjects.ruleset` holds **`TypeCheck` and `EditorCheck`** to the rule, where warnings are errors, and the two test projects are exempt **by project rather than by 286 pragmas** — a boundary you can see, where a pragma thicket is one you have to reconstruct. #653 carries splitting the ~50 genuine test accidents from the fixtures. Say the word if you'd rather have the pragmas.

**All 60 production sites swept, zero pragmas.** All four gates 0 errors / 0 warnings from a clean `obj/`. Six sites changed behaviour rather than being restated — each was a path that previously threw from inside a drawer's `OnGUI` and now degrades (a missing scan result renders the "Scan to see potential changes" hint; an ungrouped button falls back to its own `DeclarationOrder`). `SerializableDictionary`'s indexer still throws, per the `IDictionary` contract, but now names the key.

**Two things I asserted to the agent doing the work turned out to be false, and measuring beat arguing both times:**

1. I claimed `GroupCollection` implements `IReadOnlyDictionary<string, Group>`. It does — **from .NET Core 3.0**, and not on the netstandard2.1 profile Unity compiles against. An interface-only type test reported `match.Groups["name"]` in a net9.0 unit test and **nowhere a consumer builds**, including this exact line. The analyzer matches the type by name as well now, and a test declares an interface-less `GroupCollection` so deleting that gate fails the suite.
2. `GroupCollection.TryGetValue` does not exist on either target, and neither does `Group.Name` on netstandard2.0. The portable route is `Regex.GroupNumberFromName` plus the int indexer, wrapped in `Editor/Extensions/RegexExtensions.cs`. That closes the hazard better than `TryGetValue` would have: `GroupNumberFromName` returns a negative number for a name the pattern never declared, so an **undeclared** name and a **non-participating** group stop being indistinguishable — which was the actual silent failure here.

On the third item: `npm run pr:feedback -- <number>` now prints inline threads, review submissions, conversation comments and failing checks in one pass, and `ship-changes` gained a step saying which endpoint each lives on. It paid for itself on the first run — the Copilot reviewer's reason for failing ("reached their quota limit") lives in `GET /pulls/{n}/reviews`, which is neither the run log nor the comments, so that check has been read as an unexplained failure for several sessions.

Issue #653, opened 2026-08-31
Hold the test trees to WUH010 too, once the dictionary fixtures are separated from the accidents


wallstop:
Follow-on to the `WUH010` work in [#652](https://github.com/Ambiguous-Interactive/unity-helpers/pull/652), which came out of an inline review comment: *"I want to avoid hard-indexing on dictionaries and force `Try*` style APIs throughout."*

`WUH010` (a read through a dictionary indexer, which throws `KeyNotFoundException` on a miss where `TryGetValue` reports) ships opt-in, and the package holds **`TypeCheck` and `EditorCheck`** to it through `Generator~/CheckProjects.ruleset`. `TestCheck` and `EditorTestCheck` are deliberately not wired to it yet.

## The measurement that made that the call

Enabling the rule across all four check projects reported **346** sites:

| tree      | sites |
| --------- | ----: |
| `Runtime/` |     9 |
| `Editor/`  |    51 |
| `Tests/`   |   286 |

The 60 production sites were swept in #652. The 286 are not 286 defects — they are dominated by fixtures whose **subject is the indexer**:

| file                                                    | sites |
| ------------------------------------------------------- | ----: |
| `Tests/Runtime/Extensions/DictionaryExtensionTests.cs`  |    54 |
| `Tests/Runtime/DataStructures/SerializableDictionaryTests.cs` | 31 |
| `Tests/Editor/WButton/WButtonConflictingDrawOrderTests.cs`    | 21 |
| `Tests/Runtime/Serialization/SerializableDictionaryJsonTests.cs` | 16 |
| `Tests/Runtime/DataStructures/ProtobufSerializationTests.cs`     | 16 |
| `Tests/Editor/Core/Helper/EditorCacheHelperBoundedCacheTests.cs` | 16 |
| `Tests/Editor/WButton/WButtonConflictDetectionTests.cs`          | 14 |
| … 13 more files                                                  | ~119 |

Rewriting `Assert.AreEqual(1, map["a"])` in a dictionary's own test to `TryGetValue` deletes the behaviour under test. Exempting by **project** was chosen over 286 individual pragmas or 14 file-level ones, because a project boundary is a decision a reader can see, and a pragma thicket is one they have to reconstruct.

## What is actually left to do

Roughly 50 of the 286 are **not** in a dictionary-fixture file, and some of those are the ordinary accident the rule is for: a test reading a config map by key and getting `KeyNotFoundException` instead of a readable assertion failure.

[...]

Issue #635, opened 2026-08-30
less comments


wallstop:
We should encode in our agentic knowledge to strive for minimal, ideally zero, comments. Comments should only ever be for exceptional circumstances.

All multi-line, non-doc comments should follow the format:
```csharp
/*
    This is a 
    multi-line comment
*/
```

And we should aim for extremely readable class, function, and variable names. Wherever possible, use more descriptive naming standards, instead of shortening things.

We need to audit, and clean up, pretty much all existing code. There is so much random, generally. useless comments. A full, complete, audit, across all prod, runtime, editor, tests.

wallstop:
The knowledge half is done in [#660](https://github.com/Ambiguous-Interactive/unity-helpers/pull/660); the audit is not started.

## What changed

`.llm/context.md` rule 8 and `create-csharp-file.md`'s "Minimal Comments" section now say **aim for zero**, that the first move when a comment feels needed is a better name rather than a better sentence, and that names are spelled out rather than abbreviated. The multi-line block form is specified with an example.

## One correction to the rule as you stated it

"All multi-line, non-doc comments should follow the `/* */` format" condemns the **two-line MIT license header every one of the package's ~1,900 C# files carries**, which `scripts/lint-license-headers.ps1` matches literally. Rewriting it would be a 1,900-file diff that changes nothing a reader cares about, so the rule is scoped to a comment **inside a type or a member**, with the banner named as the fixed exception. Say the word if you want the banner converted too and it is one sed.

## What the audit is, measured

Scanning the 84 C# files this PR touches for runs of two or more consecutive `//` lines finds **289 runs**. Excluding the license banner (one per file) leaves roughly 200 in files this PR touched alone — and it touched about 4% of the tree.

I converted only the ~13 runs **this PR introduced**, in my own commits and in the sub-agents'. The pre-existing remainder is the audit, and it is not something to do by hand:

1. **Write `npm run lint:comment-style` with a `--fix`**, in the shape `lint-comparison-direction` already has: a scanner that reports, a fixer that rewrites what it can and names what it declines. The declining half matters — a `//` run that ends in a line of code, or one inside a `#if`, needs a human.
2. **Run the fixer, review the diff by file rather than by hunk**, because the second half of the ask ("delete the useless ones") is a judgement the fixer cannot make.
3. **Then wire it into Repo Lint**, not before — a rule the tree violates in 200 places is a rule nobody can follow.

[...]

PR #665, opened 2026-08-31
Check the assets, not just the code

### LINKED ISSUE #617: [Feature]: a validator and safe repair for serialized keys whose field no longer exists
`SerializedFieldValidator` asks which fields a developer marked for serialization Unity **declined to
write**. The mirror question has no tool: **which keys are already written that no field claims?**

## What happens

Unity keeps an unknown serialized key on load and writes it straight back out. So a field deleted in
2023 is still in the YAML today, and it reads exactly like a live one. Nothing reports it -- not the
inspector, not a build, not a domain reload.

Measured in a game project that migrated off two serialization packages: **1301 keys across 142
assets**, from `serializationData` (a removed Odin), `_debug` / `_production` /
`relatedLibraryObject` / `guid` / `icon` / `color` (a removed Databrain), and fields ordinary
migrations retired.

## Why it is worth a tool rather than a shrug

Three costs, and the second and third are the expensive ones.

1. **Diff noise forever**, in every asset of an affected type.
2. **A retired object-reference key keeps another asset looking referenced.** A guid search answers
   "this sprite is used by the level prefab" about a slot nothing reads. That cost two sessions of
   measurement on a sprite nothing rendered.
3. **A gate that reads committed YAML by current field name is blind to every asset written before a
   rename.** `[FormerlySerializedAs]` keeps the *game* correct and leaves the *file* saying the old
   name, so a project-wide authoring check silently measures only the assets somebody happened to
   re-save. One such gate was found green-about-nothing on ten entries this way.

## Recommended implementation

Two pieces, both `Editor/Validation/`, beside `SerializedFieldValidator`.

### `StaleSerializedKeyValidator`

`Scan(IReadOnlyList<string> assetPaths, List<StaleSerializedKey> findings, out int unresolvedScripts)`.

- Walk `--- !u!114` documents in `.asset` / `.prefab` / `.unity` text, read `m_Script`'s guid, and
  resolve it through `AssetDatabase.GUIDToAssetPath` -> `MonoScript.GetClass()`.
- **The declared set must be `SerializedObject` over a throwaway instance, not reflection over the
  type's fields.** Only Unity knows what its serializer accepted, and the base chain routinely leaves
  the project into package assemblies. Modelling it instead was measured reporting ~1771 findings of
  which about four were real. Reuse `SerializedFieldValidator`'s probe discipline exactly, including
  `SetActive(false)` **before** `AddComponent`.
- **Every `[FormerlySerializedAs]` alias is a live key** and must be added, walking the base chain.
  Judging by `SerializedObject` alone reported 565 aliases doing their job as orphans in one project
  -- including the only `clip:` key on a live effect, which a reader following the gate would have
  deleted.
- Report `<Type>::<key>` as the *cause* (one entry per type-and-key) alongside the per-site line,
  because a migration retires a field once and every asset of that type inherits it.
- A document whose `m_Script` does not resolve is **counted, not reported** -- that is a missing
  script, a different defect with its own signal, and guessing at its keys reports all of them.

### `StaleSerializedKeyRepair`

There is no "delete this key" API. The lever is
`AssetDatabase.ForceReserializeAssets(paths, ReserializeAssetsAndMetadata)`, which rewrites the asset
from what it loaded: every dead key dropped, every new field written at its default, every
`[FormerlySerializedAs]` alias migrated to its current name. One call cleared all 1301.

**It is not safe unsupervised, and that is the reason the wrapper is the deliverable rather than the
one-liner.** Three findings, each of which cost a round trip:

1. **An asset whose content lives in sub-objects can come back with them gone.** A URP
   `VolumeProfile` went from **20 serialized documents to 1**, losing all 19 volume components --
   every camera inheriting it would render with no post-processing. `ForceReserializeAssets` reports
   success.
2. **Restoring the file is only half of an undo.** The editor still holds the damaged object, so the
   next save writes it straight back out; it broke a second time minutes after a `git checkout`. The
   other half is `AssetDatabase.ImportAsset(path, ForceUpdate | ForceSynchronousImport)`.
3. **Prefabs need `ReserializeAssetsAndMetadata`.** With assets-only they are silently not rewritten
   at all -- measured on ten of them, which read as "these had no stale keys".

So: repair **one asset at a time**, comparing `AssetDatabase.LoadAllAssetsAtPath(path)`'s non-null
count before and after, and undo any rewrite that lowers it by writing the original bytes back and
force-importing. `LoadAllAssetsAtPath` rather than a count of YAML documents, because a
`VolumeProfile`'s components are `HideInHierarchy` sub-objects that a text reader can disagree about.

### Menu

`Tools/Wallstop Studios/Unity Helpers/Report Stale Serialized Keys` (selection or whole project) and
a separate `… /Repair Stale Serialized Keys In Selection`, so reporting is never destructive. The
repair should print the refused assets loudly; the one that broke in the measurement above was in the
set only because a crude line matcher hit a live `Vignette.color` and read it as the retired one, so
**a bad subject list is the likely failure and the wrapper has to survive it**.

## Prior art

The validator half exists as a project-local EditMode test and the repair was performed by hand;
both are described with their measurements in
<https://github.com/Ambiguous-Interactive/IshoBoy/pull/745>.

### LINKED ISSUE #618: [Feature]: nothing enforces [WNotNull] outside the inspector, so a build ships with the slot empty
`WNotNullAttribute` is a `PropertyAttribute`. It draws a warning beside the field in the inspector and
does nothing else. `WNotNullAttributeExtensions.CheckForNulls` throws on a null and, in the consuming
project measured here, had **zero call sites**.

So the package tells an author "this must be assigned", and a build ships with the slot empty.

## Use case

A designer adds a prefab variant, forgets one reference, and nobody has the asset open again. The
component gets a null: an animation that never plays, a sprite that never draws, or a
`NullReferenceException` from a callback that never checked. Nothing in CI notices, because the only
reader of the annotation is a drawer, and a drawer needs somebody looking at the inspector.

The same holds for any sibling "author must fill this" attribute a project adds -- the failure is
structural, not specific to `WNotNull`.

## Recommended implementation

An EditMode gate in `Tests/Editor/Validation`, plus the reusable half in `Editor/Validation` so a
consuming project can run it from a menu item too.

- **Derive the field set from the annotations, never from a hand-listed set of fields.** A list has
  to be updated by the change that adds a field, and forgetting makes the gate pass -- which is the
  property that makes the gate worth having in the first place.
- **Discovery is `TypeCache`.** It is Unity's own index and cannot miss a formatting variant.
- **Read the assets as text**, not by loading them. Opening a scene runs every `OnValidate` in it, so
  the act of inspecting dirties it, and closing then prompts for a save -- which blocks the editor's
  main loop. A gate must not mutate what it measures.
- **"Unfilled" must be the drawer's own answer, not a second one.** Whatever predicate renders the
  inspector warning is the definition: a reference is null when unassigned, a string when blank, a
  collection when empty. Restating it gives the build and the inspector two definitions of the same
  word.
- Resolve each asset's document to a type through its `m_Script` guid, and report `path:line: field`.

## The one number to expose

A type with no `MonoScript` -- an inline-only `[Serializable]` class, or a component no committed
asset carries -- cannot be located in any document, so its annotations are unreadable. That
population is real (31 fields in the project measured) and should be **reported as a budget** rather
than silently skipped, so it can only shrink. A gate that quietly cannot see part of its subject is
the failure mode this is guarding against.

## Prior art

Implemented project-locally as `AuthoredRequirementTests`, with the reasoning above recorded in its
`<remarks>`: <https://github.com/Ambiguous-Interactive/IshoBoy>

### LINKED ISSUE #619: [Feature]: an authored-asset gate for SerializableDictionary, for the assets #548 left behind and the null value it cannot prevent
#314 / #348 / #548 fixed the **type**: `SerializableDictionary<TKey, List<TValue>>` no longer drops
its values. Nothing looks at the **assets that were authored while it did**, or at the neighbouring
hole the type cannot prevent.

## The two states an authored map can be in

**1. Keys written, values absent.** An asset serialized before the fix records what the mapping is
*about* and nothing about what it maps to. The asset still loads, the dictionary is empty, and every
lookup misses. In the project measured, four declarations were in that state and one -- a set of
decal mappings -- had been dead since a serializer migration, with nobody noticing because "no decal
appeared" is not an error.

**2. A key whose value is a null object reference.** This one the type cannot fix, because the asset
is well-formed: an unassigned inspector slot stores a real key beside `{fileID: 0}`, so `TryGetValue`
returns **`true`** with a null. Every caller then has to remember a second check, and a designer who
adds an `"Idle"` key and forgets the clip gets an entry that looks correct and does nothing.

## Use case

A consuming project upgrades the package, gets the fix, and still has the broken assets. There is no
signal that anything is wrong -- the compile is clean and the tests pass, because a test that
populates a dictionary in code never touches the authored one.

## Recommended implementation

A text-level EditMode gate over committed `.asset` / `.prefab` / `.unity`, in `Tests/Editor` with the
reusable half in `Editor/Validation`:

- **`_keys` without `_values`** -- the shape, not the data. A correctly typed empty dictionary writes
  `_keys: []` and `_values: []`, so a mis-shaped one fails the moment anything serializes it, before a
  designer has authored the entry that would be lost.
- **A `_values` entry that is `{fileID: 0}`** beside a real key.
- **Text, not `SerializedObject`.** Loading the asset asks the same serializer that dropped the values
  what it thinks they are, and it answers "there is no such field" -- indistinguishable from an empty
  dictionary. Text also reaches baked scenes without opening one.
- **A vacuity assertion beside it**: fail if the scan found no `_keys:` block at all, because a moved
  root or a renamed backing field otherwise turns the gate green in silence.

## The trap worth encoding, learned the hard way

The same gate in the project measured also checked the **keyed-list** shape -- a `List<Entry>` where
each entry carries its own `_name` -- and it matched entries beginning `- _name:`. **That makes the
subject set a function of which field the entry type declares first.** One config type declared
`_name` first and was judged; a sibling with the same shape declared `_disabled` first and was never
looked at, ten entries in the same asset. Match a `_name` field **anywhere** in the entry block.

And an entry may hold a reference that is optional by design. Those need a declared, printed
exemption -- not silence, and not field order deciding it by accident.

### LINKED ISSUE #620: [Feature]: nothing checks that an AnimationClip's keyframes still resolve, which is #203's defect one layer out
An `AnimationClip` stores each animated object reference as a keyframe. Delete the object and the
keyframe stays, pointing at nothing. **Unity reports that nowhere**: the clip imports, the animator
plays it, and the renderer is handed `null` for that frame's duration -- so the subject vanishes
part-way through its own animation and comes back.

## Why this package in particular

`AnimationCreatorWindow` is a clip **producer**, and #203 was exactly this defect arriving from it:
*"`CreateAnimationClip()` injects null-sprite keyframes via oversized `ArrayPool` buffer"*. That was
fixed at the source. Nothing checks the result, so the next producer bug -- or an ordinary sprite
deletion, or a texture whose `spriteMode` changes -- lands silently in the same way.

## The measurement that makes the case

A guid scan of a project's clips found 34 with every frame empty. The same clips through
`AnimationUtility` found **41**. The seven extra are the point:

Their guid resolves perfectly. `Cropped_cloud1.png` exists and its `.meta` still lists a sprite with
exactly the internal id the keyframe names. What does not exist is the **sub-asset**: the texture's
`spriteMode` is `1` (Single), so the importer produces one sprite at `21300000` and none of the
slices the meta still describes. **Only the AssetDatabase can tell "the guid resolves" from "the
object resolves".** A text scan cannot, and neither can a reader eyeballing the meta.

## Use case

Any project that slices a sprite sheet, animates from it, then re-imports the texture as Single --
or that deletes one frame's art. The symptom is "the thing flickers", which nobody files as a bug
against animation.

## Recommended implementation

An EditMode gate, with the reusable half in `Editor/Validation`:

```csharp
foreach (EditorCurveBinding binding in AnimationUtility.GetObjectReferenceCurveBindings(clip))
{
    ObjectReferenceKeyframe[] keyframes = AnimationUtility.GetObjectReferenceCurve(clip, binding);
    // report every keyframe whose .value is null, naming binding.propertyName, binding.path and .time
}
```

- Scope to the consuming project's own trees; a vendored clip under a third-party folder is somebody
  else's to fix.
- **Assert a non-zero clip count and a non-zero keyframe count**, or the gate passes vacuously the
  day the scan path stops matching -- and a clip full of empty frames would still pass.
- `GetObjectReferenceCurve` can return `null`; skip rather than dereference.

## Not the same as the dead-clip question

A clip nothing references is a different issue and needs a different instrument: a guid census is an
**upper bound** on what is dead, because a clip loaded by `AssetDatabase.LoadAssetAtPath` against a
directory is named by no guid anywhere. This gate is only about clips that cannot work however they
are reached.

### LINKED ISSUE #624: [Feature]: a type with no MonoScript cannot be authored onto anything, and Unity picks the class by file name in silence
A `MonoBehaviour` or `ScriptableObject` reaches a scene, a prefab or an asset through exactly one
door: the `MonoScript` Unity builds for the file that declares it. A serialized `m_Script` is a
reference to that `MonoScript`.

**Unity picks the class for a file by name and falls back silently when nothing matches**, so the
damage depends entirely on how many types happen to sit above the one you care about.

## The two outcomes, measured

- `PlayerOutlineGenerator.cs` declared one top-level type, `PlayerOcclusionAwareOutline`. Its
  `MonoScript` resolved to that type anyway, because it was the only candidate. **Harmless, and one
  edit from not being.**
- `FPSTracker.cs` declared two -- a message struct first, then `FpsTracker`. The `MonoScript` resolved
  to the **struct**, leaving `FpsTracker` with no door at all.

Neither file matched its type by name. The difference between harmless and fatal was ordering.

## Why it survives every other check

A type with no `MonoScript` still compiles, and `AddComponent<T>()` still constructs it at runtime --
so every behavioural test passes. The gap shows up only when somebody tries to **author** the thing,
and then it looks like "the component won't drag onto the prefab", which reads as a Unity glitch.

The first file above was on a shipped player prefab: one added type above it and every reference to
it becomes a missing script, across every scene.

## Recommended implementation

Two rules, in `Tests/Editor/Validation` with the reusable half in `Editor/Validation`. Neither is
redundant:

1. **Every concrete `MonoBehaviour` / `ScriptableObject` type resolves to a `MonoScript`.** The
   symptom.
2. **Every script asset is named after the type it binds.** The cause, and what keeps rule 1 true --
   a file named after its type cannot lose the binding to a type added above it.

Notes that cost time to rediscover:

- **Discovery is `TypeCache`**, Unity's own index. A source scan has been measured missing instances
  in a real tree.
- **Do not exclude nested or open-generic component types.** Neither can carry a `MonoScript` either,
  so an exclusion is a licence to introduce the same defect in a shape the gate stops looking at.
  Abstract types are excluded, because nothing can be an instance of one.
- Together the two settle "one file per MonoBehaviour/ScriptableObject" **without a source scan**: one
  file yields one `MonoScript` bound to one class, so a second component type sharing a file can
  satisfy at most one of the two rules. That covers test assemblies too, where a helper component
  declared inside a test file is the usual offender.
- Rule 2 is also load-bearing for **any other gate that resolves a type to its declaring file**. A
  type in a file not named after it resolves to nothing, an "except its own file" exclusion then
  matches nothing, and the type's own declaration answers as a reference to itself -- reachable,
  silently, and for every type at once if resolution ever breaks wholesale.

### LINKED ISSUE #628: [Feature]: analyzer for dereferencing a serialized list row, which a deleted asset empties with nobody editing
A serialized `List<SomeAsset>` holds **references**. Delete or rename the asset a row names and Unity
leaves the row behind, **empty**. So "the list is authored correctly" is not a property any authoring
step preserves, and no editing session has to have happened for it to stop being true.

Every walk of such a list that dereferences a row without asking will throw, at the worst possible
place: inside `OnEnable`, `Awake` or `Start`, where the throw lands **before** everything after it in
the same method.

## What it costs, from real instances

Five sites in one project. The shape of the damage is the point:

- A glyph configuration walked the keyboard set **first**, so one deleted keycap asset took **every
  gamepad glyph table** with it -- the throw landed before the other four `Load` calls.
- A level generator guarded the **list** (`if (_postProcessors != null)`) and not the **rows**, so the
  throw landed before the scene was saved and the bake stopped half-way.
- A path follower threw **every frame** for the rest of the level.

## The asymmetry worth encoding

In two of the five the author had already thought about null -- in the seam they wrote *by hand*, four
lines from the offending loop. **The guarded seam is the one somebody wrote. The unguarded one is the
serialized field, which nobody thinks of as an input.** That is why the rule is not "check for null in
`OnEnable`": it is that a serialized collection of references is untrusted data.

## Use case

Any project where art or data assets are reorganised -- which is all of them. The row also empties
when a *component* is taken off a prefab, if the list names components rather than assets, and that is
a far more ordinary edit than deleting a file.

## Recommended implementation

**`WUNITY030`** -- a dereference of an element of a serialized `List<T>` / `T[]` where `T` is a
`UnityEngine.Object`, without a null test on that element.

Two design notes that came out of building this as a source linter:

- **Resolve the class hierarchy; do not keep a list of Unity base types.** A first draft carried its
  own 22-entry root list and missed the level-generator instance entirely. Sharing the resolver with
  the Unity-null diagnostic found it immediately -- and, more importantly, **two hierarchies would
  drift, and the one that drifted would be the one reporting clean**.
- **Compaction must count as the guard.** Where a list is walked more than once, the right repair is
  to drop null rows once up front rather than test each row forever. A rule that only accepts a
  per-loop `if (row == null)` forces the *worse* of the two repairs, and a rule that makes the code
  worse in order to be satisfied is a rule people route around.

Red-green against the pre-fix sources gave 6 findings across 4 files; 0 across 652 files after.

### LINKED ISSUE #631: [Feature]: a shared reader for committed Unity YAML, which four proposed checks would otherwise each write
Four of the checks proposed in #617, #618, #619 and #624 all begin the same way: **read a committed
`.unity` / `.prefab` / `.asset` as the sequence of YAML documents Unity wrote, and resolve each
document's `m_Script` to a type.** Written per-check, that is four copies of a parser and four chances
for one of them to drift into reporting clean.

## Why text, and why it is not a shortcut

Loading the asset is the wrong instrument for all four, for reasons that cost real time to learn:

- **Opening a scene mutates it.** Every `OnValidate` in it runs -- a collider that rebuilds its points
  from the sprite's physics shape is enough -- so the act of inspecting dirties the scene, and closing
  it then prompts for a save. A prompt blocks the editor's main loop, which is fatal for anything
  driving the editor over a bridge. **A gate must not mutate what it measures.**
- **Text reaches a baked scene without opening one**, and needs no editor at all for the parsing half.
- For #617 specifically, loading asks the same serializer that dropped the data what it thinks the
  data is, and it answers "there is no such field" -- indistinguishable from "the field is empty".

## Recommended implementation

`Editor/Validation/AuthoredAssetYaml` (or `Editor/Utils`), with no dependency on the checks above:

- `IEnumerable<string> EnumerateAuthoredAssets(params string[] extensions)` over a root, **re-checking
  the extension after the glob** -- Windows matches a search pattern against the 8.3 short name too,
  so `*.unity` can return a `.unityproj`.
- `IReadOnlyDictionary<string, YamlDocument> ReadDocuments(string[] lines)` -- documents keyed by
  fileID, each with its `--- !u!<classId>` type, its line range, and its top-level keys with values.
- `string ScriptGuidOf(Type type)` and the inverse, resolving a document's `m_Script` guid through
  `AssetDatabase.GUIDToAssetPath` -> `MonoScript.GetClass()`.
- Helpers for the two values every one of these checks compares against: the null reference literal
  `{fileID: 0}`, and the empty sequence `[]`.

## The one correctness note

`ScriptGuidOf` is usually written as a **name-narrowed** search -- find the `.cs` named after the type,
confirm with `GetClass()`. That is fast and correct *while* script assets are named after their types,
which is #624's second rule. When it is not, the type resolves to nothing, any "except its own file"
exclusion matches nothing, and a check can pass for every type at once in silence.

So this reader and #624 hold each other up, and whichever ships second should say so where a reader
will find it.

wallstop:
**Why:** The package checks code thoroughly and authored data not at all, so an asset that stopped being correct without anybody editing it says nothing.

**What:**

- One reader for committed `.unity`/`.prefab`/`.asset`, so five checks share a parser instead of writing five.
- Reports for a type nothing can author, an unfilled `[WNotNull]` slot, a dictionary that lost its pairing, and a keyframe that resolves to nothing.
- Stale serialized keys reported by cause, with a repair that undoes any rewrite losing sub-objects.
- `WUH012` for a serialized row dereferenced without a test; it found a real one in `AnimationCreatorWindow`.
- Two drifted `PendingValueWrapper` copies merged, and two files renamed after what they bind.

Fixes #631
Fixes #624
Fixes #618
Fixes #619
Fixes #620
Fixes #617
Fixes #628

---

> [!NOTE]
> **Medium Risk**
> New menu-driven asset repair can rewrite selected files (with rollback on content loss), and YAML-based scans depend on correct path resolution; analyzer additions are mostly opt-in warnings except extended WUH003.
> 
> **Overview**
> Adds **read-only authored-asset validation** that parses committed `.unity`/`.prefab`/`.asset` YAML without opening scenes, plus **Tools > Wallstop Studios > Unity Helpers > Authored Assets** reports for script binding problems, empty `[WNotNull]` slots, broken `SerializableDictionary` pairings, and animation object keyframes that resolve to nothing.
> 
> **Stale serialized keys** get report-by-cause menus and an optional **selection-only repair** that reserializes one asset at a time and restores the file if sub-object count drops.
> 
> **Analyzers:** **`WUH003`** now flags `is null` / `is not null` on `UnityEngine.Object`; new **`WUH012`** warns on dereferencing serialized list/array rows without a guard (drove a null-sprite skip in `AnimationCreatorWindow`); **`WUH013`** suggests `foreach` for simple counting loops on concrete sequences (opt-in like `WUH010`).
> 
> **Supporting fixes:** shared **`PendingValueWrapper`** for dictionary/set drawers; **`AuthoredAssetPaths`** for project-root file reads; new **honest-gates** contributor guidance and related doc/changelog updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04c236776b6e65f9189bdf007c3fe4b389bca58a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/CustomDrawers/PendingValueWrapper.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        /// <summary>The serialized key the pending value is written under.</summary>
+        internal const string PropertyName = nameof(boxedValue);
+
+        [SerializeReference]
+        private object boxedValue;
+
+        /// <summary>Reads the value being edited.</summary>
+        /// <returns>The pending value, which may be <c>null</c>.</returns>
+        public object GetValue()
```
wallstop, line 32: Is this ok? Raw object? Does this play nicely with Unity values and types and such?

Should we do something like a clean, TryGet with a generic interface/typecheck? Or is this fine?

Editor/Validation/AnimationClipKeyframeValidator.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            ClipPath = clipPath;
+            ClipName = clipName;
+            BindingPath = bindingPath;
+            PropertyName = propertyName;
+            Time = time;
+        }
+
+        /// <summary>The clip's asset path.</summary>
+        public string ClipPath { get; }
```
wallstop: properties should be above constructors, I believe we have an editor policy/clean ruling for this, why isn't it applied throughout?

Editor/Validation/AnimationClipKeyframeValidator.cs
```diff
@@ -0,0 +1,217 @@
...
+        )
+        {
+            ClipPath = clipPath;
+            ClipName = clipName;
+            BindingPath = bindingPath;
+            PropertyName = propertyName;
+            Time = time;
+        }
+
+        /// <summary>The clip's asset path.</summary>
```
wallstop: Why do we have all of these verbose comments? I thought we had a very clean, "no verbose comments" policy? What happened to that, why aren't we following it?

Editor/Validation/AnimationClipKeyframeValidator.cs
```diff
@@ -0,0 +1,217 @@
...
+    /// </para>
+    /// <para>
+    /// The instrument is <c>AnimationUtility</c> rather than a guid scan, and the difference is not
+    /// a refinement. A keyframe's guid can resolve perfectly while the object does not: a sprite
+    /// sheet re-imported as <c>Single</c> still has a <c>.meta</c> describing every slice, and the
+    /// importer produces none of them. Only the AssetDatabase can tell "the guid resolves" from
+    /// "the object resolves".
+    /// </para>
+    /// </remarks>
+    public static class AnimationClipKeyframeValidator
```
wallstop, line 20: Can we follow a rule of one-classs-per-file?

Editor/Validation/AnimationClipKeyframeValidator.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// <param name="findings">Receives one entry per empty keyframe.</param>
+        /// <param name="clipsInspected">Receives how many clips were opened.</param>
+        /// <param name="keyframesInspected">Receives how many object keyframes were judged.</param>
+        /// <returns><c>false</c> when the scan could not run at all.</returns>
+        /// <remarks>
+        /// Both counts are outputs so a caller can refuse a vacuous pass: a scan whose path scope
+        /// stops matching reports zero findings, and a clip whose every frame is empty would pass
+        /// beside it.
+        /// </remarks>
+        public static bool TryScan(
```
wallstop, line 31: Is this good? Should this be a static class? Or an instanced class, for testing?

Editor/Validation/AnimationClipKeyframeValidator.cs
```diff
@@ -0,0 +1,217 @@
...
+                    continue;
+                }
+
+                Object[] assets = AssetDatabase.LoadAllAssetsAtPath(assetPath);
+                if (assets == null)
+                {
+                    continue;
+                }
+
+                for (int asset = 0; asset < assets.Length; ++asset)
```
wallstop: Why are we doing counting for loops? Why not foreach? We should utilize foreach wherever we can verify that it won't generate garbage. Counting for loops should be reserved for IReadonlyList (allocates in foreach) OR when we need the index for code within the loop.

Editor/Validation/AnimationClipKeyframeValidator.cs
```diff
@@ -0,0 +1,217 @@
...
+
+                    ++clipsInspected;
+                    keyframesInspected += Inspect(assetPath, clip, findings);
+                }
+            }
+
+            return true;
+        }
+
+        private static int Inspect(
```
wallstop: Can these be extracted to internal/helper/testable methods?

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: **Sweep.** Properties now sit above the constructor in every type this PR adds: `AuthoredAssetEntry`, `AuthoredAssetDocument`, `AuthoredSequenceElement`, and the five finding/exemption structs that moved to their own files.

On "why isn't it applied throughout" -- I copied the prevailing shape rather than a rule. `DroppedSerializedField` in `SerializedFieldValidator.cs`, right next door, puts its constructor first, and that is what I matched. **I could not find the policy written down anywhere**: nothing in `.editorconfig`, nothing in `.llm/skills/`, and no analyzer for it, so the repository currently has two conventions and no way to tell which is intended.

Which makes this the third bullet, and I'd rather you decide it than guess: the ordering is worth a **rule** so the next file does not get it wrong the same way. The cheapest honest options are an `.editorconfig` `dotnet_diagnostic.SA1201` entry (StyleCop ordering, which is exactly this), or a line in `create-csharp-file.md` plus a small linter. Say which and I will land it and sweep the existing files in the same pass; I have not touched files outside this PR yet, because a repository-wide member reorder is a large diff to bury in a feature PR.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: **Sweep, and you are right.** I wrote essays. The policy is real -- `.llm/context.md` rule 8 says aim for zero comments, and `update-documentation.md` says `<summary>` is one sentence and `<remarks>` is "when needed" -- and I treated `<remarks>` as a place to narrate every measurement instead.

What changed across all sixteen new files:

- Every type-level `<remarks>` is now **at most four lines**, keeping only the one thing that stops a reader undoing the code -- "a collection-valued dictionary stores its values in `_boxedValues`, so no `_values` is not by itself a defect" -- and linking to the guide for the rest.
- Every member-level `<remarks>` over six lines is **deleted**. 54 lines of prose went.
- `<summary>`, `<param>` and `<returns>` stay, because those are the structured API docs `update-documentation.md` requires, not commentary.

The reasoning did not vanish, it moved to where prose belongs: [`docs/features/editor-tools/authored-asset-validation.md`](https://github.com/Ambiguous-Interactive/unity-helpers/blob/main/docs/features/editor-tools/authored-asset-validation.md), which this PR adds and every trimmed remark now links to.

Worth noting as a **rule** question, since it is [#635](https://github.com/Ambiguous-Interactive/unity-helpers/issues/635)'s territory: nothing enforces this. The comment audit measured 2,226 runs of two-or-more consecutive `//` lines and is "enforced by nothing", and XML `<remarks>` is not even in that count -- so a verbose-remarks linter would be a second rule, not a bigger version of the first. I have not added one here; if you want it, #635 is the place and I would rather do it as its own change than smuggle a repository-wide comment sweep into this PR.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: **Sweep.** Every type this PR adds is now alone in a file named after it. Six types moved out:

| Moved out of | Into |
| --- | --- |
| `AnimationClipKeyframeValidator.cs` | `AnimationKeyframeFinding.cs` |
| `MonoScriptBindingValidator.cs` | `MonoScriptBindingFinding.cs` |
| `AuthoredRequirementValidator.cs` | `AuthoredRequirementFinding.cs`, `AuthoredRequirementExemption.cs`, `AuthoredRequirementField.cs` (was a private nested struct) |
| `SerializableDictionaryAssetValidator.cs` | `SerializableDictionaryAssetFinding.cs` |
| `StaleSerializedKeyValidator.cs` | `StaleSerializedKeyFinding.cs` |

Same reason as the ordering thread: I matched `SerializedFieldValidator.cs`, which declares `DroppedSerializedField` beside the validator. So the existing file is now the odd one out, not the precedent.

The rule half is more than style here, and it is the same rule this PR is about. `.llm/context.md` rule 6 already says one file per `MonoBehaviour`/`ScriptableObject`, and rule 2 of the `MonoScript` binding check added in this PR is exactly "every script asset is named after the type it binds" -- which is only enforceable for types that can carry a `MonoScript`. A plain `readonly struct` carries none, so no gate catches it. Extending rule 6 from "one component or asset per file" to "one type per file" would make the convention uniform and checkable by a source linter rather than by review. Happy to open that as an issue if you want it; I did not want to assume the broader rule from one comment.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: **Sweep, applying the rule exactly as you stated it** -- `foreach` unless the collection is `IReadOnlyList` (whose enumerator boxes) or the body needs the index.

Converted to `foreach` (arrays and one concrete `List<T>`):

- `MonoScriptIndex`: the two `string[]` walks from `AssetDatabase.FindAssets`
- `MonoScriptBindingValidator`: `Assembly[]`, its `sourceFiles` `string[]`, and the `FindAssets` `string[]`
- `AuthoredRequirementValidator`: the `object[]` of attributes, the `List<string>` of carrier guids, and the `List<AuthoredRequirementField>` of required fields
- `StaleSerializedKeyValidator`: the `FieldInfo[]` and the `object[]` of aliases
- `StaleSerializedKeyRepair`: the `Object[]` from `LoadAllAssetsAtPath`
- `AnimationClipKeyframeValidator`: the guid, asset, binding and keyframe arrays -- 4 of the 5 loops in the file you commented on

Left as counting `for`, each for one of your two reasons:

- **`IReadOnlyList` parameters**: `assetPaths`, `documents`, `Entries`, `assetPathPrefixes`, `findings`, `field.Aliases`. These are the public API's parameter type, so the boxing is the thing being avoided.
- **The index is used**: `TryFindSibling` walks outward from an anchor index in both directions; `ReadDocument` and `FindEntryEnd` look ahead in `lines`; `TrySplitEntry` and `IsZeroGuid` index a `string`; `EnumerateSequenceElements` needs the line number to report.

One thing worth pinning as a **rule** rather than leaving to review: the `IReadOnlyList` exception is measurable but invisible at the call site -- `foreach` over `IReadOnlyList<T>` boxes the enumerator, `foreach` over the identical `List<T>` does not, and nothing in the source distinguishes them. That is analyzer-shaped, the same way `WUH001` catches a method-group conversion. Say the word and I will open it.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: **Fixed, and swept to the sibling validators.** `Editor/AssemblyInfo.cs` already grants `InternalsVisibleTo` to `WallstopStudios.UnityHelpers.Tests.Editor.Validation`, so this cost nothing but the decision.

Now `internal` and **tested one at a time** in the new `AuthoredAssetHelperTests`:

| Helper | What the test pins |
| --- | --- |
| `AnimationClipKeyframeValidator.Inspect` | one clip's keyframes, separable from the project scan |
| `AnimationClipKeyframeValidator.IsInScope` / `MonoScriptBindingValidator.IsUnderAnyPrefix` | case-insensitivity, backslash normalization, null and empty prefix lists |
| `AuthoredAssetYaml.TrySplitEntry` | the five spellings that are entries and the five that are not, including `{fileID: 0}` and a quoted value containing a colon |
| `AuthoredAssetYaml.LeadingSpaces` | that a tab is not indentation |
| `MonoScriptBindingValidator.SimpleNameOf` | generic arity stripping |
| `MonoScriptBindingValidator.IsAuthorable` | that an inspector is excluded and a `ScriptableObject` is not |
| `AuthoredRequirementValidator.TryClassify` / `IsEmptyValue` | reference vs string vs collection, and that a number is refused |
| `SerializableDictionaryAssetValidator.TryFindSibling` / `TryCountElements` | reachable directly rather than only through a whole-asset scan |
| `StaleSerializedKeyValidator.BuildDeclaredKeys` | the probe's key set, without needing an asset |

That is the honest version of "extract for testability" -- exposing a seam and not testing it would just be a wider surface. `IsAuthorable` in particular is worth having pinned: it encodes a judgement call (an `EditorWindow` is in scope because a saved layout names its script; an `Editor` is not because `[CustomEditor]` finds it), and that is the kind of thing that gets silently widened later.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: Deliberately static, and I think correctly -- but the testability half of your question was a real gap, and that is now fixed rather than argued away.

**Why static holds.** These validators have no state. Every one is `(inputs) -> findings`, and the only thing resembling state is `MonoScriptIndex`'s guid cache, which is deliberately *not* on them -- it is process-wide because the type-to-`MonoScript` map is a property of the project, not of a scan, and it has an explicit `ClearCaches()` that every menu command and every test `SetUp` calls. Instancing the validators would give each one a constructor that stores nothing.

**Why the usual reason to instance does not apply.** Instancing normally buys a seam to substitute. The thing worth substituting here is Unity itself -- `AssetDatabase`, `TypeCache`, `AnimationUtility` -- and an interface over those is a large abstraction whose fake would be a second, unverified model of Unity's behaviour. That is exactly the failure `SerializedFieldValidator` already documents: modelling Unity's serialization rules instead of asking Unity was measured reporting ~1771 findings of which about four were real. I would rather the tests pay for a real editor, which they now do -- these fixtures ran green on 2022.3, 6000.3 and 6000.5 in this PR's matrix.

**What was actually missing.** The parts that need no editor were locked inside `private static`, so a test could only reach them through a whole-asset scan. Those are `internal` now and tested individually -- entry splitting, indentation, path scoping, generic arity, authorability, field classification, sibling lookup, element counting, the probe's key set. That gets the testability you are asking about without inventing a Unity double.

If you would still rather have an instance seam, the shape I would pick is not instancing the validators but a single `IAuthoredAssetSource` over the four AssetDatabase calls they make -- one small interface, one fake, four methods. Say so and I will open it as an issue; I did not want to add an abstraction on a guess.

Editor/CustomDrawers/PendingValueWrapper.cs
wallstop: Fair question, and I should be clear about what this PR did and did not do here: **I did not write this type, I deduplicated it.** It existed twice, nested in `SerializableDictionaryPropertyDrawer` and again in `SerializableSetPropertyDrawer`, and the two copies had already drifted -- one named its property with `nameof`, the other with the literal `"boxedValue"`. It moved to its own file because a nested `ScriptableObject` carries no `MonoScript`, which is the shape rule 1 of this PR's binding check reports. The `object` field came along unchanged.

**On whether `object` is right.** I think it is, and a generic cannot replace it. The only caller shape is:

```csharp
PendingValueWrapper wrapper = ScriptableObject.CreateInstance<PendingValueWrapper>();
wrapper.SetValue(CloneComplexValue(pendingValue, valueType));
return SerializedProperty.DataEquals(existingValueProperty, wrapper.FindValueProperty(serialized));
```

The point is to give an arbitrary pending dictionary value a `SerializedProperty` so Unity's own `DataEquals` can compare it against the authored one. `valueType` is the dictionary's `TValue`, **decided at runtime**, and `ScriptableObject.CreateInstance<T>` needs a closed type at compile time -- so `PendingValueWrapper<T>` cannot be constructed here, and Unity will not create a `ScriptableObject` of an open generic anyway. A `TryGet<T>` accessor would type the *read* but not the *field*, which is the part that has to be `object` for `[SerializeReference]` to serialize the dynamic type.

**Where I would not defend it, and what I checked.** `[SerializeReference]` accepts managed references only: it does not serialize a `UnityEngine.Object` reference, and Unity refuses primitives. So this path is correct only for the complex values `CloneComplexValue` is named for, and **I did not verify the guard that keeps primitives away from it** -- the caller sits behind a `usedSerializedComparison` branch I did not trace. That is pre-existing behaviour either way; my change cannot have altered it, because the field, the attribute and the two methods are byte-identical to the copy that was in the dictionary drawer.

So: no behaviour change intended or made, and I would rather not redesign a drawer internal inside an authored-asset-validation PR. If you want the primitive path pinned, I will open an issue to trace the guard and add a fixture -- that is a real gap, just not one this PR opened.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: Style should be:

Nested Types
const
static properties
static fields
properties (public -> protected -> internal -> private)
fields (public -> protected -> internal -> private)
constructors (public -> protected -> internal -> private)
static methods (public -> protected -> internal -> private)
methods (public -> protected -> internal -> private)

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: <summary> is also only when needed.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: Please pin it/make it an analyzer/warning/opt-out-able

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: Follow-up: **this comment found a real bug**, within one CI run of me acting on it.

Extracting the seams meant I could finally assert that the two path-scope helpers agreed. They did not. `AnimationClipKeyframeValidator.IsInScope` -- which I wrote fresh during the sweep -- returned `false` for a null path. `MonoScriptBindingValidator.IsUnderAnyPrefix`, which predates it, dereferenced it:

```text
AuthoredAssetHelperTests.APathIsInScopeOnlyUnderOneOfThePrefixes(null,False)
  -- System.NullReferenceException
AuthoredAssetHelperTests.APathWithNoPrefixesAtAllIsOutOfScope
  -- System.NullReferenceException
```

Identical on `6000.5.2f1 editmode` and `6000.3.16f1 editmode`. It is a genuine violation of the defensive-programming rule ("handle all inputs gracefully"), and nothing else would have caught it: both helpers are only ever called with paths that `AssetDatabase` just handed back, so the throw is unreachable from the validators' own code paths and reachable from any other caller.

**The repair is the class, not the instance.** Adding a null check to the copy that happened to be reported would have left two helpers answering one question two ways -- which is the actual defect. There is now one `AuthoredAssetYaml.IsUnderAnyPrefix`, both validators delegate to it, and it has a single set of tests covering null path, null prefixes, an empty prefix entry, case-insensitivity and backslash normalization. Ten inputs, verified against the real source outside Unity, the two that failed included.

So: **fix**, then **sweep** (one helper instead of two), and the rule half is already the one you raised -- exposing a seam for test is not paperwork, it is the thing that made a divergence visible.

Generator~/WallstopStudios.UnityHelpers.Analyzers/SerializedRowDereferenceAnalyzer.cs
```diff
@@ -0,0 +1,520 @@
...
+                )
+                {
+                    return true;
+                }
+            }
+
+            if (operation is IIsPatternOperation pattern && IsRowReference(pattern.Value, row))
+            {
+                return true;
+            }
```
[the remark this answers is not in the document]
wallstop: **Fixed, and the defect is worse than described** -- which makes this the best catch on the PR.

You are right that `IsNullTestOf` accepted any `is` pattern. The reason that matters here is sharper than the general false-negative argument: **a type pattern is not a null test for a `UnityEngine.Object` at all.** A destroyed object keeps its managed wrapper, so `row is Sprite` is `true` for an object whose native half is gone, and the very next `row.Load()` throws `MissingReferenceException`. This repository already has the rule -- `.llm/context.md`: "`is not null` is NOT a substitute" for `!=` -- and `WUH003` exists to report exactly that confusion. So my code was silencing `WUH012` on precisely the row it exists for.

`IsNullPattern` now accepts only a constant-null pattern, unwrapping negation so `is not null` still counts, and rejects everything else:

```csharp
while (pattern is INegatedPatternOperation negated) { pattern = negated.Pattern; }
return pattern is IConstantPatternOperation constant
    && constant.Value != null
    && constant.Value.ConstantValue.HasValue
    && constant.Value.ConstantValue.Value == null;
```

Five new fixtures, red-green against the previous build -- all three of these reported nothing before and report now:

| Guard | Before | Now |
| --- | --- | --- |
| `if (keycap is UnityEngine.ScriptableObject) { keycap.Load(); }` | silent | **reported** |
| `if (keycap is not UnityEngine.ScriptableObject) { } keycap.Load();` | silent | **reported** |
| `if (keycap is { }) { keycap.Load(); }` | silent | **reported** |
| `if (keycap is null) { continue; }` | clean | clean |
| `if (keycap is not null) { keycap.Load(); }` | clean | clean |

`is null` is deliberately still accepted even though it is the wrong operator for a Unity object: this rule asks whether the author treated the row as an input, and `WUH003` is the rule that tells them `is null` misses a destroyed object. Two rules, one report each, rather than both firing on one line.

22 analyzer fixtures pass. Thanks -- this one was a genuine hole.

Editor/Validation/MonoScriptBindingValidator.cs
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        private static HashSet<string> ScopedAssemblyNames(IReadOnlyList<string> assetPathPrefixes)
+        {
+            HashSet<string> names = new(StringComparer.Ordinal);
+            Assembly[] assemblies = CompilationPipeline.GetAssemblies(AssembliesType.Editor);
+            if (assemblies == null)
+            {
+                return names;
+            }
```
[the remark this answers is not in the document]
wallstop: **Fixed, both halves.** The risk you name is the one this check is least able to notice about itself: a scope that quietly loses a tree reports zero findings, and zero findings is exactly what a passing scan reports.

**Production.** `ScopedAssemblyNames` now unions `AssembliesType.Editor` with `AssembliesType.Player` and dedupes. An assembly excluded from one set is absent from it entirely, so asking both is the cheap way to stop a platform-restricted asmdef falling out of scope. Widening is safe in the other direction too: discovery is `TypeCache`, so an assembly not loaded in the editor contributes no types and therefore no findings -- the union can only add subjects, never invent them.

**Test.** The guard I had -- `0 < typesConsidered` over `Runtime/` and `Editor/` together -- catches total vacuity and *not* what you described, because the combined count stays non-zero on the strength of whichever tree survived. There is now a per-tree control:

```csharp
Assert.IsTrue(0 < runtimeTypes,   "Runtime/ fell out of the assembly scope.");
Assert.IsTrue(0 < runtimeScripts, "Runtime/ fell out of the script scope.");
Assert.IsTrue(0 < editorTypes,    "Editor/ fell out of the assembly scope.");
Assert.IsTrue(0 < editorScripts,  "Editor/ fell out of the script scope.");
```

Scanning each tree alone means losing either one fails the run rather than being absorbed. That is the assertion that would actually have caught the scenario in your comment.

Worth noting the shape, since it recurs across this PR: every one of these checks returns its subject counts as outputs precisely so a caller can refuse a vacuous pass -- and I had still written the weaker of the two available assertions. Good catch.

Editor/Validation/AuthoredAssetYaml.cs
```diff
@@ -0,0 +1,600 @@
...
+                lines = Array.Empty<string>();
+                documents = Array.Empty<AuthoredAssetDocument>();
+                return false;
+            }
+
+            string[] read;
+            try
+            {
+                read = File.ReadAllLines(assetPath);
+            }
```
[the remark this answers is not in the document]
wallstop: **Fixed, and the second half of your sentence is why it mattered.** The path resolution is the cause; "scans silently skip files" is the damage, and a silent skip is precisely the vacuous pass this whole PR exists to prevent. I was depending on an unstated invariant -- Unity sets the working directory to the project root -- inside checks whose entire argument is that unstated invariants are how a gate goes quietly green.

New `AuthoredAssetPaths` resolves an asset path against the project root (`Application.dataPath`'s parent), leaves an already-rooted path alone, and returns the input unchanged when the root cannot be determined. Every filesystem entry point now goes through it:

- `AuthoredRequirementValidator`, `SerializableDictionaryAssetValidator` and `StaleSerializedKeyValidator` resolve before `TryReadDocuments` and keep reporting the **original** asset path, exactly as you suggested.
- The two menus enumerate through `AuthoredAssetsUnderProjectRoot()`, which resolves the root for the walk and maps the results back with `ToAssetPath` -- so a project-wide report still prints `Assets/...` a reader can click, rather than an absolute path.

`AuthoredAssetYaml` itself stays a pure filesystem reader with no `UnityEngine` reference, which is deliberate: it is the half I can compile into a standalone console harness and exercise without an editor, and 48 assertions this session depended on that. Putting the Unity-aware resolution one layer out keeps both properties.

Five fixtures, including the round trip your suggestion turns on:

```csharp
Assert.IsTrue(Path.IsPathRooted(ToFileSystemPath("Assets/Anything.asset")));
Assert.AreEqual(once, ToFileSystemPath(once));                      // idempotent
Assert.AreEqual(assetPath, ToAssetPath(ToFileSystemPath(assetPath)));  // round trip
```

plus null/empty inputs and a path outside the project, which is left alone.

The deeper version of your point -- that an unreadable file should be *reported* rather than skipped, whatever the reason -- is still true and not fixed here: it would change the `TryScan` signature on five validators to carry an unreadable-paths list, which is more than this PR should absorb this late. Filing it.

Editor/Validation/StaleSerializedKeyRepair.cs
```diff
@@ -0,0 +1,183 @@
...
+            byte[] original;
+            try
+            {
+                if (!File.Exists(assetPath))
+                {
+                    return StaleSerializedKeyRepairOutcome.RefusedUnreadable;
+                }
+
+                original = File.ReadAllBytes(assetPath);
+            }
```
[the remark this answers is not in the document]
wallstop: **Fixed, both sites.** Same root cause as the reader thread, and you are right that the split matters here specifically: the AssetDatabase wants the **asset** path and the filesystem wants the **resolved** one, and `RepairAsset` was handing the asset path to both.

`RepairAsset` now resolves once and keeps the two apart deliberately:

```csharp
string filePath = AuthoredAssetPaths.ToFileSystemPath(assetPath);
...
if (!File.Exists(filePath)) { return RefusedUnreadable; }
original = File.ReadAllBytes(filePath);
...
AssetDatabase.ForceReserializeAssets(new[] { assetPath }, ReserializeAssetsAndMetadata);
int after = LoadedObjectCount(assetPath);
```

`SameBytes` takes the file path, and `Restore(assetPath, filePath, original)` writes bytes to the file path and re-imports through the asset path -- line 141 in your note, and it is the more dangerous of the two: the restore is the undo half of the sub-object guard, so a write that silently went to the wrong relative location would leave a damaged asset with a refusal logged as though it had been repaired.

Worth being explicit that the previous behaviour was **fail-safe rather than fail-dangerous** -- `File.Exists` returning false made it refuse before touching anything, so it would have declined every asset rather than corrupted one. That does not make it correct: a repair that refuses everything for an invisible reason is the same class of problem as a scan that reports clean for an invisible reason.

This does not change the standing gap on this type: the rewrite path is still exercised only on its refusal branches, tracked in [#667](https://github.com/Ambiguous-Interactive/unity-helpers/issues/667) with the four fixtures that would prove the sub-object guard. The path fix makes those fixtures correct to write; it does not substitute for them.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: Thanks -- that is what I was missing, and I am applying it. **One conflict I will not resolve silently, because a gate enforces the opposite.**

Your order starts with **Nested Types**. `.llm/context.md` rule 6 says a nested type goes at the **END** of its containing type, and `npm run lint:nested-type-placement` enforces it -- currently `1918 file(s) clean`, and it runs in Repo Lint. So "nested types first" would fail CI on every file it touched, and `:fix` would move them back.

Three ways out, and it is your call:

1. **Nested types stay last, the rest of your order applies.** Smallest change, no gate touched. The order becomes: const → static properties → static fields → properties → fields → constructors → static methods → methods → nested types.
2. **Nested types move first**, and `lint-nested-type-placement.js` inverts with it (plus rule 6 and its `:fix`). One coherent rule, one linter change, and a repository-wide reorder of every file that has a nested type.
3. **Nested types stop existing**, which the one-type-per-file thread points at anyway. Then the conflict is moot for new code and the linter keeps holding the legacy cases.

I have applied **everything below Nested Types** to the 16 new files in this PR and left nested types where the linter demands, so the PR is green either way. Say which of the three you want and I will land it -- including the linter inversion and the repo-wide sweep if it is (2).

Two smaller notes on the order as given, so the rule is unambiguous when it is written down:

- **`static properties` before `static fields`, and `properties` before `fields`** is the reverse of the usual C# convention and of what the neighbouring files do. I have followed exactly what you wrote; flagging only so it is deliberate rather than a typo.
- **`const` has no accessibility tier listed** while the others do. I have kept consts in declaration order within the block.

On enforcement, since you asked for the previous rule to be pinned: this one is `.editorconfig` `dotnet_diagnostic.SA1201`/`SA1202` shaped (StyleCop ordering, exactly these two axes), which would hold every file without a bespoke linter. Happy to wire that up once the nested-type question is settled -- it is the piece that decides whether SA1201's built-in order can be used as-is or needs an override.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: **Applied, and the guidance that contradicted you is fixed too** -- which was the actual reason I kept them.

`.llm/context.md` said *"All public members require `<summary>` XML tags"* and `update-documentation.md` said `<summary>` is a "Brief description (one sentence)", i.e. always. So I wrote one on every member and thought I was following the rule. Both now say what you said:

> `<summary>` is for when it is needed, not for every member: it earns its place when the name cannot carry the meaning, and repeating the name back is worse than silence.

**No gate was requiring them** -- `lint:xml-doc-summaries` checks there is not *more than one* summary per block, not that there is one -- so the only thing keeping them was the written rule.

What went, 25 lines of it:

- **`Initializes a new instance of the <see cref="X"/> struct.`** on every constructor. Pure ceremony: it restates the signature directly above it.
- **`<returns>A human-readable description.</returns>`** on every `ToString`.
- Summaries that only restated the member name.

What stayed, and why I did not strip further: `ClipName`'s summary is *"The clip's name, which a sub-asset needs to be identified"* -- the second half is the reason the field exists and the name cannot carry it. Same for `Time` ("in seconds") and `BindingPath` ("the transform path the curve animates"). Those are the case the rule leaves room for; deleting them would be following the letter and losing the point.

`lint:xml-doc-summaries` reports 1932 files clean and the editor trees still compile.

Editor/Validation/AnimationClipKeyframeValidator.cs
wallstop: **Pinned as `WUH013`** -- analyzer, warning, suppressible, and off by default for a measured reason.

It reports a counting `for` over an **array** or **`List<T>`** whose body only ever uses the index to reach into that same sequence. The discriminator is the sequence's *type*, which is exactly why it had to be an analyzer: the loop over `List<T>` and the identical loop over `IReadOnlyList<T>` are the same tokens and opposite answers.

Your three exceptions are the three it does not report, encoded as fixtures:

| Shape | Reported |
| --- | --- |
| `for` over `string[]` / `List<T>`, index only indexes it | **yes** |
| sequence is `IReadOnlyList<T>` / `IList<T>` -- enumerator **boxes** | no |
| body uses the index itself, or indexes a second collection with it | no |
| non-zero start, stride != 1, or backwards | no |
| already a `foreach` | no |

**Off by default, and that is the interesting part.** I turned it on and built all four typecheck trees: **127 sites** across `Runtime/`, `Editor/` and `Tests/`. That is the criterion `.llm/context.md` already sets for an opt-in member -- *"the rule is right and the shape is everywhere"* -- and it is the second member to meet it after `WUH010`. On by default it would bury the eleven correctness rules on a consumer's first build after taking the upgrade.

So it ships off, `Assets/Default.ruleset` turns it on with one line, and [#671](https://github.com/Ambiguous-Interactive/unity-helpers/issues/671) tracks working the 127 down and then adding it to `CheckProjects.ruleset` beside `WUH010` -- which is exactly the path `WUH010` took in #653. An analyzer the package's own code does not obey is documentation with a build step, so I would rather commit to that sequence than switch it on and pragma the difference away.

One implementation note worth recording, because it cost a red test: an **indexer wraps its argument in an `IArgumentOperation`** where an array element reference does not. So `rows[index]` on `List<T>` and on `T[]` have different parent chains, and the first draft reported the array and silently missed every `List<T>`. 10 fixtures, 303 analyzer tests green.

Editor/Validation/StaleSerializedKeyRepair.cs
```diff
@@ -0,0 +1,189 @@
...
+        private static void Restore(string assetPath, string filePath, byte[] original)
+        {
+            try
+            {
+                File.WriteAllBytes(filePath, original);
+            }
+            catch (Exception)
+            {
+                return;
+            }
```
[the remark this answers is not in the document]
wallstop: **Fixed, and this is the worst failure this type has** -- you are right that it was silent.

The sequence you describe is the one the whole wrapper exists for: `ForceReserializeAssets` has already rewritten the file, the sub-object count came back lower, and the undo is the only thing standing between that and a damaged asset. Swallowing the write failure meant the caller reported `RefusedLostSubObjects` -- *"nothing happened, you are fine"* -- about a file that now held the rewritten bytes.

Three changes, because logging alone would still return the wrong outcome:

1. **`Restore` returns `bool`** and logs an error naming both paths and the recovery: the file holds the rewritten bytes, restore it from source control before saving the project.
2. **A new outcome, `RefusedUndoFailed`.** The caller no longer reports the refusal it intended:

```csharp
return Restore(assetPath, filePath, original)
    ? StaleSerializedKeyRepairOutcome.RefusedLostSubObjects
    : StaleSerializedKeyRepairOutcome.RefusedUndoFailed;
```

3. Both call sites -- the sub-object guard and the `ForceReserializeAssets` catch -- go through it, so there is no path where a failed undo reads as a clean refusal.

The menu already prints every non-`Repaired` outcome and warns when any is present, so this surfaces without further wiring.

Worth being explicit about what is still unproven, since it is adjacent: `RefusedUndoFailed` is unreachable in the tests, because the rewrite path itself is only exercised on its refusal branches. That is [#667](https://github.com/Ambiguous-Interactive/unity-helpers/issues/667), and this outcome adds a fifth fixture to the four listed there -- a read-only target that makes the restore write fail. The guard is better than it was; it is still a guard whose failure branch has never run.

Generator~/WallstopStudios.UnityHelpers.Analyzers/CountingLoopAnalyzer.cs
```diff
@@ -0,0 +1,334 @@
...
+        {
+            if (parent is IArrayElementReferenceOperation array)
+            {
+                return NamesSequence(array.ArrayReference, sequence);
+            }
+
+            return parent is IPropertyReferenceOperation property
+                && property.Property.IsIndexer
+                && NamesSequence(property.Instance, sequence);
+        }
```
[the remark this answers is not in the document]
wallstop: **Correct, and the worst kind of false positive** -- the advice was not just noise, it was wrong. `foreach` cannot assign back into the sequence, for an array or a `List<T>` alike, so following the diagnostic on `rows[index] = value` would have silently dropped the write.

`OnlyIndexesThatSequence` treated every indexer use as foreach-safe because it only asked *which* sequence the index reached into, never whether the element was being **read or written**. A write is the one use of the index that genuinely requires the counting loop, and it was the one shape the rule was blind to.

`IsWrittenThrough` now refuses three forms:

```csharp
case IAssignmentOperation assignment:       return assignment.Target == element;   // rows[i] = v, rows[i] += v
case IIncrementOrDecrementOperation inc:    return inc.Target == element;          // rows[i]++
case IArgumentOperation argument:           return argument.Parameter?.RefKind != RefKind.None;  // Foo(ref rows[i])
```

Simple assignment and compound assignment are both `IAssignmentOperation`, so `=` and `+=` are covered by one case. Five new fixtures, all previously reported and now silent:

| Loop body | Before | Now |
| --- | --- | --- |
| `rows[index] = "replaced";` | **reported** | silent |
| `rows[index] += "suffix";` | **reported** | silent |
| `Replace(ref rows[index]);` | **reported** | silent |
| `List<T>` written through its indexer | **reported** | silent |
| `Process(rows[index]);` (read) | reported | reported |

15 fixtures on this analyzer, 303 across the suite.

Worth noting the rule *shipped off by default* at a measured 127 sites, so this never reached a consumer -- but it would have been the first thing anyone saw on turning it on, and "the analyzer told me to break my code" is how a rule loses its audience permanently. Good catch.

Editor/Utils/ScriptableObjectSingletonCreator.cs
```diff
@@ -1174,7 +1174,7 @@ private static string TryGetAbsoluteAssetsPath(string assetsRelativePath)
                 return string.Empty;
             }
 
-            string projectRoot = Path.GetDirectoryName(Application.dataPath);
+            string projectRoot = Validation.AuthoredAssetPaths.ProjectRoot();
```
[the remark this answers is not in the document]
wallstop: **This one is a false positive, and I checked rather than assumed** -- the code compiles.

C# resolves an unqualified namespace segment by walking outward through the enclosing namespaces. From `WallstopStudios.UnityHelpers.Editor.Utils` the lookup for `Validation` tries:

1. `WallstopStudios.UnityHelpers.Editor.Utils.Validation` — does not exist
2. `WallstopStudios.UnityHelpers.Editor.Validation` — **exists**, and that is the one it binds

So `Validation.AuthoredAssetPaths.ProjectRoot()` resolves. `AuthoredAssetPaths` is `internal` and both files are in `WallstopStudios.UnityHelpers.Editor`, so accessibility is fine too.

The evidence rather than the reasoning, since that is what settles it:

- `ScriptableObjectSingletonCreator.cs` is **not** among `EditorCheck`'s exclusions, so it is one of the 132 files that project compiles.
- `npm run typecheck:editor` builds `Runtime/**` plus `Editor/**` against Unity reference assemblies in four configurations, and it is green on this change — 20 build legs across the four typecheck trees.

I would rather have this reported and dismissed than not reported, though. The reason the line reads oddly is real: a relative namespace qualifier is easy to misread, and it is doing work here that a `using` would make obvious. I have left it as-is because adding `using WallstopStudios.UnityHelpers.Editor.Validation;` to a 1400-line file for one call site is the larger change, but if you would rather it were fully qualified I will do it — it is one token either way.

Generator~/CheckProjects.ruleset
```diff
@@ -14,7 +14,7 @@
 
   WUH010 (a dictionary read through the key indexer, where TryGetValue reports the miss) is the one
   member of the WUH### family that ships off by default: a read of a key that is known present is
-  correct and ubiquitous, so on by default it would bury the other nine on a consumer's first build
+  correct and ubiquitous, so on by default it would bury the other eleven on a consumer's first build
   after a package upgrade. The package still holds itself to it, and this file is how.
```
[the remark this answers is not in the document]
wallstop: **Fixed, both places** -- and the fix immediately broke the build, which is worth recording.

The ruleset's comment claimed `WUH010` was "the one member of the WUH### family that ships off by default" and that "the other eleven WUH### rules" keep their defaults. Both are stale now that `WUH013` also ships off. Updated to name two, and to point at [#671](https://github.com/Ambiguous-Interactive/unity-helpers/issues/671) for the opt-in path.

The thing worth knowing for anyone editing this file next: **`--` is illegal inside an XML comment**, and this file is one 25-line `<!-- ... -->` block. My first edit used the repository's usual `--` dash and the compiler refused the whole ruleset:

```text
CSC : error CS8035: Error reading ruleset file .../CheckProjects.ruleset
  - An XML comment cannot contain '--', and '-' cannot be the last character. Line 16, position 61.
```

That is not a lint failure, it is every check project failing to compile, because a ruleset it cannot parse is a ruleset it will not load. Rewritten with commas and parentheses; `typecheck:unity` is green again.

Thanks for catching the staleness — a comment that says "this is the only one" is exactly the kind that stops being true silently, and this file's whole purpose is to explain a decision to a future reader.

Generator~/WallstopStudios.UnityHelpers.Analyzers/UnityHelpersDiagnostics.cs
```diff
@@ -268,7 +268,7 @@ internal static class UnityHelpersDiagnostics
...
+                "This reads '{0}' through its key indexer, which has nothing to return for a key that is absent: it throws 'KeyNotFoundException', or -- for 'GroupCollection' -- hands back a 'Group' that never matched, so a name the pattern does not declare reads as an ordinary miss forever. Call 'TryGetValue' and handle the absent key on the spot. WUH010 is the one member of this family that is off by default, because reading a key that is known present is correct and ubiquitous and an on-by-default rule here would bury the other eleven; turn it on with a '<Rule Id=\"WUH010\" Action=\"Warning\" />' line in 'Assets/Default.ruleset'.",
```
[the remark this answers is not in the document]
wallstop: **Fixed.** `WUH010`'s help text told the reader it was the only opt-in member of the family, which stopped being true in this PR.

It said:

> WUH010 is the one member of this family that is off by default, because reading a key that is known present is correct and ubiquitous and an on-by-default rule here would bury the other eleven

Now:

> WUH010 is off by default, as WUH013 is, because reading a key that is known present is correct and ubiquitous and an on-by-default rule here would bury the correctness rules

Two changes rather than one, and the second is the durable half: **"the other eleven" is a count that goes stale every time the family grows**, which is exactly how this text became wrong. It has already been corrected twice this session — nine to eleven when `WUH011` and `WUH012` landed — so it now says "the correctness rules" and stops being a number anyone has to remember to update.

This message ships to consumers in the diagnostic itself, so a stale claim here is worse than a stale comment: it is advice a user reads at the moment they are deciding whether to turn a rule on.

The same staleness in `CheckProjects.ruleset` is fixed on [the sibling thread](https://github.com/Ambiguous-Interactive/unity-helpers/pull/665#discussion_r3900997820), where the identical count appeared twice more.

Generator~/WallstopStudios.UnityHelpers.Analyzers/CountingLoopAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+                case IAssignmentOperation assignment:
+                    return assignment.Target == element;
+                case IIncrementOrDecrementOperation increment:
+                    return increment.Target == element;
+                case IArgumentOperation argument:
+                    return argument.Parameter != null && argument.Parameter.RefKind != RefKind.None;
+                default:
+                    return false;
+            }
+        }
```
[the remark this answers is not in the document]
wallstop: **Right again, and it exposed a distinction I had collapsed.** My previous fix only looked at the element's *immediate* parent, so a store one level up read as a plain access.

The distinction that matters, and that I was not making:

| Shape | `foreach` equivalent? | Reported |
| --- | --- | --- |
| `points[index].X = 1;` — **struct** array | **no**, the loop variable is a copy | now silent |
| `nodes[index].X = 1;` — **class** array | **yes**, the write goes through the reference | still reported |

So the refusal cannot be "any member store"; it has to be "a member store on a value-type element". `IsWrittenThrough` now walks up the member chain **while the current node is a value type**, then asks whether the top of that chain is an assignment, compound assignment, increment, or `ref`/`out` argument.

A second vector came out of writing the fixtures, which your description does not name and which I would have shipped: **`points[index].Bump()`**. A mutating method on a struct element mutates the slot, and `foreach` would run it against a copy — silently, not as a compile error. Whether a given method mutates is not cheaply knowable, so **any call on a value-type element** is now left to the counting loop. That is a deliberate false negative: a missed conversion, which is the safe direction for an opt-in style rule, versus advice that changes behaviour.

Seven fixtures on this axis, including the two that must keep reporting:

- `points[index].X = 1;` / `+= 1;` / `++;` / `.Bump();` — silent
- `System.Console.WriteLine(points[index].X);` — **still reported**, because reading a struct member is exactly what `foreach` is for
- `nodes[index].X = 1;` on a class array — **still reported**

313 analyzer tests green.

This is the third false positive in this analyzer, all the same shape: I kept asking *what* an operation touched and not *how*. First a type pattern read as a null test, then an indexer write read as a read, now a member store read as an access. Worth recording on the rule itself rather than just fixing it.

Issue #680, opened 2026-09-02
The license-year walk spends 90% of its time in --find-copies-harder, and 24 files depend on it


wallstop:
Found while closing [#674](https://github.com/Ambiguous-Interactive/unity-helpers/issues/674), which replaced one `git log --follow` per file with a single history walk in `scripts/license-year-lib.sh`. That walk is now the whole cost, and one flag is nearly all of it.

## The measurement

Over 2,163 tracked `.cs` files in this checkout:

| Subject | Wall clock |
| --- | --- |
| `update-license-headers.sh --dry-run`, before #674 | 9 m 49 s |
| after #674 | **1 m 53 s** |
| the single walk alone, inside that | 1 m 42 s |
| the identical walk **without** `--find-copies-harder` | **1.0 s** |

So the flag costs about 100x, and it is ~90% of what the fixer now takes.

## It cannot simply be dropped

Folding both variants and diffing over all 2,163 paths: **24 files resolve to a different year without it.** One example, so the shape is concrete rather than a count:

```text
Runtime/Core/Serialization/JsonConverters/ColorConverter.cs   2023 -> 2025
```

Today's committed headers agree with the `--find-copies-harder` answer, so removing the flag would make `audit-license-years.sh` report 24 mismatches against files nobody has touched. The flag is load-bearing, not decoration.

## Why it might be cheaper than it looks

`user` time is only about 14 s against 1 m 42 s wall. The rest is I/O, and this container's `.git` is a **bind mount from the Windows host** (`devcontainer.json` `mounts`), which is the slowest possible place to do a rename/copy-detecting full-history walk. A native Linux CI runner has no such penalty, so the number that matters for CI may be much smaller than the number measured here.

**Measure it on a CI runner before optimizing anything.** Optimizing a bind-mount artifact would be work spent on a cost the real gate does not pay — the same trap `.llm/context.md` records for benchmarks that measure the instrument.

## Options, if the CI number is also bad

1. **Persist the folded map**, not just the per-path answers. `.git/license-year-cache` is keyed by path and is the audit's; a walk-result cache is a different artifact with different invalidation (it is valid until `HEAD` moves). #668's rule still binds: a staged-rename answer must never be written to a path-keyed cache that outlives the index.
2. **Narrow the walk's pathspec** to the trees the callers actually ask about.
[...]

Issue #644, opened 2026-08-31
#633: audit relational components, reflection, and singleton lifecycle


wallstop:
Parent: #633

## Outcome

Prove relational-component binding, reflection fallback, cache publication, and singleton lifecycles across their complete shape/state matrices; preserve AOT correctness first, then optimize only measured dispatch and traversal costs.

Scope: relational attributes/resolvers, reflection/generated access tiers, caches/factories, runtime/resource/scene singletons, reset hooks, and DI integrations.

## Existing evidence to preserve

Prior work (#529, #534, #535) fixed relational cases but did not establish a complete semantic matrix. The audit also found a typed-query negative result, a no-op cache regression around 30%, costly diagnostics (#564), cache first-fill/race concerns (#536), and singleton lifecycle work (#582). These are constraints: do not reintroduce a cache or typed path because an isolated microbenchmark looks attractive.

### Relational binding

For every relation, define cardinality (single, array, list, set), origin/direction, hierarchy order, active/inactive behavior, tag/layer/predicate filters, optional/required behavior, duplicate policy, interface/base/concrete matching, Unity destroyed-object handling, and deterministic ordering.

### Reflection and generated tiers

Expression, dynamic-IL where available, generated, and reflection-only paths must be semantically identical. Unsupported generation fails soft to an AOT-safe tier. Cache publication is atomic: no consumer sees a partially initialized plan or a failure cached from an incomplete first attempt.

### Singletons

Model uninitialized, authored instance, resource-loaded instance, runtime-created instance, NeverCreate, duplicate, destroyed, scene transition, cache reset/clear, domain reload, and application shutdown. Define main-thread-only operations and how worker access fails or marshals.

### 1. Independent oracles and shape matrix

- [ ] Create a direct Unity-query plus simple-reference-binder oracle sharing no production planner/cache code.
- [ ] Generate the cross-product of cardinality, direction, hierarchy depth, inactive state, filters, optionality, duplicates, interfaces, destroyed objects, and multiple matches.
- [ ] Add metamorphic checks: unrelated objects change nothing; creation reordering matters only where documented; cache warm/cold agrees.
- [ ] Force every execution tier and compare value, order, exceptions/diagnostics, and side effects.
- [ ] Reproduce first-invocation failure, concurrent first fill, reset during lookup, reentrant resolver, and unsupported-AOT cases before fixes.
- [ ] Remove package unsafe reflection access under #637; retain old behavior only as a test oracle if useful.

[...]

Issue #643, opened 2026-08-31
#633: model and benchmark buffer/pool ownership and retention


wallstop:
Parent: #633

## Outcome

Give every buffer, array, comparer, and pooled resource a provable ownership/lifetime model; fix escape, double-return, stale-reference, and retention hazards; then keep pooling only where end-to-end data shows a win.

Scope: package buffer/pool implementations, PooledResource, PooledArray, DisposalLease, shared comparer pools, and every production rent/return site.

## Why this is necessary

The #633 inventory found roughly 88 production rent sites spanning reference/value arrays, exact-size and at-least-size assumptions, exception-prone callbacks, and shared state. Specific hazards include:

- backing-array Length accidentally standing in for logical extent;
- a caller assuming an oversized rented array is exact-sized;
- reference arrays returning without clearing all logically written slots;
- exact-size cardinality causing unbounded or low-value retention;
- disposing one borrower affecting shared comparer infrastructure;
- pooling overhead crossing over unfavorably for small or rare operations;
- DisposalLease behavior not proved under default copies, duplicate dispose, exceptions, and races.

## Required ownership model

Document and test this state machine for each abstraction:

created → rented → returned → available → purged/disposed

Also define owner and borrower responsibilities, logical length versus capacity, initialization and clear-on-return, escape rules, default/copied/stale leases, double return, return after pool disposal, outstanding leases at disposal, callback throw/reentrancy, thread-safety versus SINGLE_THREADED, and maximum retained items/bytes.

No public contract may rely on unspecified ArrayPool contents or exact capacity unless the API itself guarantees them.

### 1. Complete a rent-site ledger

For every site, record owner; requested, logical, and capacity sizes; value/reference payload; initialization; written range; clear behavior; escape behavior; exception edges; concurrency; current allocations; peak bytes; retained bytes; and measured pool/new crossover. Link source locations in the implementation PR.

- [ ] Gate raw backing-array exposure behind explicit logical length, or prove every consumer uses the logical extent.
- [ ] Add a regression for the known oversized-array exactness assumption.
- [ ] Re-run the ledger mechanically after refactors so no unclassified rent site lands.

### 2. Prove lifecycle behavior with shared operation traces

Run identical traces against normal and SINGLE_THREADED implementations:

- [ ] default lease, copied lease, duplicate dispose/return, stale generation, nested rent, and return-after-dispose;
- [ ] concurrent rent/return/purge/dispose;
[...]

Issue #640, opened 2026-08-31
#633: make effect/tag lifecycle reentrancy-safe and benchmark scheduling


wallstop:
Parent: #633

## Context

`EffectHandler.RemoveEffect` invokes cleanup/user code before the handle is fully removed from every registry. Reentrant removal can observe it as active and recurse; a behavior may remove its handle from `OnTick`/`OnPeriodicTick` while the outer loop still enumerates the pooled behavior list that removal clears and returns. Callback failure can also leave partially registered state.

Expiration currently scans the full expiration dictionary every frame. That may be optimal for tiny populations, but it has never been compared with indexed scheduling under representative churn.

## Lifecycle contract

Removal is a two-phase transition:

1. **Detach:** mark the handle inactive and remove it from applied/stack/expiration/handle/periodic/callback indexes before invoking user code.
2. **Notify/release:** perform teardown and return owned resources only when no active traversal still references them.

Removing an unknown/already-removing handle is a no-op. Queries and reentrant callbacks see the old handle as inactive. Reapplication creates an independent new handle. Self-removal stops later tick/periodic callbacks for the removed handle. Pooled behavior/cosmetic/periodic lists remain owned until their outermost traversal exits. Package state and leases complete cleanup in `finally`; preserve/document the existing callback order and propagate callback exceptions only after internal state is consistent.

`RemoveAllEffects`, expiry, stack replacement, cap eviction, and destruction must use the same transition.

## Implementation plan

1. Build a fake-clock model for apply, refresh, stack, tick, remove, and expire.
2. Add recording callbacks that remove, reapply, rent the same buffer type, or throw from every phase.
3. Write red cases for recursive `OnEffectRemoved`, self-removal with later behaviors, reapplication during teardown, pooled-list poisoning, failure during cleanup, and replace/cap eviction reentrancy.
4. Introduce a small internal detached-removal context plus traversal depth/deferred release; keep steady tick paths allocation-free.
5. Route every removal entry point through the same idempotent core and assert exact callback/event counts and order.
6. Add fixed-seed model/property traces with minimized repro cases.
7. Update XML/tag/effect docs and `CHANGELOG.md` with active-state, callback, and exception semantics.

## Expiration experiment

Compare the scan with an array-backed indexed binary min-heap keyed by `(expirationTime, handleId)`, plus `handleId → index` for O(log n) add/refresh/remove. Pop due handles into the existing reusable buffer before callbacks; do not leave tombstones or allocate per frame.

[...]

PR #685, opened 2026-09-02
Scrub the license serial, and fix nine lifecycle defects

wallstop:
## Why this matters

Unity writes its license identity into `unity.log` and `configure.log` while it activates. GitHub masks a registered secret in the rendered job log. GitHub does not rewrite the bytes of an uploaded artifact.

Six jobs uploaded a Unity output directory that still held the serial. Anyone who could download the artifact could read the serial, for 14 days on the test artifacts and 90 days on the benchmark artifacts. A serial is a seat on a two-seat organization license, so a leaked one is a supply problem as well as a credential problem.

The same gap was found and closed in the sibling DxMessaging repository. This is that fix, ported.

### The redactor

`scripts/unity/credential-patterns.js` is the one list of credential shapes. `scripts/unity/redact-unity-artifacts.js` rewrites every match under a directory.

- Each value becomes `<redacted:kind>`. No placeholder can match a pattern, so a second pass changes nothing.
- A label survives where one exists, so a log still reads `UNITY_PASSWORD=<redacted:credential-assignment>` and says which credential was removed.
- A file with a NUL byte is binary and is left alone, so a player build survives the walk.
- A file that cannot be read is reported. It is never treated as clean.
- A masked value such as `TOKEN=***` is not a hit. A false positive would train an operator to skip the step.

`.github/actions/redact-unity-artifacts` is a composite `pwsh` step that runs the script and fails the job if the script fails. A silent pass would recreate the leak.

### The jobs

Six jobs scrub after their last Unity invocation and before their first affected upload, each **scoped to exactly the directories that job uploads**. A wider scope reaches files the job never publishes — the Docker license cache under the same tree is written by root and cannot be rewritten by the runner — and the redactor fails closed on a file it cannot scrub, so a wider scope failed the export smoke job over a file that was never at risk.

| Workflow | Job | Scrubbed |
| --- | --- | --- |
| `unity-tests.yml` | `unity-tests` | `.artifacts/unity/${{ matrix.unity-version }}-${{ matrix.test-mode }}` |
| `unity-tests.yml` | `unity-tests-standalone` | `.artifacts/unity/${{ matrix.unity-version }}-${{ matrix.test-mode }}` |
| `unity-tests.yml` | `unity-tests-single-threaded` | `.artifacts/unity/${{ matrix.unity-version }}-${{ matrix.test-mode }}-single-threaded` |
| `unity-tests.yml` | `unitypackage-smoke` | under `.artifacts/unity/unitypackage-smoke-project/`: `Assets/Editor`, `Packages`, `ProjectSettings`, `unitypackage-output` |
| `unity-benchmarks.yml` | `benchmarks` | `.artifacts/unity/benchmarks/${{ matrix.unity-version }}-${{ matrix.test-mode }}` and `perf-staging` |
| `release.yml` | `unitypackage` | under `.artifacts/unity/unitypackage-project/`: `Assets/Editor`, `Packages`, `ProjectSettings`, `unitypackage-output` |

The two export jobs need four paths rather than one because `.unity-license-cache` is a sibling of those four directories under the same project root, and no job uploads it. The self-hosted legs gain a second benefit from the same rule: a reused runner workspace can hold another leg's leftover output under `.artifacts/unity`, and a leg now scrubs only its own directory.

The redactor's fail-closed behaviour is unchanged. There is no `sudo`, no `chmod`, no ignore list and no skip-unreadable path. A file that holds credential material and cannot be rewritten still fails the job. The fix is that no job now walks a tree it does not publish.

The three self-hosted test legs and the benchmark leg have no Node on PATH, so each installs the repository-wide Node pin first. Every added step carries `timeout-minutes: 2`.

The Unity workflow matrix contract banned `actions/setup-node` on those legs. It banned it to stop asmdef rediscovery, and Node was the vehicle. That ban now names the one permitted step instead of banning the tool, so a `setup-node` added for any other reason still fails the contract.

### The guard

Patching six jobs does not close the failure mode, because the next job that uploads a Unity directory reopens it. `scripts/tests/test-unity-artifact-redaction.js` asserts the invariants instead:

- every upload of a Unity-log-bearing path is preceded, in the same job, by a redaction step that covers that path;
- every redaction path is the directory an upload names, with nothing above it;
- **every Unity upload requires `steps.redact_unity_artifacts.outcome == 'success'`.** Redaction failing closed only protects the artifact if the upload then does not run, and the diagnostic uploads were the worse half: `failure()` is true *because* redaction failed, so a redaction failure was itself the trigger for publishing the tree it failed to scrub;
- every redaction step has a Node toolchain in front of it, and the action still calls the tested script rather than restating the patterns inline.

The workflows are read with a small line scanner, held honest by a self-test that pins the jobs, steps and uploads it must find.

## What else is on this branch

The branch also carries a correctness sweep across four issues. Nine production defects, seven regression tests, and one new gate.

**Effect and tag lifecycle (#640).** Four defects that end the same way — gameplay state raised by a handle that no longer exists, with nothing left that can take it off. `ForceApplyAttributeModifications` applied a modifier after a callback tore the effect down; `ApplyEffectTags` did the same on the tag side; removal decremented tags the handle never raised, taking a count belonging to a *different* effect; and a shared cosmetic was re-applied on every `Refresh` — the default stacking mode — giving two `OnApplyEffect` against one `OnRemoveEffect`. Also: `_attributes` was enumerated across user callbacks at four sites, where the `MoveNext` throw escapes the removal sequence entirely.

**Buffers and pools (#643).** `NextSubset` handed a pooled staging list to a *deferred* iterator, which read it after the `using` had returned and cleared it. `JsonArrayAccumulator` discarded the lease each rent returned, leaking a disposal slot per JSON array deserialized. `PolygonCollider2D.Invert` released N pooled lists after three Unity calls that can throw.

**Relational components and singletons (#644).** `??` on a `ScriptableObject` in all four DI integrations, which passed a destroyed metadata cache through and then skipped the whole scene silently. `ScriptableObjectSingletonRegistry.ClearAllInstances` mutated its own collection mid-iteration. `[ChildComponent]`'s single-field fast path bound a *disabled* `Behaviour`, but only for a sealed element type — which is why both existing tests, both using `BoxCollider`, were green. And `TryAssignParentSingleFast` could never run at all, so it is removed.

**Gate honesty (#636).** `npm run typecheck:controls` proves each of the four type-check projects still reports a `WPROTO###`, a `WUH###` and a `CS####` over its own tree. Two separate control files, because a compiler error in the same compilation silences the `WUH###` analyzer entirely — measured, and now recorded in `.llm/context.md`.

**Also:** `Editor/` is swept to zero comment-block-form violations and leaves the baseline (#635); the license-year walk is 2.7x faster with byte-identical output, and #680's proposed split is refuted with numbers.

## How we know it is correct

- `npm run validate:tests` green: 77/77 contract checks.
- `npm run typecheck:unity` green: 16 builds, all four define configurations across all four trees, 0 errors.
- The redactor was run over a 457-file tree of real CI artifacts downloaded from this repository's own runs. It removed **21** occurrences of the serial from **4** files in **93 ms**, and a second pass reported no change.
- **Red-green in a real editor (6000.4.6f1)** for the four Tags regressions: with the three production files reverted and the tests unchanged, **0/4 pass**, each failing on exactly its own defect (armor 57.0 where 50.0 was expected, a tag raised after detach, another effect's tag decremented, a cosmetic applied twice). Restored: **4/4**. The `[ChildComponent]` fix was proven the same way: the single field bound the disabled component before, and does not after, while the array field never did.
- Deleting the `benchmarks` scrubbing step makes the guard test fail and name both uploads it exposed. Removing one upload's redaction gate makes the new gating assertion fail and name the file, job, step and path.

## Not changed

`scripts/unity/run-unity-docker.sh` has its own `redact_unity_license_output` sed filter for the local Docker path. It is a streaming stdout filter, exported into a subshell with `export -f`, and it also redacts email addresses, which the shared pattern list does not. Replacing it with a per-line Node call is not a safe trade, so it stays as it is.

That filter also sets the honest limit of this change for the two Docker export jobs: their uploaded `unity.log` was already scrubbed by that sed before `tee` wrote it, so the redaction step is defence in depth there. The exposure this change actually removed was on the self-hosted Windows legs, where `run-ci-tests.ps1` writes the log directly. A leg artifact built from this branch carries 12 `<redacted:unity-serial>` placeholders and no raw serial.

## Follow-ups filed

[#687](https://github.com/Ambiguous-Interactive/unity-helpers/issues/687) — `Runtime/Integrations/**` is compiled by no check project, so no `WUH###` rule has ever run over those 19 files. That is how the `??`-on-a-`ScriptableObject` violation above lived in four of them. The remaining audit findings are enumerated on #640, #643 and #644 rather than carried here.

[#690](https://github.com/Ambiguous-Interactive/unity-helpers/issues/690) — `export-unitypackage.sh` points the Docker Unity license cache at `.artifacts/unity/<project>/.unity-license-cache`, so root-owned license material is written into the tree CI uploads from. Nothing uploads it today and `upload-artifact` skips dot-prefixed paths by default, but the Windows path already forbids this shape and the Docker path has no equivalent rule. That directory is what the over-broad scope above collided with.

---

> [!NOTE]
> **High Risk**
> Changes close a credential leak path in public artifact uploads and touch security-sensitive CI wiring, while also modifying gameplay effect/tag lifecycle and singleton/DI behavior in production code.
> 
> **Overview**
> **Stops Unity license serials from shipping in downloadable CI artifacts** by adding a composite `redact-unity-artifacts` action that runs `scripts/unity/redact-unity-artifacts.js` after the last Unity step and **before** any upload. Unity test, benchmark, release, and unitypackage-smoke jobs now scrub only the directories they publish (narrow scope avoids fail-closed redaction on root-owned Docker license cache files). Artifact uploads run only when `steps.redact_unity_artifacts.outcome == 'success'`; self-hosted legs install Node first. Contract tests in `test-unity-artifact-redaction.js` enforce that pairing.
> 
> **Hardens local gates:** `contract-suites` checkout uses `fetch-depth: 0` so the license-year copy-detection gate sees real history, and a new step runs `npm run typecheck:controls` so green `typecheck:unity` runs still prove each of four check projects can surface `WPROTO###`, `WUH###`, and `CS####` (#636).
> 
> **Fixes multiple runtime/editor defects** called out in `CHANGELOG.md`: effect/tag/cosmetic lifecycle bugs on re-apply and removal (#640), `NextSubset` pooled-buffer misuse and related pool/collider issues (#643), relational `[ChildComponent]` / destroyed metadata cache / singleton teardown (#644), plus other listed fixes.
> 
> **Repo hygiene:** sweeps multi-line `//` comments in Editor asset processors and dictionary drawer code to `/* */` block form; updates `.llm` guidance (foreach vs counting loops, `WUH013` site counts, license-year walk notes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit acb34ef94ed274942055d0e60d07713ba60eb5f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/unity-tests.yml
```diff
@@ -656,6 +656,28 @@ jobs:
...
+        with:
+          node-version: "22.18.0"
+
+      - name: Redact credentials from Unity artifacts
+        if: always()
+        timeout-minutes: 2
+        uses: ./.github/actions/redact-unity-artifacts
+        with:
+          paths: .artifacts/unity
+
```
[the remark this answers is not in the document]
wallstop: Fix, sweep and rule — all three.

**Fix.** The three locations you named now require the redaction step to have succeeded. The step carries `id: redact_unity_artifacts`, and each upload's condition gains `steps.redact_unity_artifacts.outcome == 'success'`. The diagnostic-only uploads were the worse half, exactly as you said: `failure()` is *true because redaction failed*, so a redaction failure was itself the trigger for publishing the tree it failed to scrub. Those now read `${{ (failure() || cancelled()) && steps.redact_unity_artifacts.outcome == 'success' }}`.

**Sweep.** Every Unity-output upload in the repository, not only the three: three legs in `unity-tests.yml`, the export smoke diagnostics, `release.yml`'s export diagnostics, and both benchmark uploads in `unity-benchmarks.yml` (including `Upload staged perf results`, which was gated on `run_tests` alone).

**Rule.** `scripts/tests/test-unity-artifact-redaction.js` gains *every Unity artifact upload is gated on redaction succeeding*: it resolves each job's redaction step id and fails any Unity-log-bearing upload whose `if` does not require that step's outcome. Proven red by removing the gate from one upload; it names the file, job, step and path.

Separately, the redaction failure that made this branch red is fixed by scoping each step to the directories its own job uploads (`9cb8641`), so a root-owned license cache the job never publishes can no longer fail it. The fail-closed rule for a file the walk *does* reach is unchanged and now has a test of its own.

Runtime/Tags/EffectHandler.cs
```diff
@@ -867,8 +867,17 @@ private Exception RunAttributeRemoval(EffectHandle handle)
             }
 
             Exception firstFailure = null;
-            foreach (AttributesComponent attributesComponent in _attributes)
+            using PooledResource<List<AttributesComponent>> lease = SnapshotAttributes(
+                out List<AttributesComponent> attributes
+            );
+            for (int index = 0; index < attributes.Count; ++index)
             {
+                AttributesComponent attributesComponent = attributes[index];
```
wallstop: Why is this a counting for loop? We should prefer foreach on anything that has a value-typed enumerator, unless we absolutely need the index inside the loop (or we are working on a strict, interface-typed, IReadonlyList or similar.

Runtime/Tags/TagHandler.cs
```diff
@@ -806,25 +887,77 @@ private void ApplyEffectTags(AttributeEffect effect)
...
+            }
+
+            List<string> effectTags = effect.effectTags;
+            if (effectTags == null)
             {
                 return;
             }
 
+            int applied = 0;
+            for (int index = 0; index < effectTags.Count; ++index)
```
wallstop: Similar foreach, here and throughout

Runtime/Tags/EffectHandler.cs
wallstop: No reason — it should have been `foreach`, and it is now. Fix, sweep and rule.

**Fix.** This loop and its three siblings from the same change (`RunAttributeRemoval`, both `InternalApplyEffect` paths, and the periodic tick) all walk a pooled `List<AttributesComponent>` and never used the index. `TagHandler.ApplyTrackedEffectTags` too, from your other comment.

**Sweep.** Every `WUH013` site under `Runtime/`: **119 → 3**, 114 conversions across 47 files. Each was read before converting, against the cases where a counting loop is load-bearing — the body mutating the walked collection, an index used for anything but `xs[i]`, a non-forward walk, a write-back through the indexer, an interface-typed collection. None of the 114 was any of those.

The three left are the exemption, and each says so where it sits. The clearest is `ProcessEffectExpirations`, whose own comment already recorded it: *"Indexed rather than enumerated: a teardown callback is free to destroy this handler, and OnDestroy clears this very list."*

`Editor/` (190) and `Tests/` (103) are the remainder. They are a separate sweep with a different risk profile — the Editor tree has 60 sites in `SpriteSheetExtractor.cs` alone — and I have recorded the full per-file breakdown on #671 rather than doing it here at the tail of a large branch.

**Rule, and this is the part worth your attention: `WUH013` was wrong, which is why its number was inflated.**

`CountingLoopAnalyzer.NamesSequence` compared the field **symbol** only, ignoring the receiver. So `this.rows[i]` and `other.rows[i]` read as the same sequence, and every hand-written equality method walking two instances in step was reported:

```csharp
for (int i = 0; i < modifications.Count; ++i)
{
    if (modifications[i] != other.modifications[i]) { return false; }   // reported. wrongly.
}
```

`foreach` there loses the parallel index and changes what the method computes, so the rule was pointing at correct code — the thing that stops a rule ever being turned on. It now compares the receiver as well, and refuses a receiver shape it cannot name rather than guessing. Four `Runtime/` findings disappear with it, in `AttributeEffect` and `AnimatedSpriteLayer`.

Red-green: `AWalkOfTwoInstancesInStepIsTheCorrectShape` fails against the old comparison and passes against the new one, with `AWalkOfOneInstancesOwnFieldIsStillReported` beside it so the fix cannot be silently over-broad. The shipped DLL is rebuilt and byte-verified; 335 analyzer tests pass.

**#671's number was 127. Measured across all four check projects, it is 420** — `Editor/` 190, `Runtime/` 126, `Tests/` 104. The old figure predates the Editor and Tests trees being wired in. Both the count and the breakdown are now on that issue.

The policy is written down in [high-performance-csharp](https://github.com/Ambiguous-Interactive/unity-helpers/blob/security/scrub-unity-logs-before-upload/.llm/skills/high-performance-csharp.md) with the four exemptions, and a one-line rule in `.llm/context.md` points at it. Two code examples in that skill were themselves counting loops over concrete collections, and one table row prescribed a `for` loop as the alternative to LINQ — all fixed, since guidance that contradicts itself is worse than none.

Verified after the sweep: `typecheck:unity` 16/16, `lint:repo` 78/78, `validate:tests` 77/77, and 8/8 Tags regression fixtures in editor 6000.4.6f1.

Runtime/Tags/TagHandler.cs
wallstop: Done, and "throughout" taken literally for `Runtime/`. Full write-up on the sibling thread — the short version:

- **This loop**: `ApplyTrackedEffectTags` now walks `effect.effectTags` with `foreach`.
- **Throughout `Runtime/`**: `WUH013` **119 → 3**, 114 conversions across 47 files. The three left are the exemption and each says why where it sits.
- **`Editor/` (190) and `Tests/` (103)** are a separate sweep, recorded on #671 with the per-file breakdown rather than done at the tail of this branch.
- **The rule itself was broken.** `WUH013` compared the field symbol without the receiver, so `this.rows[i]` vs `other.rows[i]` read as one sequence and every parallel-index equality method was reported — pointing the rule at correct code. Fixed, with a red-green test pair; the shipped DLL is rebuilt and byte-verified.
- **#671's 127 is an undercount. The real number is 420**, measured across all four check projects.

The policy now lives in `high-performance-csharp` with the four cases where a counting loop is still correct, and `.llm/context.md` carries a one-line rule pointing at it — so the next author reaches for `foreach` before writing the loop rather than after review.

PR #695, opened 2026-09-03
Stop every pooled rent allocating

### LINKED ISSUE #691: TextureScale.ThreadedScale swallows a worker exception and returns a partial image
Found while fixing #643's finding 7 (the buffer released while workers were still writing). That
race is fixed; this is the adjacent defect the fix made visible, and it is a different failure mode.

## The defect

`Runtime/Utils/TextureScale.cs`, the parallel branch of `ThreadedScale`. Each of the `cores - 1`
background slices is launched as a bare `Task.Run` whose body is:

```csharp
Task.Run(() =>
{
    try
    {
        if (useBilinear) { BilinearScale(...); } else { PointScale(...); }
    }
    finally
    {
        countdown.Signal();
    }
});
```

Nothing holds the returned `Task`, nothing awaits it, and nothing attaches a continuation. The
`finally` signals the countdown **whether or not the slice completed**, so the main thread's
`countdown.Wait()` returns normally and `ThreadedScale` hands back a texture whose rows for that
slice are whatever the pooled buffer already held.

The exception goes to `TaskScheduler.UnobservedTaskException`, which by default is silent from .NET
4.5 on, and Unity does not surface it. So the observable is a **wrong image, with no log** —
a scaled sprite with a band of another texture's pixels in it.

## Why it is reachable

`BilinearScale` and `PointScale` are index arithmetic over `Color[]`, so the honest answer is that
whether they can throw for reachable inputs is **unproven** — #643's audit said the same. What is
certain is the shape: if either ever does throw, this is what a consumer sees, and the single-threaded
branch (`cores <= 1`) does not have the problem, so the same input produces different behaviour on a
one-core machine than on an eight-core one.

## What would close this

- [ ] A red test first. Force a slice to throw — a subclassed or reflected entry point, or a
      deliberately malformed input if one exists — and assert the caller learns about it. Assert
      the single-threaded branch and the parallel branch agree.
- [ ] Decide what the caller should see. `ThreadedScale` is reached from public `TextureScale.Bilinear`
      and `TextureScale.Point`, so per [defensive-programming](../.llm/skills/defensive-programming.md)
      throwing is likely wrong; logging the failure and reporting it is likely right. Consider a
      `TryBilinear`/`TryPoint` pair, which is the pattern the package already uses elsewhere.
- [ ] Capture the first worker exception (an `Interlocked.CompareExchange` into a shared field is
      enough — do not allocate an aggregate per slice) and act on it after the wait.
- [ ] Sweep for the same shape: a `Task.Run` whose result nobody observes, and a `CountdownEvent`
      signalled from a `finally` that cannot distinguish completion from failure.

## Not this issue

The buffer-release race — `countdown.Wait()` sitting outside the `try` whose `finally` signals — is
fixed. That one released the rented destination and premultiplied arrays back to the pool while
workers were still indexing into them, and disposed the `CountdownEvent` they were about to signal.

Related: #643.

### LINKED ISSUE #693: Warm spatial range queries allocate in PlayMode, on three of four 2D structures
Split out of #642, whose "not done" list records that `git grep AllocatingGCMemory -- Tests/Runtime/DataStructures/**` returns nothing, so **"warm-path queries allocate nothing" has never had a gate behind it**. I wrote that gate in #692. It fails, and the measurement is the point of this issue.

## The measurement

`Is.Not.AllocatingGCMemory()` over 256 warm `GetElementsInRange(center, 20f, results)` calls, against a tree of 2,000 points, with the destination list pre-sized and both it and the traversal's pooled rents warmed by 256 prior calls. The control ran first and reported the recorder live.

| Fixture | Editor `6000.4.6f1` | CI PlayMode (all four editors) |
| --- | --- | --- |
| `QuadTree2DTests` | pass | **fail** |
| `RTree2DTests` | pass | **fail** |
| `BalancedKDTree2DTests` | pass | **fail** |
| `UnbalancedKdTree2DTests` | pass | pass (0.005 s) |

Run [33708351764](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/33708351764) — all four playmode legs, 2021.3.45f1 / 2022.3.45f1 / 6000.3.16f1 / 6000.5.2f1, identical:

```text
failed test: …BalancedKDTree2DTests.WarmRangeQueriesDoNotAllocate --
  a warm range query fills a caller's list and rents from a pool, so it allocates nothing once
  both are warm   Expected: not allocates GC memory   But was: <NUnit.Framework.TestDelegate>
```

All four **editmode** legs passed, which is why this had to reach PlayMode to be seen at all.

## What the split result means

`BalancedKDTree2DTests` and `UnbalancedKdTree2DTests` construct the **same type** — `KdTree2D<Vector2>`, differing only in `balanced: true`/`false`. One allocates and the other does not, so this is not "the harness allocates" and not "every query allocates". It is something on the balanced traversal that the unbalanced one does not take, and whatever `QuadTree2D` and `RTree2D` share with it.

## What would close this

- [ ] Find where the allocation is, per structure. `Unity_Profiler_GetFrameGcAllocations` against a PlayMode frame, or a `Recorder` on `GC.Alloc` around a single query, will name it faster than reading the traversals.
- [ ] Decide per structure whether it is fixable or inherent, and say which. An enumerator that boxes, a closure rebuilt per call, a pooled buffer growing past its warmed capacity, and per-access pool statistics are the four shapes worth checking first.
- [ ] Then restore the gate. It is written and reviewed already — see the `WarmRangeQueriesDoNotAllocate` removed in #692 — including the control-first `AllocationProbe.IgnoreWhenUnmeasurable()` (an IL2CPP standalone player cannot report an allocation, so the constraint there is the absence of a measurement) and an assertion that the traversal actually found points.
- [ ] Correct or qualify the claim wherever the documentation makes it, since it is currently false for three of the four 2D structures in PlayMode.

## Why the gate is not in the branch

A red test is not a gate. `#692` ships the shared `Tests/Core/AllocationProbe` — the control that was duplicated across `RestorableGlobalTests` and `SceneResidencyTests` and is now one copy — so restoring the assertion is a few lines once the allocation is understood.

Related: #642, #636.

wallstop:
**Why:** A pooled rental — the cheapest primitive the package has — allocated on a growth boundary for its pool's first ten thousand rents, and each pool then retained 131 KB of usage samples for the process.

**What:**

- Rentals are allocation-free once a pool exists, and usage tracking costs a fixed 2 KB per pool.
- `TextureScale.Bilinear`/`Point` report a failed worker slice instead of returning a partly scaled texture with no log.
- `Helpers.Find(tag)` lets go of the object it cached when the scene unloads; `Helpers.ClearTagCache()` drops the rest.
- The relational field caches are thread-safe, and `CyclicBuffer` and `ImmutableBitSet` no longer trust a capacity a payload only claims.
- The spatial warm-query allocation gate is back, now that the allocation it found is gone.

Fixes #693
Fixes #691

---

> [!NOTE]
> **Medium Risk**
> Touches hot pooling, parallel texture scaling, and multi-threaded relational field assignment caches; failures could affect memory, image correctness, or concurrent Awake/OnEnable assignment.
> 
> **Overview**
> **Pool purge tracking** no longer stores up to ten thousand timestamped samples per pool (which caused allocations on growth and ~131 KB retained per pool). `RollingHighWaterMark` now uses a fixed ring of time buckets, so recording stays **O(1) and allocation-free** and purge sizing **averages every sample in the rolling window**, not a capped recent slice.
> 
> **`Helpers.Find(tag)`** drops cache entries when the cached object is destroyed (including on **scene unload** via `SceneManager.sceneUnloaded`), and **`Helpers.ClearTagCache()`** clears the whole cache. **`[SiblingComponent]` / `[ChildComponent]` / `[ParentComponent]` / `[ValidateAssignment]`** field metadata caches use **`ConcurrentDictionary`** when not `SINGLE_THREADED`.
> 
> **`TextureScale.Bilinear` / `Point`** wait for workers, **capture the first slice failure**, log a second if needed, and **rethrow** instead of returning a partly scaled texture silently.
> 
> **`CyclicBuffer`** JSON/proto restore **raises stated capacity** to fit delivered items (aligned with binary paths). **`ImmutableBitSet`** **clamps capacity** to the word array actually provided. **`EffectHandler`** skips null/destroyed cosmetic components when tearing down. Hot paths use **`ref` locals** for repeated struct-array indexing (`Cache`, `RestorableGlobal`, PowerSort+), with guidance added in **`.llm/skills/avoid-allocations.md`**. Editor asset-postprocessor tests are mostly **comment/doc clarifications**; no new product behavior there.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73729318fbe0dcbae09c8dd3fd31db771206e89d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Utils/RollingHighWaterMark.cs
```diff
@@ -216,76 +190,165 @@ public void Clear()
...
+                _buckets[index].Count++;
+            }
+            else
+            {
+                /*
+                    Every live epoch is within RingLength - 1 of the newest, so live epochs never
+                    share a slot; AdvanceTo has already emptied whatever expired. This slot is
+                    therefore free.
+                */
+                _buckets[index].Occupied = true;
```
wallstop: Why can't we index once and put back? Does repeatedly re-indexing really do anything? Can't we like, ref-index? Here and elsewhere.

wallstop: You're right, and it matters most exactly here — this is the path the whole PR exists to make free, and it was re-indexing on every pooled rent. `73729318`.

**Fix.** `RecordCore` read `_buckets[index]` **eleven** times (six in one branch, five in the other); `AdvanceTo` three and five in its two loops. Each is an address computation and a bounds check. All three take a `ref Bucket` now.

**Sweep** — every place in `Runtime/` and `Editor/` where the same struct array element is read 3+ times in a row with nothing between the accesses that can replace the array:

- `IListSortPowerPlus.BuildPowerSortPlusRuns`, nine accesses across two loops.
- Its heap validation, six accesses over two elements. The bounds guard had to come out of the `&&` chain first — a reference taken above the test that makes it legal is a reference out of range.
- `Cache`: the LFU victim scan, `EvictEntry`, `AddToProbation`, `AddToProtected`.
- `RestorableGlobal`: `BorrowCore`, the constructor's free-list init, `TakeSlot`'s post-resize init, and `ReleaseCore`'s tail.

**Where it deliberately is not applied, which turned out to be the interesting half.** The reference aliases the array, so it lives only as long as that array does. Three sites run caller-supplied code between accesses to one element, and one of those is genuinely dangerous:

- `RestorableGlobal.ReleaseCore` — **live hazard**. `lock (_gate)` is a re-entrant `Monitor`, `_comparer` is the caller's, and a re-entrant borrow reaches `Array.Resize` through `TakeSlot`. So its reference is taken *below* the comparer; the accesses above it stay repeated.
- `Cache.SetUnlocked` and `Cache`'s disposal loop — a `Weigher` and an `OnEviction` handler do run between accesses, but `_lock` is a `ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion)` in **both** compilation modes, so a re-entrant mutator throws before it can reach `Grow`. Those two are conservative rather than required, and I only know that because a review pass made me prove the hazard instead of asserting it. My first draft stated it as fact for all three.

Also skipped: `SevenZip` (vendored verbatim), `List<Rect>` in `SpriteSheetAlgorithms` (no `ref` indexer — `CollectionsMarshal.AsSpan` is .NET 5+), and `SerializableType`'s three reads (one-time static init, not worth the noise).

**Rule: none, deliberately.** The shape is mechanically detectable and the exception is not — nothing static can see whether a callee replaces the array, and a `WUH###` is capped at warning-and-suppressible precisely so it is never wrong in a way that costs the reader. Here being wrong means recommending a silently lost write. The guidance went into [`avoid-allocations.md`](https://github.com/Ambiguous-Interactive/unity-helpers/blob/fix/spatial-allocations-and-worker-faults/.llm/skills/avoid-allocations.md) beside the value-type section instead, hazard first, with the `NoRecursion` distinction spelled out so the next reader does not inherit my first draft's overstatement.

`typecheck:unity` green on the default and `SINGLE_THREADED` branches; CI is running.

Issue #669, opened 2026-08-31
UI/UX: Validation Tool UI


wallstop:
We need to completely align the UI of the validation tool to this thing:

The full spec is here 

[Sentinel Prototype (standalone).html](https://github.com/user-attachments/files/31667231/Sentinel.Prototype.standalone.html)

[Unity validation framework.zip](https://github.com/user-attachments/files/31667232/Unity.validation.framework.zip)

wallstop:
We should ensure that this tool can be entirely turned off in the settings (potentially make it opt-in). No editor UI, no running, etc.

wallstop:
Fix / sweep for the [disable-all request](https://github.com/Ambiguous-Interactive/unity-helpers/issues/669#issuecomment-5563407455) is in #737 (`fc376aab`). **Edit > Preferences > Unity Helpers > Sentinel > Enable Sentinel** is a per-user master switch.

Disabling cancels running/pending work, detaches automatic hooks, closes the workspace and removes toolbar, overlay and inspector surfaces. The sweep covers restored/detached windows, Undo subscriptions, save/import/scene/play-mode/build hooks and reentrant cancellation. Profiles, results and the separate automatic-run preference are preserved. Existing installations retain their enabled preference default; automatic runs remain opt-in. Explicit headless batch validation remains available by deliberate invocation, as documented.

Native Unity 6000.4.6f1 evidence: 85 passes and one existing fingerprint skip across five fixtures; 11 new preference/cancellation/results cases. The preference was also tested with initially disabled state and restored afterward. Before/after screenshots are on the PR and in the feature guide. Default/Odin editor and editor-test builds passed; fleet CI is running.

The broader prototype interaction ledger remains open.

wallstop:
The master-disable slice in #737 passed full fleet acceptance at `249dc218` in [run 34091696794](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/34091696794), including all four editor versions and SINGLE_THREADED. Native interactive testing also passed all six preference cases with a panel control proving the shown-window branch actually ran.

The first fleet run exposed a fixture attempting to show a window under `-nographics`. The final fixture guards only the graphics-dependent Show call; all disable/cancellation/persistence assertions still run headless, with no new skips or ignored errors. The production disable behavior and 11 added editor regressions remain intact. Screenshots and the user guide are on the PR.

[...]

Issue #655, opened 2026-08-31
Asset Validation window: adopt the design system instead of inline styles


wallstop:
Split out of [#634](https://github.com/Ambiguous-Interactive/unity-helpers/issues/634), whose title was the performance half. That half is fixed in session 240; this is the rest of what the report asked for, and it is a different skill.

## What the report said

> It does not follow the design kit
>
> It does not appear to be using UI Toolkit

The second is a misdiagnosis with a real cause: `ValidationWindow` **is** UI Toolkit (`CreateGUI`, `UnityEngine.UIElements`), but it loads **no `StyleSheet` at all**, so every visual is an inline `style.*` assignment or a hard-coded `Color`. It therefore reads exactly like IMGUI to a viewer, which is the complaint.

## The specific gaps

| Site | Today | Should be |
| --- | --- | --- |
| `ValidationWindow.cs` root, toolbar, filter rows | inline `style.padding*`, `style.flexGrow` | a `.uss` class |
| `ValidationWindow.ColorFor` | `new Color(0.94f, 0.42f, 0.38f)` literals | severity classes resolved from the theme, so the window follows light/dark |
| `_progress` | a `Label` reading `"N / M"` | a `ProgressBar`, with the reason for the current choice re-examined |
| `_severity` | a `Button` that cycles three states | a dropdown, once the `PopupField`/`DropdownField` namespace split across supported editors is settled |
| `MakeRow` | a bare `Label` per row | a row with an icon, columns, and click-to-ping affordance -- or a `MultiColumnListView` |

The package already has the pattern to copy: `Editor/Sprites/AnimationViewerWindow.cs` plus `Editor/Styles/AnimationViewer.uss` and its `.uxml`. Nothing under `Editor/Validation/` references a `.uss`.

## Why it is not just "add a stylesheet"

The severity colours and the row chrome are the first place a shared editor theme would be consumed, and `PLAN.md`'s DxKit slice already owns "build the shared UI Toolkit theme and Hub, then migrate windows, settings, and drawers in small characterization-tested batches". Styling this window in isolation would either invent a second palette or block on that theme. Deciding which is the first task here.

## Acceptance criteria

- No hard-coded `Color` and no inline layout style in `ValidationWindow`.
- The window renders correctly in both editor skins, with before/after screenshots.
- Whatever palette it consumes is the same one the next migrated window will consume, or the issue records why this window is the exception.

wallstop:
[...]

PR #708, opened 2026-09-04
Fix runtime queries and editor validation

### LINKED ISSUE #635: less comments

### LINKED ISSUE #654: [Feature]: Migrate UnityMethodAnalyzer to a proper Roslyn Analzyer now
### Feature Description

Right now it's heuristic-based, I think now that we're shipping Roslyn Analyzers, we can get a lot more accuracy and power that we were missing

### Use Case / Motivation

Better reporting

### Proposed Solution

Something like this, but way better.

### Alternatives Considered

_No response_

### Unity Version

None

### Package Version

None

### LINKED ISSUE #671: Work WUH013's 127 counting loops down, then opt the package in
#665 adds `WUH013`, which reports a counting `for` loop over an array or `List<T>` whose body only ever uses the index to reach into that same sequence — the loop `foreach` walks without allocating and says more clearly.

It ships **off by default**, which is the second time this family has done that, and for exactly the criterion `.llm/context.md` names: *the rule is right and the shape is everywhere.*

## The measurement

**127 distinct sites** across `Runtime/`, `Editor/` and `Tests/`, measured 2026-09-01 by turning the rule on and building all four typecheck trees. On by default that would bury the eleven correctness rules on a consumer's first build after taking the upgrade, which is the same reasoning that keeps `WUH010` opt-in.

## Why this is worth doing rather than leaving off forever

`WUH010` is the precedent and it did get worked down: #653 wired all four check projects to it after the sweep, and the ten fixtures whose *subject* is the indexer carry a file-level `#pragma` naming the reason. The rule only earns its place once the package holds itself to it — an analyzer nobody's own code obeys is documentation with a build step.

The reason to prefer `foreach` here is not style. A counting loop over a sequence the body only indexes states an ordering and an offset it does not use, so a reader has to check whether the index is load-bearing before they can change anything. `foreach` answers that at a glance.

## What would close this

1. Convert the reportable sites. Most are mechanical: the loop becomes `foreach` and the element gets a name.
2. Where a site is genuinely better as a counting loop, leave it and let the rule say why — but note that the three legitimate shapes are **already not reported**: an interface sequence (whose enumerator boxes), a body that uses the index itself, and any walk that is not the ordinary forward one. A site that still reports after those exclusions is very likely a real conversion.
3. Add `<Rule Id="WUH013" Action="Warning" />` to `Generator~/CheckProjects.ruleset` beside `WUH010`, which makes it binding on `Runtime/`, `Editor/`, `Tests/Runtime/` and `Tests/Editor/`.
4. Update the roster counts in `.llm/context.md`, `docs/performance/analyzers.md` and the ruleset comment.

## Acceptance criteria

- Reintroducing a reportable counting loop into any of the four trees fails a local command, proven by doing it.
- The remaining `#pragma`-suppressed sites each name their reason in the file they apply to.

Related: #665, #653.

### LINKED ISSUE #707: Range overlap mishandles open and empty intervals
Found while continuing the gameplay correctness audit in #633.

`Range<T>.Overlaps` currently tests whether any endpoint lies inside the other interval. That is not sufficient for interval intersection:

```csharp
Range<int>.Exclusive(0, 10).Overlaps(Range<int>.Exclusive(0, 10)); // false, should be true
Range<int>.InclusiveExclusive(0, 10).Overlaps(Range<int>.Inclusive(10, 20)); // true, should be false
Range<int>.Exclusive(5, 5).Overlaps(Range<int>.Inclusive(0, 10)); // true, should be false
```

An inverted range created by editing the public min/max fields can also report overlap. The existing `OverlapsHandlesExclusiveBoundaries` test incorrectly expects the second example to overlap.

RCA: independently accepting an endpoint ignores that the range owning it may exclude it, and two coincident open intervals have no included endpoint despite sharing their interior.

Red evidence: the actual Range and Objects sources compiled in an isolated net9 NUnit harness fail 21 of 24 focused overlap cases. The new independent point-membership oracle exercises 3,600 interval pairs across all endpoint-inclusion combinations, plus inverted public bounds and long extrema. Unity metadata alone is shimmed; interval logic is not copied or mocked.

Acceptance:
- Compare interval ordering and jointly included touching boundaries without endpoint arithmetic.
- Coincident open intervals overlap; zero-width intervals are nonempty only with both endpoints included; inverted public bounds return false.
- Preserve symmetry and inclusive behavior, and correct the incorrect test expectation.
- Pass the independent finite point-membership matrix and extreme-boundary regressions.
- Update API and feature documentation plus the released-behavior changelog.

Generic intervals use endpoint ordering; this does not promise to enumerate representable values between distinct endpoints of an arbitrary T.

### LINKED ISSUE #709: Child binding selects deep matches before nearer children
Parent: #644

## Reproduction

On Unity 6000.4.6f1 at `1e7695d3`, create a root with no SpriteRenderer, a first child containing a chain of three deeper SpriteRenderers, and a second immediate child with a SpriteRenderer. Assign the existing ChildSingleTester and ChildMaxCountTester fixtures through the public AssignChildComponents method.

Observed: the single field selects the first deeper renderer; MaxCount=3 returns all three deep renderers and excludes the immediate child. The unlimited collection correctly starts with the immediate child.

Expected: the documented breadth-first contract selects the immediate child first; a capped collection is the first N eligible matches in that same order.

## RCA

TryAssignChildSingleFast takes the first Unity GetComponentsInChildren result, which differs from the documented breadth-first traversal. The collection fast path runs FilterChildren (including MaxCount) before EnsureBreadthFirstOrder, permanently discarding nearer matches before ordering.

## Fix and acceptance

Use the existing breadth-first single resolver and bounded collection traversal; retain the unbounded collection path. Pin single selection and array/list/set quotas against an independent hierarchy traversal, including reversed sibling order, deep branches, self exclusion, warm repeats and non-generic collector fallback. Run focused Unity regression invocations, source compatibility builds and repository validation. Update XML, feature docs and changelog. This issue covers the concrete ordering defect; #644 remains the broader audit.

wallstop:
**Why:** Runtime queries return incorrect results, and editor diagnostics and validation need reliable checks and consistent controls.

**What:**

- Correct range overlap, nearest-child selection, inherited invokers and malformed-reader handling.
- Replace heuristic Unity callback checks with Roslyn diagnostics and a compiler report.
- Complete the validation workspace and migrate both windows to a shared editor theme.
- Convert eligible counting loops and enforce WUH013 across package source checks.
- Audit source comments and remove comment-format baseline exemptions.

Fixes #707
Fixes #709
Fixes #654
Fixes #671
Fixes #635
Related: #655, #669. Native checks and screenshots in both editor skins await editor reconnection.

---

&gt; [!NOTE]
&gt; **Medium Risk**
&gt; Reflection and IL emission changes affect core dispatch paths; incorrect caching or IL could cause subtle runtime bugs, though coverage is extensive.
&gt; 
&gt; **Overview**
&gt; Fixes **typed reflection invokers** so derived, sibling, and struct receivers dispatch correctly and do not share one cached delegate per `MethodInfo` alone. Instance caches now key on `(method, typeof(TInstance))`, receiver checks use `DeclaringType.IsAssignableFrom(typeof(TInstance))`, value-type IL uses `EmitValueReceiverCall` for interface and inherited calls, static invokers validate return types before caching, and `ClearMethodCache` clears typed delegate caches.
&gt; 
&gt; Rewrites **`Range.Overlaps`** to respect endpoint inclusivity, reject empty or inverted intervals, and treat touching endpoints as overlap only when both sides include the shared point—plus docs, changelog (#644, #707), and broad tests.
&gt; 
&gt; Agent GitHub guidance now asks fallbacks to **record which MCP capability was missing** in the same message as the fallback.
&gt; 
&gt; <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e7695d35d454e77953090ab0944e4abec2e2e38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/AssetProcessors/AssetPostprocessorDeferral.cs
```diff
@@ -59,33 +59,16 @@ internal static void Schedule(Action drain)
...
-                identity-based. Using List<Action>.Contains would invoke
-                Delegate.Equals and collapse structurally-equal-but-semantically-distinct
-                lambdas — for example, lambdas produced by a local function that
-                captures only outer-method variables all share the same Method+Target
-                because the compiler lowers them onto the outer method's display class.
-            */
+            // Deduplicate delegate identities; structural equality can merge separately created callbacks sharing captured state.
             bool alreadyPending = false;
-            for (int i = 0; i < PendingDrains.Count; i++)
+            foreach (System.Action PendingDrainsElement in PendingDrains)
```
wallstop: Why is this PascalCase? local variables should be camelCase. We need to respect editorcofnig

wallstop: Yes, this should be camelCase. The counting-loop conversion carried the source field's capitalization into the new local. I fixed this local and swept all four source trees, finding and correcting 12 affected foreach locals across seven files. I also added an explicit ordinary-local camelCase rule to .editorconfig using its existing lower_camel_case_style; the more specific constant rule remains intact. These changes are local and under validation now; I'll include them in the next verified push.

wallstop: Pushed in 621c0530 (current PR head 4ec5f1f0). Final sweep found 14 locals across the same seven files: the multiline scan added two tuple-typed foreach declarations to the initial 12. All declarations and references are camelCase, with no collisions or changes beyond the expected renames. The explicit .editorconfig local rule is included. Independent review, targeted editor/runtime-test source builds, and both preflight modes passed.

Issue #705, opened 2026-09-04
chore: Reduce/Optimize Test Runtimes


wallstop:
Is there any further optimizations or cost reductions we could make to our Unity test run matrix, without sacrificing coverage, that could reduce e2e test times in CI?

wallstop:
PR #704 now includes a measured iteration-speed contribution in a1735ece.

Baseline: Unity run 33914473384 and hosted type-check job 101158329020 on 9a1409c5.

| Work | Measured baseline | Change |
| --- | --- | --- |
| Hosted type-check job | 42m14s; sources about 26m46s followed by about 15m of negative controls | Run both complete commands in parallel, in isolated hosted checkouts |
| SerializableType eviction test | 45.035s total across four PlayMode versions | Replace 32 filler searches with the two distinct keys required to evict; explicitly assert cache hit and actual recomputation |
| SerializableType live-shrink test | 12.903s total across four PlayMode versions | Create three entries rather than twelve; prove pre-shrink overflow and exact post-shrink size |

Coverage: no Unity version, mode, category, test case, source variant, or negative control was removed. Result-equality assertions remain; new assertions prove the cache transitions instead of relying on oversized setup. The stable `Type-check Runtime and tests` gate accepts only success from both phases. Failure, cancellation, skipped and empty results are exercised using the exact workflow predicate. Eleven host contract cases pass, also with no external commands on PATH; targeted host test compiles, YAML, actionlint, formatting, spelling and staged preflight pass. No local editor suite was launched.

Tradeoff: hosted parallelism targets elapsed time, not billed-compute savings. It duplicates setup and loses warm build outputs between phases. Actual post-change savings await CI; the new run is pending.

Investigated but deferred: sharing default EditMode/PlayMode persistent projects. The apparent 153.56s import opportunity was on a mode pair assigned to different physical hosts. Only the 6000.3 pair shared a host in this run; its second import cost 7.895s. Workspace locking, active-project pruning and cold-cache migration were disproportionate to that observed saving, so that exploratory patch is not included. Existing project isolation and source-inventory invalidation remain unchanged.

wallstop:
[...]

PR #712, opened 2026-09-05
Extend serialized hierarchies across assemblies

### LINKED ISSUE #612: Cross-assembly subtypes without a runtime registry: emit the base's chain in the extending assembly
## What

[#603](https://github.com/Ambiguous-Interactive/unity-helpers/issues/603) asked for cross-assembly subtype dispatch and framed the answer as a **runtime registry**. That was refused, correctly: unordered registrars, undetectable tag collisions, and stripping hazards are all silent-data-corruption failure modes.

**There is a second mechanism that #603 never considered, and it is not a registry.** Three facts, all measured against the shipped generator rather than assumed:

1. **A consumer's compilation can read everything about the package base.** Compiling a consumer assembly against a real generated package assembly and asking Roslyn for `Upstream.Weapon`:

   ```text
   resolved Upstream.Weapon: True
   attribute: WProtoContractAttribute args=[]
   attribute: WProtoIncludeAttribute args=[100, Upstream.Melee]
   members with [WProtoMember]:
     Damage tag=1
   nested formatter present in metadata: True
   ```

   So the consumer's generator already knows **every field number spent on `Weapon`** -- members and includes -- at build time. Collision detection needs no new information.

2. **Replacing a formatter is already supported and deliberate.** `WProtoFormatterProvider`'s own documentation: *"Last registration wins, so a consumer can deliberately replace a formatter this package ships."*

3. **The generated chain is ordinary static code.** Nothing in it requires the emitting assembly to be the one that declared the base.

## The mechanism

The extending assembly emits a **complete replacement formatter for the base**, containing the package's subtypes (read from metadata) plus its own, and registers it. No registry lookup, no `MakeGenericType`, no reflection -- the dispatch stays the same straight-line `is`-chain, just emitted one assembly later.

```csharp
// In ConsumerAssembly, generated:
// Weapon's chain, re-emitted: Melee at 100 (read from UpstreamAssembly's metadata),
// PlasmaCutter at 200 (declared here).
```

This is IL2CPP-safe for the same reason the current chain is, and it is `[Preserve]`-able the same way.

## Why this can be compile-time safe where a registry cannot

| Hazard from #603 | Registry | Replacement chain |
| --- | --- | --- |
| Tag collides with one the base already spent | undetectable at build time | **build error** -- the tags are in metadata |
| Registrars run unordered | writes under the wrong tag or none | only one assembly emits a given base's chain in the common case |
| Registration after first use | needs freeze semantics | does not arise; registration is startup-only as today |
| Stripping | must preserve types nothing references | same static references as today |

## The one genuinely undecidable case, and how it is contained

**Sibling assemblies.** `A` and `B` both reference the package, neither references the other, both extend `Weapon`. Neither compilation can see the other, so neither can refuse the collision. Both emit a replacement; one wins at startup and the other's subtype throws.

This is not solvable inside a single compilation, and any design must say so. Three containments, in order of preference:

1. **Project-time arbitration.** `WProtoSubtypeTagAssigner` already surveys **every assembly** through `TypeCache` and writes committed per-assembly manifests. That is exactly the scope where this collision exists, and it already assigns numbers without collision within an assembly. Extending it to arbitrate across assemblies is the natural fit, and it keeps numbers committed rather than derived.
2. **Build gate.** `WProtoSubtypeTagBuildGate` already refuses a player build over unnumbered subtypes; two replacements for one base is the same kind of refusal.
3. **Runtime guard, last resort.** The provider detects a second replacement for a base and throws at startup, deterministically, instead of silently dropping one. Loud, not silent -- which is the property #603's registry could not offer.

**Reserved ranges would prevent it rather than detect it.** A package could declare an extension range (`[WProtoExtensible(1000, 1999)]`), and independently shipped extenders draw from it. Worth deciding early, because it changes what a package author writes.

## What I am NOT claiming

- That this is cheap. It is a real feature.
- That case 3 above meets "no runtime issues" in the strictest reading. Two independently shipped packages extending one base are arbitrated when the **project** assembles them, not when either package compiles, because neither can see the other. Every other case is a build error.
- That the current refusal is wrong. It is right until this exists; `WPROTO040` should keep firing.

## Acceptance criteria

- A consumer assembly declares a subtype of a package base and it round-trips, as that subtype, through a base-typed member and a base-typed collection.
- A tag colliding with one the base already spends is a **build error** in the consumer's compilation, naming what holds it.
- The emitted dispatch is static -- no reflection, no `MakeGenericType` -- and survives managed stripping, proven on a standalone player.
- Two assemblies replacing one base's chain is refused: at project assembly if the assigner can see both, at the build gate otherwise, and at startup as a deterministic throw if it reaches a player.
- Round-trip against protobuf-net 2.4.9 and 3.2.56 preserved in both directions.
- `WPROTO040`'s message and the serialization guide stop describing the limit as permanent, and point here.

## Note on how this was found

Filed after closing #603 as a permanent refusal. That closure was correct **about a runtime registry** and the diagnostic wording overreached by implying the feature itself could never exist. The three measurements above were taken afterwards; they are what changed the assessment.

### LINKED ISSUE #655: Asset Validation window: adopt the design system instead of inline styles

### LINKED ISSUE #706: TypeCheck CI is extremely expensive
Can we optimize it and/or move it to one of our self-hosted linux runners?

Image

### LINKED ISSUE #710: dev: research codegraph
Codegraph might help speed up local development. If it's good, we should auto-integrate it into the devcontainer for agentic sessions. https://github.com/colbymchenry/codegraph

### LINKED ISSUE #713: Random quality battery mismatch (PractRand)
A scheduled PractRand run disagreed with `scripts/random-quality/expected-outcomes.json`.

- Battery: **PractRand 0.95**
- Seed: `00010203-0405-0607-0809-0a0b0c0d0e0f`
- Byte budget per generator: **8GB**
- Run: https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/33988552133

A generator expected to PASS that failed is a statistical regression.
A deliberate expected-failure control that PASSED is evidence the harness is broken.

# Random quality battery report (32-bit stream)

- Battery: **PractRand 0.95** (`5738c322ccb983b63fbe693f305afa2a7c73e64a80b303c9139278c453ce27b4`)
- Source: https://sourceforge.net/projects/pracrand/files/PractRand-pre0.95.zip/download
- Seed: `00010203-0405-0607-0809-0a0b0c0d0e0f`
- Stream width: **32-bit**
- Byte budget per generator: **8GB**
- Command: `RNG_test stdin32 -tlmin 1KB -tlmax 8GB -multithreaded`
- Run: https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/33988552133

| Generator | Quality | Expected | Observed | Reached | Result |
| --- | --- | --- | --- | --- | --- |
| BlastCircuitRandom | - | pass | missing | - | MISMATCH |
| DotNetRandom | - | fail | missing | - | MISMATCH |
| FlurryBurstRandom | - | pass | missing | - | MISMATCH |
| IllusionFlow | - | pass | missing | - | MISMATCH |
| LinearCongruentialGenerator | - | fail | missing | - | MISMATCH |
| PcgRandom | - | pass | missing | - | MISMATCH |
| PhotonSpinRandom | - | pass | missing | - | MISMATCH |
| RomuDuo | - | pass | missing | - | MISMATCH |
| Sfc64Random | - | pass | missing | - | MISMATCH |
| SplitMix64 | - | pass | missing | - | MISMATCH |
| SquirrelRandom | - | fail | missing | - | MISMATCH |
| StormDropRandom | - | pass | missing | - | MISMATCH |
| SystemRandom | - | fail | missing | - | MISMATCH |
| WaveSplatRandom | - | fail | missing | - | MISMATCH |
| WDoomRandom | - | fail | missing | - | MISMATCH |
| WyRandom | - | pass | missing | - | MISMATCH |
| XoroShiroRandom | - | pass | missing | - | MISMATCH |
| XorShiftRandom | Fair | fail | fail | 64 kilobytes (2^16 bytes) | OK |
| Xoshiro128StarStar | - | pass | missing | - | MISMATCH |
| Xoshiro256StarStar | - | pass | missing | - | MISMATCH |

### BlastCircuitRandom

No PractRand report was produced for this generator.

### DotNetRandom

No PractRand report was produced for this generator.

### FlurryBurstRandom

No PractRand report was produced for this generator.

### IllusionFlow

No PractRand report was produced for this generator.

### LinearCongruentialGenerator

No PractRand report was produced for this generator.

### PcgRandom

No PractRand report was produced for this generator.

### PhotonSpinRandom

No PractRand report was produced for this generator.

### RomuDuo

No PractRand report was produced for this generator.

### Sfc64Random

No PractRand report was produced for this generator.

### SplitMix64

No PractRand report was produced for this generator.

### SquirrelRandom

No PractRand report was produced for this generator.

### StormDropRandom

No PractRand report was produced for this generator.

### SystemRandom

No PractRand report was produced for this generator.

### WaveSplatRandom

No PractRand report was produced for this generator.

### WDoomRandom

No PractRand report was produced for this generator.

### WyRandom

No PractRand report was produced for this generator.

### XoroShiroRandom

No PractRand report was produced for this generator.

### Xoshiro128StarStar

No PractRand report was produced for this generator.

### Xoshiro256StarStar

No PractRand report was produced for this generator.

# Random quality battery report (64-bit stream)

- Battery: **PractRand 0.95** (`5738c322ccb983b63fbe693f305afa2a7c73e64a80b303c9139278c453ce27b4`)
- Source: https://sourceforge.net/projects/pracrand/files/PractRand-pre0.95.zip/download
- Seed: `00010203-0405-0607-0809-0a0b0c0d0e0f`
- Stream width: **64-bit**
- Byte budget per generator: **8GB**
- Command: `RNG_test stdin64 -tlmin 1KB -tlmax 8GB -multithreaded`
- Run: https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/33988552133

| Generator | Quality | Expected | Observed | Reached | Result |
| --- | --- | --- | --- | --- | --- |
| BlastCircuitRandom | - | pass | missing | - | MISMATCH |
| DotNetRandom | - | fail | missing | - | MISMATCH |
| FlurryBurstRandom | - | pass | missing | - | MISMATCH |
| IllusionFlow | - | pass | missing | - | MISMATCH |
| LinearCongruentialGenerator | - | fail | missing | - | MISMATCH |
| PcgRandom | - | pass | missing | - | MISMATCH |
| PhotonSpinRandom | - | pass | missing | - | MISMATCH |
| RomuDuo | - | pass | missing | - | MISMATCH |
| Sfc64Random | - | pass | missing | - | MISMATCH |
| SplitMix64 | - | pass | missing | - | MISMATCH |
| SquirrelRandom | - | fail | missing | - | MISMATCH |
| StormDropRandom | - | pass | missing | - | MISMATCH |
| SystemRandom | - | fail | missing | - | MISMATCH |
| WaveSplatRandom | - | fail | missing | - | MISMATCH |
| WDoomRandom | - | fail | missing | - | MISMATCH |
| WyRandom | - | pass | missing | - | MISMATCH |
| XoroShiroRandom | - | pass | missing | - | MISMATCH |
| XorShiftRandom | Fair | fail | fail | 64 kilobytes (2^16 bytes) | OK |
| Xoshiro128StarStar | - | pass | missing | - | MISMATCH |
| Xoshiro256StarStar | - | pass | missing | - | MISMATCH |

### BlastCircuitRandom

No PractRand report was produced for this generator.

### DotNetRandom

No PractRand report was produced for this generator.

### FlurryBurstRandom

No PractRand report was produced for this generator.

### IllusionFlow

No PractRand report was produced for this generator.

### LinearCongruentialGenerator

No PractRand report was produced for this generator.

### PcgRandom

No PractRand report was produced for this generator.

### PhotonSpinRandom

No PractRand report was produced for this generator.

### RomuDuo

No PractRand report was produced for this generator.

### Sfc64Random

No PractRand report was produced for this generator.

### SplitMix64

No PractRand report was produced for this generator.

### SquirrelRandom

No PractRand report was produced for this generator.

### StormDropRandom

No PractRand report was produced for this generator.

### SystemRandom

No PractRand report was produced for this generator.

### WaveSplatRandom

No PractRand report was produced for this generator.

### WDoomRandom

No PractRand report was produced for this generator.

### WyRandom

No PractRand report was produced for this generator.

### XoroShiroRandom

No PractRand report was produced for this generator.

### Xoshiro128StarStar

No PractRand report was produced for this generator.

### Xoshiro256StarStar

No PractRand report was produced for this generator.

### LINKED ISSUE #714: Skip local Unity diagnostics when checkout never completed
## Observed failure

[Unity package export smoke job 101381555898](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/33991528769/job/101381555898), commit `826a11f9bdb8e28c6c51087bfc0e83a52a4ca4f5`, was correctly rejected before checkout because PR #712 had advanced to `4aa8a9078a55b3ebee141c5bf2882a13d62def2d`. Two later steps then failed separately because `.github/actions/dump-unity-log-tail` and `.github/actions/redact-unity-artifacts` did not exist in the unpopulated workspace.

The primary stale-head refusal is intended. The extra failures are a workflow dependency bug: status-function conditions such as `always()` and `failure() || cancelled()` bypass the default success gate without requiring the checkout that supplies these local actions.

## Scope and sweep

Audit all repository-local actions in workflow failure/cancellation paths. The unconditional diagnostic/redaction pattern also appears in `unity-benchmarks.yml` and `release.yml`; default and SINGLE_THREADED Unity redactors likewise have explicit status-function conditions. A self-hosted workspace left from an earlier run must not substitute for a successful checkout in the current job.

## Acceptance

- Local action invocation requires a successful current-job checkout, either explicitly or through a verified prerequisite.
- Skipped, failed and cancelled checkout do not cause missing-action errors or process a previous workspace's artifacts.
- After checkout succeeds, relevant diagnostics/redaction still run on failed or cancelled test/export steps; artifact uploads still require successful redaction.
- License return and lock release retain their existing acquisition-based conditions.
- Regression controls evaluate the actual workflow predicates across checkout outcomes and fail when the guard is removed; existing workflow and acceptance contracts remain green.

The fix is being prepared in aggregate PR #712. No implementation or new-head CI result is claimed yet.

wallstop:
**Why:** Downstream contracts need to extend serialized hierarchies without losing inherited state.

**What:**

- Extend serialized hierarchies across assemblies while preserving inherited state.
- Reduce script compilation work for generic protobuf and JSON contracts.
- Correct exact random sampling failures and statistical report validation.
- Restore Sentinel layout and graph controls in both editor themes.
- Add native acceptance, checkout guards, and the CodeGraph evaluation.

Fixes #612, fixes #655, fixes #706, fixes #710, fixes #713, fixes #714.

| Theme | Before | After |
| --- | --- | --- |
| Light | Before light | After light |
| Dark | Before dark | After dark |

---

&gt; [!NOTE]
&gt; **High Risk**
&gt; Cross-assembly serialization changes wire dispatch, replacement registration, and build gates—mistakes affect save compatibility and runtime subtype identity. CI workflow changes alter perf baseline policy and TestU01 failure semantics.
&gt; 
&gt; **Overview**
&gt; **Cross-assembly WallstopProto hierarchies** are now supported: consumer assemblies can declare subtypes of a base in another assembly when that base ships a generated extension body (`ReadWithSubtypes`). The generator emits **replacement formatters** that merge upstream dispatch metadata with local subtypes, registers them via `[WProtoReplacement]`, and enforces **one extending assembly per external base** (compile-time and in the subtype tag assigner/build gate). Tag assignment inventories cross-assembly bases, reserves upstream field numbers, and refuses writes on conflicts. Diagnostics **WPROTO040** / **WPROTO044** and large generator test coverage reflect the new rules instead of blanket same-assembly refusal.
&gt; 
&gt; **Generator compile cost** for generic protobuf/JSON contracts drops by scanning syntax trees once (`ClosureScan.Types`) and reusing that list for marshal and JSON converter registration; formatter naming goes through **`WProtoGeneratedNames`**.
&gt; 
&gt; **CI and hooks:** pre-commit meta checks skip `Samples~`-style paths; TestU01 SmallCrush **fails the job when the driver exits non-zero** (SIGPIPE on the host no longer masks harness faults); scheduled Unity benchmarks **stop auto-promoting `baseline.json`** (reports/XML still commit); `unity-tests` adds **workflow_dispatch native acceptance** (sentinel/intmap/all) with Il2Cpp provisioning when needed.
&gt; 
&gt; **Sentinel validation UI:** USS is reformatted and **`.dx-editor .sentinel-body`** scopes layout; graph connection mesh **index order** is fixed; rule builder **shows fix-value only** for rename/import-size fixes via shared constants.
&gt; 
&gt; <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15c4a53cfeceae9605a01b2ed0217c30a7ba599d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

&gt; [!NOTE]
&gt; **High Risk**
&gt; Cross-assembly serialization changes wire dispatch, replacement registration, and build gates—errors affect save compatibility and subtype identity; CI changes alter TestU01 failure semantics and perf baseline promotion policy.
&gt; 
&gt; **Overview**
&gt; **Cross-assembly WallstopProto** now lets consumer assemblies add subtypes to a base in another assembly when that base ships a generated extension body (`ReadWithSubtypes`). The generator emits **replacement formatters**, registers them at startup, and blocks **two assemblies** from extending the same external base (compile-time, tag assigner, and build gate). Tag assignment pulls in upstream reserved numbers and refuses writes on conflicts; diagnostics **WPROTO040** / **WPROTO044** replace the old same-assembly-only refusal.
&gt; 
&gt; **Generator compile cost** drops by scanning syntax once (`ClosureScan.Types`) for marshal/JSON registrations and centralizing formatter names in **`WProtoGeneratedNames`**.
&gt; 
&gt; **CI and hooks:** pre-commit meta rules treat nested `~` folders under `Samples~` correctly; TestU01 **fails on driver exit** instead of treating SIGPIPE as success; PractRand mismatch issues only **close on recoverable** full-inventory runs; Unity jobs gate log dump/redaction on **successful checkout**; scheduled benchmarks **stop auto-updating** `baseline.json`; **`unity-tests`** adds optional **native acceptance** on manual dispatch (sentinel/intmap/serialization/all).
&gt; 
&gt; **Sentinel / editor:** validation preferences apply through shared **`ApplyPreferences`**; rule builder fix fields follow fix-type constants; graph line mesh **triangle index order** fixed; **`ValidationWindow.uss`** reformatted with `.dx-editor` scoping. **`ShowIfConditionEvaluator`** uses internal **`TryConvertValue`** instead of out-param conversion at call sites.
&gt; 
&gt; <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c441d0816f7ef35adc0d77cc153b3895509d04a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

&gt; [!NOTE]
&gt; **High Risk**
&gt; Cross-assembly subtype dispatch, replacement registration, and manifest/build gates directly affect wire compatibility and runtime subtype identity; CI changes alter TestU01 failure semantics and perf baseline promotion policy.
&gt; 
&gt; **Overview**
&gt; **Cross-assembly WallstopProto** lets a referencing assembly add subtypes to a base defined elsewhere when that base’s generated formatter exposes an extension entry point (`ReadWithSubtypes`). The generator emits **replacement formatters**, registers them at startup, and blocks **two assemblies** from extending the same external base. Subtype tag assignment now inventories cross-assembly bases, merges upstream reserved/dispatch field numbers, and refuses manifest writes on conflicts; the **pre-build gate** surfaces replacement conflicts for shipped assemblies. Diagnostics **WPROTO040** / **WPROTO044** and generator tests replace the prior blanket same-assembly refusal.
&gt; 
&gt; **Compile-time cost** for generic protobuf/JSON registration is reduced by scanning syntax once (`ClosureScan.Types`) and sharing results across marshal and JSON converter maps; nested formatter names go through **`WProtoGeneratedNames`**.
&gt; 
&gt; **CI and hooks:** pre-commit meta checks ignore nested `~` tooling paths under `Samples~`; **TestU01** fails when the driver exits non-zero (host SIGPIPE no longer masks harness faults); PractRand comparison exposes a **recoverable** flag so mismatch issues only auto-close on full-inventory recovery; Unity workflows gate log dumps and credential redaction on **successful checkout**; scheduled benchmarks **stop auto-promoting** `perf-results/baseline.json` while still committing reports/XML; **`unity-tests`** adds manual **native acceptance** (`sentinel` / `intmap` / `serialization` / `all`) with Il2Cpp provisioning when needed.
&gt; 
&gt; **Sentinel:** rule preferences apply via shared **`ApplyPreferences`** (window, status surfaces); the rule builder shows fix-value fields only for rename/import-size fixes; graph connection mesh **triangle index order** is corrected; **`ValidationWindow.uss`** is reformatted with `.dx-editor` scoping. **`ShowIfConditionEvaluator`** routes conversions through **`TryConvertValue`** internally.
&gt; 
&gt; <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0311807fee19189838cd31978ed2c474b5c22d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

&gt; [!NOTE]
&gt; **High Risk**
&gt; Cross-assembly subtype dispatch, replacement registration, and manifest/build gates directly affect wire compatibility and runtime subtype identity; CI changes alter TestU01 failure semantics and perf baseline promotion policy.
&gt; 
&gt; **Overview**
&gt; Consumer assemblies can **extend WallstopProto bases defined in referenced assemblies** when the upstream type ships a generated extension entry point (`TryReadWithSubtypes`). The generator emits **replacement formatters** that merge upstream dispatch metadata with local subtypes, registers them at startup, and **rejects two assemblies extending the same external base** at compile time, in subtype tag assignment, and in the pre-build gate. Tag planning now inventories cross-assembly bases, pulls upstream reserved/dispatch field numbers, and blocks manifest writes on conflicts; **WPROTO040** / **WPROTO044** describe missing extension bodies instead of blanket same-assembly refusal, with broad generator tests.
&gt; 
&gt; **Compile cost:** generic protobuf/JSON closure registration shares one syntax scan (`ClosureScan.Types`) across marshal and JSON maps; formatter symbols go through **`WProtoGeneratedNames`**.
&gt; 
&gt; **CI and hooks:** pre-commit meta rules skip nested `~` tooling paths under `Samples~`; **TestU01** fails when the driver exits non-zero; PractRand comparison adds a **recoverable** output so auto-closed mismatch issues require full-inventory recovery; Unity jobs only run log dumps/redaction after **successful checkout**; scheduled benchmarks **stop auto-promoting** `baseline.json`; **`unity-tests`** adds manual **native acceptance** (`sentinel` / `intmap` / `serialization` / `all`) with Il2Cpp provisioning when needed.
&gt; 
&gt; **Sentinel / editor:** validation rule preferences apply via shared **`ApplyPreferences`**; the rule builder shows fix-value fields only for rename/import-size fixes; graph connection mesh **triangle index order** is fixed; **`ValidationWindow.uss`** is reformatted with `.dx-editor` layout scoping. **`ShowIfConditionEvaluator`** routes conversions through internal **`TryConvertValue`**.
&gt; 
&gt; <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26f23b6f63497053d18371a1a4ac1b5c2d4ece65. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/Tools/WProtoSubtypeTagAssigner.cs
```diff
@@ -449,6 +475,42 @@ private static void AddReserved(Inventory inventory, Type baseType, Report repor
...
+                        baseType.Assembly
+                    )
+                )
+                {
+                    if (held.BaseTypeName == baseName)
+                    {
+                        inventory.Reserved.Add(held);
+                    }
+                }
+                Type formatter = baseType.GetNestedType("WProtoFormatter", BindingFlags.Public);
```
wallstop: Should any of these be const, named variables, in a single common place?

wallstop: Yes. Fixed locally in `15c4a53c`: `WProtoGeneratedNames` owns the formatter/root-formatter and subtype read/write/measure names. The editor uses the runtime definition; the generator source-links the same file, so it adds no runtime assembly dependency.

Sweep: both editor lookups, generator semantic discovery, emitted declarations/calls and generated file names now share those constants. Rule: added this emission/discovery pattern to the existing magic-string guidance.

Validation: all 153 generated files are byte-identical; complete Release suites pass (v3 762/762, v2 761/761); the full source matrix and staged preflight pass. I’m batching this with the TestU01 repair and waiting for the current native CI tail to finish before pushing, so the commit is not on the remote yet.

wallstop: Published in [15c4a53c](https://github.com/Ambiguous-Interactive/unity-helpers/commit/15c4a53cfeceae9605a01b2ed0217c30a7ba599d). The fix, sweep, and shared-source rule described above are now on this PR. The previous head’s complete normal Unity matrix and package export passed before the push; CI for this new head is running.

Runtime/Core/Random/AbstractRandom.cs
```diff
@@ -792,14 +798,22 @@ private double SampleDoubleWithInfiniteRange(double min, double max, out bool sa
 
             ulong range = orderedMax - orderedMin;
             int attempts = 0;
+            bool allDrawsUnbiased = true;
             while (true)
             {
-                ulong sample = orderedMin + NextUlong(range);
+                ulong offset = SampleUlongBelow(range, out bool unbiased);
+                if (requireExact && !unbiased)
```
wallstop, line 803: Why this? Why not Try* with bool return and an out double? double return with out bool seems like an anti-pattern - here and elsewhere.

wallstop: Agreed. Fixed locally: all four shared sampling cores now use `TrySample*`, return `bool`, and write the sample through a typed `out` parameter. The public `TryNext*` methods still clear failed outputs to default; compatibility methods retain their documented degraded values and draw counts.

Sweep: no `out bool` remains in `Runtime/Core/Random`. The same failure-reporting shape existed in `ShowIfConditionEvaluator.ConvertValue`; its implementation and internal callers now use a boolean-returning helper. The previously shipped public signature remains as a thin compatibility wrapper. Other `out bool` uses report a boolean value or metadata, rather than success.

Rule: the defensive-programming guidance now states this return/output convention explicitly. Validation so far: all 33 bounded-contract host cases pass and all seven cap/endpoint mutants fail as expected. Full compilation/native follow-up is pending; these edits are not published yet.

.githooks/pre-commit.ps1
```diff
@@ -505,6 +505,7 @@ function Invoke-MetaChecks {
     $sourceRootPattern = '^(Runtime|Editor|Tests|Samples~|Shaders|Styles|URP|docs|scripts)/'
     $metaRequired = @($StagedPaths | Where-Object {
         $_ -match $sourceRootPattern -and
+        $_ -notmatch '(^|/)[^/]*~(/|$)' -and
```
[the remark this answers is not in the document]
wallstop: Confirmed. The blanket exemption confused ignored package contents with assets that Package Manager later imports into Assets. Fixed locally in the hook, agent preflight, and metadata linter: children of the root Samples~ directory require committed metadata; the container itself and nested ignored tooling directories remain exempt.

Sweep/rule: corrected the existing create-unity-meta guidance and its contradictory sample exemption. The updated regression fails against the previous linter and passes with the fix (43/43); the repository metadata scan passes. Added sample paths to the real preflight detection/auto-fix cases and retained actual hook controls for ignored tooling and ordinary directories. Full hook regressions are running before publication.

Runtime/Core/Random/AbstractRandom.cs
wallstop: Published in [0c207e6d](https://github.com/Ambiguous-Interactive/unity-helpers/commit/0c207e6d): the bool-return fix, repository sweep and defensive-programming rule are now on this PR. Full source/define compilation passed; all 33 bounded host cases, 14 ShowIf conversion/comparison controls and both complete Release protobuf suites passed (762 v3 / 761 v2, no skips). All seven sampling mutants were rejected. The 80 raw-stream baseline cells also match main exactly. Native CI for the new head is running.

.githooks/pre-commit.ps1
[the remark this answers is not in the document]
wallstop: Published in [78c973db](https://github.com/Ambiguous-Interactive/unity-helpers/commit/78c973db). Fix, sweep and metadata guidance are now on the PR. The complete hook regression aggregate passed: 150 agent-preflight, 21 pre-commit and 49 pre-push controls, with no failures; the metadata exclusion suite passed 43/43 and the repository metadata scan passed. Imported sample contents require metadata, while the Samples~ container and nested ignored tooling remain exempt.

scripts/unity/run-acceptance.ps1
```diff
@@ -0,0 +1,175 @@
...
+        'WallstopStudios.UnityHelpers.Tests.Core', $upstreamNamespace).Replace(
+        'WallstopStudios.UnityHelpers.Tests.Serialization', $consumerNamespace)
+    $consumer = $consumer -replace '(?m)^\s*(?:using NUnit\.Framework;|\[(?:TestFixture|Test|Category\("[^"]+"\))\])\s*\r?\n', ''
+    foreach ($assembly in @('Upstream', 'Consumer')) {
+        $directory = Join-Path $Project "Assets/SerializationAcceptance/$assembly"
+        New-Item -ItemType Directory -Force -Path $directory | Out-Null
+        $references = @('WallstopStudios.UnityHelpers')
+        if ($assembly -eq 'Consumer') { $references += $upstreamNamespace }
+        @{ name = "WallstopStudios.UnityHelpers.Acceptance.$assembly"; references = $references; autoReferenced = $true } |
+            ConvertTo-Json -Depth 5 | Set-Content -LiteralPath (Join-Path $directory "$assembly.asmdef")
```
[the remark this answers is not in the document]
wallstop: I could not reproduce this with the actual helper on PowerShell 7.5.3. Its generated Upstream.asmdef contains:

```json
"references": [
  "WallstopStudios.UnityHelpers"
]
```

The helper pipes a hashtable to `ConvertTo-Json`; its `references` value is already an array created with `@(...)`. A singleton array passed as the pipeline input itself would be a different case. Consumer likewise emits a two-element array.

No production change is needed for the reported shape. I added an explicit array-type check to the existing emitted-asmdef controls locally: the 42 acceptance controls pass, and a seeded change to `references = $references[0]` fails with the expected JSON-array assertion. The native campaign remains queued/running independently; I am not claiming that the host check proves Unity execution.

.github/workflows/unity-benchmarks.yml
```diff
@@ -224,6 +224,7 @@ jobs:
         run: git config --global core.longpaths true
 
       - name: Checkout
+        id: checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
```
wallstop, line 228: Should any of these checkout options use a low depth, for faster checkout? Or does that not matter (here and elsewhere)

wallstop: Here it is already shallow: the [exact pinned checkout action](https://github.com/actions/checkout/blob/3d3c42e5aac5ba805825da76410c181273ba90b1/action.yml) defaults to `fetch-depth: 1`. Adding `1` explicitly would not reduce the fetch.

Sweep at `c441d081`:

| Checkout scope | Effective depth |
| --- | --- |
| All three benchmark steps; ordinary/SINGLE_THREADED Unity tests | 1 already |
| All workflows combined | 19 shallow steps; 20 full-history steps |
| License-year checks, multiple central-policy pins, MkDocs creation/revision dates | Have identified history consumers |

[Native job 101381317328](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/33994067162/job/101381317328) also confirms the actual `git fetch ... --depth=1`; fetching began at 21:55:50.909Z and checkout began at 21:55:51.841Z. That is one observed fetch, not a claimed speedup.

Some other depth-0 steps are candidates for improvement. I recorded their complete workflow groups and the required shallow-checkout/write-back measurements on #705 so they survive this session's wrap-up. No code change is needed here. The rule is to keep the shallow default unless a measured history/tag consumer requires more; we should not replace the known full-history consumers blindly.

Tests/Editor/Capture/EditorSurfaceCaptureHostWindow.cs
```diff
@@ -54,6 +84,13 @@ internal static void CloseHost(EditorSurfaceCaptureHostWindow window)
             }
 
             window.rootVisualElement.Clear();
+            if (window._ownedPanel != null)
+            {
+                window._ownedPanel.Dispose();
+                window._ownedPanel = null;
+                Object.DestroyImmediate(window); // UNH-SUPPRESS: batch host has no native parent to close.
+                return;
```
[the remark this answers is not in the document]
wallstop: Confirmed and fixed locally. `CloseHost` now clears ownership first, isolates root clearing and panel disposal, and always attempts host destruction. Cleanup exceptions are logged so a real failure remains visible to the test runner without aborting later cleanup.

Sweep: the pre-existing root `Clear()` and native-window destruction fallback could also escape; both are guarded. A compiled probe of the actual host implementation with explicit Unity stubs reproduced the old clear-path escape; normal, clear-failure, disposal-failure and destruction-failure controls now pass. Added a native data-driven regression for normal/throwing panel disposal, asserting disposal was attempted and the host is destroyed.

Rule: retained the helper's documented no-throw teardown contract and visible failure logging. Compilation/preflight and publication are pending. I am leaving the current native capture run intact so its rendering evidence can finish before the next publication.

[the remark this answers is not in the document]
wallstop: Published in c700ddb7, included in head 26f23b6f. The fix and sweep isolate root clearing, panel disposal and host destruction, retaining visible exception logging and the no-throw teardown contract. The actual-source host probe passes all four normal/fault controls; the previous implementation fails each of the three fault controls. Editor test compilation and staged preflight pass. The native disposal regression is included in the new CI run; native validation remains pending.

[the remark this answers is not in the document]
wallstop: Native verification is now available for head `26f23b6f`: [Unity 6000.3 job 101408982219](https://github.com/Ambiguous-Interactive/unity-helpers/actions/runs/34004425896/job/101408982219) passed EditMode, PlayMode, standalone and cleanup. The final EditMode XML has both `HostDestructionContinuesAfterPanelDisposal(False/True)` cases and `BatchHostAttachesAnEditorPanelWithoutOpeningAPopup` marked Passed. Downloaded artifact 9980743995 matches GitHub SHA-256 `d964878747c20378bca7610aa33ef97cb78ce0544baed054528024cef5670844`. This completes native regression validation of the cleanup fix; the other versions and separate Sentinel visual acceptance remain in progress.

PR #726, opened 2026-09-06
Keep extreme spatial entries queryable

### LINKED ISSUE #720: Finite extreme coordinates overflow spatial node centers
## Reproduction

Verified against main `6d1b428cdcf54aaed0d1d4b59e6f2457c33c7888` in Unity 6000.4.6f1 through the Unity MCP bridge, with the unmodified `KDTree2D.cs` restored from that commit for the controlled baseline.

```csharp
Vector2 point = new Vector2(float.MaxValue, 0f);
KdTree2D<Vector2> tree = new KdTree2D<Vector2>(
    new[] { point, point }, value => value, bucketSize: 32, balanced: true);
List<Vector2> results = new List<Vector2>();
tree.GetElementsInRange(point, 1f, results);
```

Expected: both coincident finite points are returned. Actual: zero results. `Boundary.center.x` is positive infinity even though all input coordinates are finite; `Boundary.extents.x` is approximately 0.0005.

## Root cause and controls

`KdTree2D.CreateBounds` calculates `(min + max) * 0.5f`, overflowing before division. The aggregate becomes unusable for finite queries. This is separate from #718, whose ingestion filter excludes explicitly nonfinite stored geometry; that filter cannot repair arithmetic over accepted finite coordinates.

The probe selected the exact four-parameter range method (including optional `minimumRange`), invoked it five times, and seeded the destination with a sentinel to prove a real query occurred. Controls: `[zero,(2,0)]` returns one; `[zero,float.MaxValue]` returns one; `[zero,-float.MaxValue,+float.MaxValue]` returns one at the origin. The coincident-max case returns zero. An earlier three-parameter reflection probe did not invoke the API and is explicitly discarded.

## Acceptance

- Audit bounds creation, child aggregation and minimum-size expansion across all eight immutable spatial variants and both QuadTree constructors.
- Preserve finite entries when intermediate center/extent/size arithmetic overflows; avoid solving this by silently excluding finite inputs.
- Cover same-sign extremes, opposite-sign extremes, coincident points, neighboring representable values, each axis/sign, leaf and split trees, plus R-tree finite edges.
- Verify radius, bounds and nearest queries against independent finite-value oracles, retaining original source identity and tie order.
- Preserve documented positive-infinite query radii and explicit infinite query bounds.
- Document any unavoidable representability limitation precisely and add native Unity regressions across supported versions.

No implementation or passing acceptance is claimed by this issue. The current session keeps #718's nonfinite-input fix independently reviewable.

### LINKED ISSUE #722: Enum scalar factory accepts managed-reference structs
## Problem

At main `571256461f0d97935d313301828c6df1716fb506`, `WProtoScalarFormatters.Enum<T>(int size, bool signed)` accepts any `struct` whose storage size matches the supplied width, including a struct containing a managed reference. Its formatter later reinterprets that storage as integers. The cold registration boundary must prevent this invalid formatter from existing.

## Safe reproduction

```csharp
struct ReferenceSlot { public object Value; }
var formatter = WProtoScalarFormatters.Enum<ReferenceSlot>(IntPtr.Size, false);
```

The factory currently returns a formatter. Expected: reject the non-enum type. Do not execute a formatter read with this type: it would write integer bits into managed-reference storage.

A separate wire-contract failure is `Enum<SignedByteEnum>(1, false)`: it accepts incorrect signedness and encodes -1 as unsigned 255 instead of the signed int32 varint representation.

## Evidence

The shared Unity/host regression fixture was run against unchanged production source: three failures (reference-containing struct, primitive structs, incorrect signedness), six passing controls. This is a concrete registration defect discovered during #637; it does not replace that issue's broader unsafe-free and player-performance acceptance.

## Acceptance

- Reject every non-enum type before constructing a memory-reinterpreting formatter.
- Validate declared width and signedness against all eight actual enum underlyings.
- Preserve valid registration and wire bytes for defaults, boundaries, aliases, flags and undefined values.
- Sweep all public scalar-factory construction paths for the same unchecked type/layout assumption.
- Run the shared regression against both protobuf-net host configurations and native Unity; retain all applicable AOT/typecheck checks.
- Update API and serialization documentation; leave #637's remaining unsafe-memory removal and measurements explicitly open.

### LINKED ISSUE #723: Singleton reset callbacks can abort or recurse cleanup
## Problem

At main `571256461f0d97935d313301828c6df1716fb506`, `ScriptableObjectSingleton<T>.ClearInstance` invokes consumer `OnInstanceCleared` before resetting the cached lazy instance. No finally or reentry guard owns that transition. A callback exception prevents reset; a callback that clears the same singleton invokes itself recursively. Direct editor reset callers are affected even though the registry isolates exceptions between registered actions.

## Controlled reproduction

Use an existing loaded ScriptableObject singleton whose `OnInstanceCleared` callback calls `ClearInstance` once, throws, or does both. Bound the recursive callback in the regression itself to avoid a stack overflow. Expected: one callback, old live instance remains readable during the callback, reset completes and the next reset can succeed. Current source leaves the old cache after a throw and calls the callback twice on bounded reentry.

Reset also lacks a main-thread guard even though it invokes consumer Unity lifecycle code. A worker reset must be rejected before any callback or cache mutation, using the existing managed thread guard.

## Acceptance

- Exercise normal, reentrant, throwing and reentrant-throwing callbacks with independent callback-count/cache/asset-identity assertions.
- Complete reset despite consumer failure, log the exception once, retain the existing live instance while its callback runs, and allow a later reset.
- Ignore same-type nested reset without repeated callbacks; preserve other singleton types' reset behavior.
- Reject worker reset before Unity access, callbacks or cached-state mutation.
- Sweep equivalent lifecycle callback invocations and registry/direct editor entry points.
- Document the transition and run focused native Unity tests, including actual NUnit setup/teardown and log assertions.

Discovered while auditing #644. That broader relation/reflection/lifecycle matrix and paired player performance campaign remain open; this issue tracks the concrete cleanup defect without substituting for those requirements.

### LINKED ISSUE #724: Return serial license after partial Docker activation
## Observed main failure

Main commit `571256461f0d97935d313301828c6df1716fb506`, Unity run 34035631255, package-export job 101503866992: Unity reported entitlement activation, then licensing client error 1500 (an asynchronous operation timed out). No usable license artifact appeared. Export exited before compilation.

The container wrapper calls `require_license_artifact`, which exits early, before its normal `return_serial_license` call. Only INT/TERM have inner cleanup handlers; ordinary early exit has none. The container was removed without an in-container return attempt. The outside adapter correctly reported missing positive cleanup evidence, and central release quarantined reservation `ceea97ba-5481-4f58-97bb-edeaec8dc82c`.

## Acceptance

- After starting serial activation, every ordinary exit attempts return in that same container, including partial activation with no usable artifact.
- Preserve the failed activation outcome even when return succeeds; never classify partial activation as a successful export.
- Keep normal and signal return ownership intact, avoid duplicate completed return attempts, and retain bounded watchdogs.
- Add executable fault-injection cases for partial activation and successful/failed cleanup, alongside existing normal/timeout/signal tests.
- Do not clear the observed quarantine without central recovery evidence; successful future cleanup cannot prove the old reservation safe.
- Verify the repaired export lifecycle in CI and restore main's required aggregate to green.

This fixes a concrete wrapper omission. It does not provide the centrally trusted container identity/cleanup capability still tracked by #323 and build-lock #153.

wallstop:
**Why:** Extreme coordinates and invalid registrations could break queries or serialization, while failed callbacks and partial activation could interrupt cleanup.

**What:**

- Preserve finite geometry and nearest ordering across immutable spatial trees.
- Reject invalid enum registrations before accessing their storage.
- Complete singleton resets despite callback failure or reentry.
- Return serial licenses after partial Docker activation.

**Validation:** All CI passes, including four Unity versions, `SINGLE_THREADED`, package export, type checks and contract suites. NUnit artifacts verify all 25 singleton recovery cases in EditMode and all 2,140 spatial/enum cases in PlayMode and Standalone across the version matrix. Both host protobuf configurations passed locally.

Unity 6000.5 passed on one unchanged-head retry after a compiler crash; #727 retains the evidence and unresolved cause.

Fixes #720
Fixes #722
Fixes #723
Fixes #724

---

> [!NOTE]
> **Medium Risk**
> Spatial tree query semantics change for extreme coordinates and nearest-neighbor ordering; singleton clear behavior is more strict (main thread) but should reduce production footguns. Proto enum registration may throw where invalid shapes previously slipped through.
> 
> **Overview**
> Fixes **spatial indexing at float extremes** by centralizing overflow-safe bounds and distance math in `SpatialQueryMath` and routing KD/Quad/Oct/R trees and `BoundingBox3D` through it. Node enclosures use conservative `CreateConservativeBounds` / `Union` and double-precision squared distances for nearest-neighbor ordering; **QuadTree** child bounds are derived from actual entry min/max instead of geometric quadrant splits; **RTree3D** range queries measure distance to the original closed element edges. Docs and a large parameterized test suite document and lock in the behavior ([#720](https://github.com/Ambiguous-Interactive/unity-helpers/issues/720)).
> 
> **WallstopProto** `WProtoScalarFormatters.Enum` now validates that `T` is an enum and that `size`/`signed` match the underlying type before building a formatter, with new tests wired into the generator harness ([#722](https://github.com/Ambiguous-Interactive/unity-helpers/issues/722)).
> 
> **Singleton cache resets** are hardened: `ClearInstance` requires the main thread, queues cross-type clears to avoid nested destruction hazards, logs callback failures but always finishes resetting the cache, and **ScriptableObject** resets ignore reentrant clears ([#723](https://github.com/Ambiguous-Interactive/unity-helpers/issues/723)).
> 
> **Docker Unity wrapper** attempts Pro serial license return on every container exit once serial activation was tried—including failed/partial activation—plus shell tests for that path ([#724](https://github.com/Ambiguous-Interactive/unity-helpers/issues/724)).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac90635c7a6989fc602909c845b5a5958275e7a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Utils/ScriptableObjectSingleton.cs
```diff
@@ -91,38 +92,49 @@ private static string GetResourcesPath()
...
         /// <summary>
-        /// Clears the cached singleton instance, allowing a fresh load on next access.
-        /// If an instance was loaded, <see cref="OnInstanceCleared"/> is invoked before clearing.
+        /// Clears the singleton cache on the main thread, allowing a fresh load on next access.
         /// </summary>
         internal static void ClearInstance()
         {
-            // An unrealized Lazy may cache a factory exception, so always replace it on reset.
-            if (_lazyInstance.IsValueCreated)
+            UnityMainThreadGuard.EnsureMainThread();
```
wallstop, line 99: Should this apply to RuntimeSingleton as well?

wallstop: Yes. Fixed and swept in `ac90635c`. `RuntimeSingleton.ClearInstance` now checks the main thread before discovery and always resets its cache. The registry serializes nested clears and deduplicates pending/in-progress types, so synchronous destruction callbacks cannot reenter destruction of the same GameObject or its parent. Cleanup continues across failed objects and queued requests.

The native baseline reproduced eight failures. The final regression set passes all 13 runtime cleanup cases, including both parent/child orders, shared GameObjects, callback failure, queue ordering/reuse, inactive duplicates and worker calls. The combined recovery run passed 25 cases, and the existing runtime singleton PlayMode suite passed all 77.

The sweep also checked the registries and cache-only startup reset paths; those already snapshot/isolate callbacks or preserve authored objects without invoking consumer cleanup. The rule is now captured in native lifecycle regressions and the singleton/XML documentation. All compilation configurations, repository lint, preflight and strict docs pass locally; hosted CI is running on the update.

PR #731, opened 2026-09-06
Preserve runtime state across resets

### LINKED ISSUE #725: Redact Unity access-token previews from diagnostics
## Observed gap

Main Unity export job 101503866992 printed a Unity Licensing module line with this shape:

```text
[Licensing::Module] Successfully updated the access token SYNTHETIC-PREVIEW...
```

The value above is synthetic. No actual credential material is included here.

The Docker stream redactor currently handles serials and email addresses only. The shared artifact credential patterns cover bearer headers and environment assignments, but not this labelled Unity token-preview line. Consequently a token fragment survives both console output and the generic artifact pattern inventory.

## Acceptance

- Redact the value of Unity's labelled access-token messages, including truncated previews, while retaining useful diagnostic labels.
- Sweep all Docker, PowerShell, log-tail and uploaded-artifact paths for the same output shape.
- Add synthetic full-token, preview, already-redacted and no-value cases; demonstrate idempotence and no credential bytes in test/CLI output.
- Preserve failure diagnostics and existing serial/email/key scrubbing.

Discovered while investigating the licensing timeout in #724. This is separate from that issue's missing cleanup attempt; it must not require reproducing a real credential in a test or issue.

### LINKED ISSUE #728: Preserve cached Gaussian draws in UnityRandom snapshots
Found while completing the RNG continuation audit in #638.

`UnityRandom.InternalState` stores seed presence in `RandomState.Gaussian` (`0.0` for seeded, null for unseeded), replacing the actual cached Box–Muller partner. The snapshot constructor then calls `RestoreCommonState`, which treats that sentinel as a real cached Gaussian. Consequently copying/restoring a seeded generator can inject a zero, and restoring an unseeded generator loses its pending draw. Raw-word-only continuation tests miss this; mixed-copy coverage currently excludes UnityRandom.

Reproduced against main 41272030 in Unity MCP 6000.4.6f1, restoring the engine global in finally: initialize seed 4242 and draw one Gaussian (0.35835835636961971). The next expected draw is -0.59636735012838749. Restoring the seeded snapshot returns 0; restoring the unseeded adapter's snapshot returns 1.0145252102904454.

Acceptance:

- Preserve Gaussian cache, bit/byte buffers, engine position and seeded/unseeded metadata together through InternalState, the snapshot constructor and Copy.
- Exercise empty and populated Gaussian caches, including legitimate cached zero, and mixed raw/bit/byte/Gaussian continuation.
- Preserve behavior through JSON and protobuf/WallstopProto snapshot round trips where supported.
- Read legacy snapshots without fabricating a Gaussian from the former seed sentinel; document that a cache older snapshots discarded cannot be recovered.
- Sweep every RandomState constructor/restore path for metadata sharing a sampling-cache field. Preserve raw engine draws and isolate UnityEngine.Random global state in tests.
- Update XML/feature docs and release notes; run native Unity regressions and applicable CI configurations.

The broader performance/backend inventory in #638 remains separate.

### LINKED ISSUE #729: Exclude cleared singletons before deferred destruction
Found while auditing lifecycle transitions in #644.

In PlayMode, `RuntimeSingleton.ClearInstance()` schedules `Object.Destroy` and clears its static cache immediately. An `Instance` access in the same frame runs Unity discovery, which still sees the scheduled object, and returns the cleared instance again. Consumers receive stale authored state and can cache an object that will disappear at frame end. Existing fresh-instance coverage yields until destruction before querying, so it misses this interval.

Reproduced in Unity 6000.4.6f1 native PlayMode, run `44bcf441-caf0-491a-ad42-7009addda3a2`: `RuntimeSingletonTests.ClearInstanceAllowsFreshInstanceBeforeDeferredDestruction` failed (zero passes, one failure). Set the original singleton&#39;s testValue to 99, call ClearInstance, and immediately request Instance; `AreNotSame` fails because it returns the original.

Acceptance:

- Once cleared for destruction, a runtime singleton must never be returned or newly cached by Instance/HasInstance, including the same frame before Unity processes Destroy.
- Preserve authored-instance discovery and NeverCreate policy, including remaining valid duplicates and repeated clear/replacement cycles within a frame.
- Preserve callback/reentrancy ordering, thread guards, cache-only startup reset behavior and immediate EditMode cleanup.
- Keep ordinary lookup inexpensive and ensure pending-destruction tracking cannot retain dead objects indefinitely.
- Sweep runtime discovery and registry reset paths for the same state transition; add native PlayMode regressions and retain existing lifecycle coverage.
- Update XML/feature docs and release notes; pass relevant typecheck configurations and Unity CI.

The complete relational/reflection/performance audit in #644 remains separate.

wallstop:
**Why:** Saving random state and clearing singletons could return incorrect runtime data.

**What:**

- Preserve Gaussian caches in UnityRandom snapshots and copies.
- Exclude retired singleton components from discovery and creation.
- Redact Unity token previews across diagnostic output paths.

Fixes #725
Fixes #728
Fixes #729

---

> [!NOTE]
> **Medium Risk**
> Runtime singleton lookup and `UnityRandom` snapshot semantics changed in ways that affect play-mode lifecycle and saved random state; CI changes are additive scrubbing but touch many failure-output paths.
> 
> **Overview**
> **Runtime correctness:** `UnityRandom` snapshots and copies now keep Gaussian, bit, and byte caches (including a legitimate zero Gaussian), use `state2` to mark seeded vs unseeded snapshots, and migrate legacy saves that misused the Gaussian slot as a seed flag.
> 
> **Runtime singletons:** Cleared or duplicate-retired components are marked pending destruction, drop out of `HasInstance` and scene discovery immediately, and child singletons on the same hierarchy are retired together. `Instance` can spawn a replacement after `ClearInstance()` before deferred Play Mode destruction, while synchronous clear callbacks cannot create replacements; `NeverCreate` no longer rediscovers pending instances.
> 
> **CI security:** Shared `ConvertTo-UnitySafeLogText` / artifact patterns scrub Unity access-token previews (plus serials and emails) from GitHub diagnostic actions, PowerShell Unity runners, Docker license output, and failure annotations—without stripping real error text—with regression tests wired through the redaction harness.
> 
> **Docs/tests:** Changelog and guides reflect #728/#729; `ScriptableObjectSingleton` tests/docs clarify cache-only `ClearInstance()` keeps the asset alive.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9a09e65aacec042edc5734d984062a8dd58e95d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Utils/RuntimeSingleton.cs
```diff
@@ -43,13 +43,13 @@ namespace WallstopStudios.UnityHelpers.Utils
...
     [DisallowMultipleComponent]
-    public abstract class RuntimeSingleton<T> : RuntimeSingletonBase
+    public abstract class RuntimeSingleton<T> : RuntimeSingletonBase, IRuntimeSingletonLifecycle
         where T : RuntimeSingleton<T>
     {
         /// <summary>
         /// Gets a value indicating whether an instance is currently assigned.
         /// </summary>
-        public static bool HasInstance => _instance != null;
+        public static bool HasInstance => _instance != null && !_instance._isPendingDestruction;
```
wallstop, line 52: Does this apply to the scriptable object singleton version as well?

wallstop: I’m sweeping that implementation too. Its clear path resets the lazy cache and invokes `OnInstanceCleared`; it does not schedule destruction of the asset, so reloading the same still-live asset is intentional. `HasInstance` already checks Unity’s native null semantics. I’m verifying the cache-clear, destroyed-reference and callback cases in the native Unity fixtures before giving the final conclusion here.

wallstop: Confirmed in native Unity: this pending-destruction fix is specific to `RuntimeSingleton`. `ScriptableObjectSingleton.ClearInstance` releases the lazy cache without destroying the asset. Immediately afterward `HasInstance` is false, and the next lookup can correctly return the same live asset. Once Unity actually destroys an asset, the existing native-null check excludes it and the next main-thread lookup repairs the cache.

Sweep: inspected both ScriptableObject reset/registry paths and all discovery/cache reads; there is no package-owned destruction to mark. Native run `81903095-44a1-475a-8e3a-102077e7a115` passed 14 cases, zero failures/skips: live Resources asset preservation/reload, replacement asset reload, destroyed-reference recovery, worker/shutdown guards, failed factories, and callback reentrancy/cycles/exceptions.

Fix/rule: renamed the misleading `ClearInstanceDestroysLoadedAsset` test and strengthened it to require exactly one callback, immediate cache release, and the same live asset before and after the next frame. The guide now explains the two reset contracts. Independent review is clean; these test/documentation changes are being added to this PR. No ScriptableObject production change is needed.

PR #735, opened 2026-09-07
Preserve exact component bindings

### LINKED ISSUE #730: Handle Git-owned locks during partial commits
A supported `git commit --only -- <paths>` fails in the pre-commit lock guard even without a competing Git process. Ordinary staged commits succeed.

Observed while preparing the runtime fixes on main base `4127203025b35f9df7104d0094df2319bde82868`: the hook waited 10 seconds, reported `index.lock still held after waiting`, and aborted. Selecting the intended staged files before a normal commit succeeded without bypassing hooks or deleting locks.

An isolated temporary repository reproduced this with the current `scripts/git-staging-helpers.sh`: initialize two tracked files, modify and stage both, install a pre-commit hook that sources the helper and calls `ensure_no_index_lock 100`, then commit only one path. The partial commit exits 1; a normal commit exits 0. During the partial commit Git sets `GIT_INDEX_FILE` to `.git/next-index-<pid>.lock` while holding `.git/index.lock` itself. `get_index_lock_path` always selects the latter and waits for the parent process to release it.

Expected: valid partial commits can run the real hook checks while unrelated index writers remain protected. Do not delete the parent's lock or disable hooks.

Acceptance: add actual Git partial-commit regression coverage; handle the effective index and parent-owned lock safely across the Bash/PowerShell staging helpers; retain genuine external-lock contention and ordinary staged-commit coverage. Verify the complete hook-regression suite. This is a follow-up because normal staged commits provide a safe working path and the active PR's scope is runtime state and diagnostic redaction.

### LINKED ISSUE #733: Respect exact concrete relational component matching
Found while completing the relational audit on main c7bc16dabec729b509cefaadf96deeee9cc9c9fe.

When AllowInterfaces=false, the attribute documentation promises exact concrete matching. Metadata instead treats nonsealed component classes as interface-like, so an exact concrete match can be refused; the Component shortcut can also accept derived types. This affects sibling, parent and child resolution and scalar/array/list/set destinations. A typical reproduction is a nonsealed concrete component field with AllowInterfaces=false alongside an instance of that exact type and a derived instance earlier in traversal order.

The fix must distinguish interfaces from nonsealed classes, apply exact runtime-type filtering before single-result selection or collection truncation, and disable shortcuts whose broad Component result loses that distinction. Preserve documented hierarchy order, inactive filtering, skip-self semantics, warm/cold metadata behavior and the existing AllowInterfaces=true contract.

Acceptance:
- Direct Unity-query oracle tests cover all three relations and four cardinalities, concrete/base/interface/Component requests, derived-first ordering, inactive state, scope options and repeated hierarchy/cache changes.
- Invalid or absent matches retain the documented optional/required behavior; no unrelated API behavior changes.
- Default, legacy, Odin and SINGLE_THREADED builds pass, and real Unity EditMode/PlayMode fixtures pass across supported CI versions, including the applicable standalone lane.
- Update XML/feature/migration documentation and CHANGELOG; complete independent adversarial review.

A source fix and oracle fixture are in progress in the current worktree. No native execution result is claimed yet because the Unity MCP bridge is timing out. This focused defect does not close the parent audit's backend/performance and complete-shape evidence requirements.

### LINKED ISSUE #734: Release owned editor resources on refresh and close
The remaining resource audit on main c7bc16dabec729b509cefaadf96deeee9cc9c9fe confirms several owner-lifetime defects originally inventoried in #648.

- FitTextureSizeWindow creates a SerializedObject in OnEnable without disposing it on disable/rebind.
- ScriptableSpriteAtlasEditor clears cached SerializedObjects on project changes without disposal, has no close cleanup, and retains foldout keys for removed configs.
- AnimationEventEditor clears a cache mixing Unity-owned AssetPreview textures and copied textures without destroying its owned copies. CopyTexture also leaks an allocated texture if pixel access throws.
- Short-lived SerializedObjects are undisposed in atlas edits, sprite reference replacement, buffer settings, inline-inspector fallback analysis and dictionary comparison.
- Both collection drawers' pending-wrapper cleanup destroys the wrapper but drops its SerializedObject undisposed; failed initialization can lose newly created local resources before pending ownership is recorded.

Acceptance:
- Dispose owned SerializedObjects/properties before releasing their targets, both on normal cleanup and failed construction/initialization. Keep cleanup idempotent for dead targets and repeated clear/close.
- Preserve borrowed editor SerializedObjects and Unity-owned previews. Destroy only copied textures, including on failed copies, clip refresh and window disable.
- Preserve retained atlas results/foldout state for live configs while releasing removed entries; repeated project changes and reopen cycles must remain valid.
- Sweep package Editor sources for this ownership class and add focused behavioral regressions for success, failure, refresh, close and already-destroyed resources.
- Real Unity lifecycle tests and relevant host/editor gates pass; document resource behavior and user-visible fixes and complete independent review.

Implementation and tests are in progress. Unity MCP execution is currently unavailable through repeated timeouts, so source inspection is not claimed as native verification. This child tracks resource ownership; cancellation atomicity, project-scale performance and the broader parent audit remain open.

wallstop:
**Why:** Exact component bindings failed, editor resources leaked, and partial commits stalled.

**What:**

- Respect exact component types across relational bindings.
- Release editor resources during refresh, replacement, and close.
- Add configurable protobuf wire limits.
- Pin Unity RNG streams to frozen baselines.
- Handle partial-commit indexes in Git staging helpers.

Fixes #730
Fixes #733
Fixes #734

---

> [!NOTE]
> **Medium Risk**
> Touches widely used relational attributes (binding semantics change under `AllowInterfaces = false`), deserialization limits on untrusted input, and many editor code paths; behavior is heavily tested but upgrades may need field audits.
> 
> **Overview**
> **Exact relational binding** when `AllowInterfaces` is false now requires the component’s **runtime type** to match the field type, including non-sealed concrete types (derived components and polymorphic `Component`/`MonoBehaviour` fields no longer match). Shared EditMode/PlayMode oracles document the behavior.
> 
> **Editor lifecycle fixes** dispose owned `SerializedObject`/`SerializedProperty` and pending collection-drawer wrappers on failure, refresh, and close; atlas, fit-texture, and animation-event windows release cached serialized state and **only destroy copied sprite preview textures**, not Unity-owned previews.
> 
> **WallstopProto** adds immutable **`WProtoReadLimits`** (message size, length-delimited fields, tag count, nesting) with a public parameterless constructor for `new()` factories, wires them through **`WProtoReader`** and **`WProtoFacade.TryDeserialize`**, and documents hostile-payload handling.
> 
> **RNG regression**: frozen SHA-256 baselines for raw streams across generators, plus **`NativePcgRandom`** struct-copy and **`UnityRandom`** adapter stream checks.
> 
> Supporting updates: Git staging helpers resolve the **effective index lock** for `GIT_INDEX_FILE`/partial commits; changelog/docs/skills and test-count messaging move to **14,000+** tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81bd71fe1fd74c9330c4343e6ffc84a45bdbedb1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Runtime/Core/Serialization/WallstopProto/WProtoReadLimits.cs
+++ b/Runtime/Core/Serialization/WallstopProto/WProtoReadLimits.cs
@@ -1,0 +1,50 @@
+// MIT License - Copyright (c) 2026 wallstop
+// Full license text: https://github.com/wallstop/unity-helpers/blob/main/LICENSE
+
+namespace WallstopStudios.UnityHelpers.Core.Serialization.WallstopProto
+{
+    /// <summary>Configures immutable wire-size, field-count, and recursion limits for a read.</summary>
+    /// <remarks>
+    /// Limits apply to each reader region, including merged sub-messages. Field counts include
+    /// unknown and duplicate tags and group terminators, but not untagged packed elements.
+    /// These limits bound encoded input, not the size of objects a custom formatter constructs.
+    /// </remarks>
+    public sealed class WProtoReadLimits
+    {
+        internal static readonly WProtoReadLimits Default = new WProtoReadLimits();
+
+        /// <summary>Creates limits, treating negative values as zero and capping nesting at 64.</summary>
+        /// <param name="maximumMessageBytes">Maximum bytes in a root or nested reader region.</param>
+        /// <param name="maximumLengthDelimitedBytes">Maximum bytes in any length-delimited field.</param>
+        /// <param name="maximumFieldCount">Maximum tags read in one region, including skipped groups.</param>
+        /// <param name="maximumNestingDepth">Maximum combined sub-message and group depth.</param>
+        public WProtoReadLimits(
+            int maximumMessageBytes = int.MaxValue,
+            int maximumLengthDelimitedBytes = int.MaxValue,
+            int maximumFieldCount = int.MaxValue,
+            int maximumNestingDepth = WProtoReader.MaxNestingDepth
+        )
+        {
+            MaximumMessageBytes = maximumMessageBytes < 0 ? 0 : maximumMessageBytes;
+            MaximumLengthDelimitedBytes =
+                maximumLengthDelimitedBytes < 0 ? 0 : maximumLengthDelimitedBytes;
+            MaximumFieldCount = maximumFieldCount < 0 ? 0 : maximumFieldCount;
+            MaximumNestingDepth =
+                maximumNestingDepth < 0 ? 0
+                : WProtoReader.MaxNestingDepth < maximumNestingDepth ? WProtoReader.MaxNestingDepth
+                : maximumNestingDepth;
+        }
+
+        /// <summary>Maximum encoded bytes in a root or nested reader region.</summary>
+        public int MaximumMessageBytes { get; }
+
+        /// <summary>Maximum bytes in a length-delimited field, including unknown and packed fields.</summary>
+        public int MaximumLengthDelimitedBytes { get; }
+
+        /// <summary>Maximum tags in one reader region, including duplicate and unknown tags.</summary>
+        public int MaximumFieldCount { get; }
+
+        /// <summary>Maximum combined sub-message and group depth, never above 64.</summary>
+        public int MaximumNestingDepth { get; }
+    }
+}
--- a/Runtime/Core/Serialization/WallstopProto/WProtoReadLimits.cs
+++ b/Runtime/Core/Serialization/WallstopProto/WProtoReadLimits.cs
@@ -13,6 +13,10 @@ namespace WallstopStudios.UnityHelpers.Core.Serialization.WallstopProto
     {
         internal static readonly WProtoReadLimits Default = new WProtoReadLimits();
 
+        /// <summary>Creates default wire limits with nesting bounded to 64.</summary>
+        public WProtoReadLimits()
+            : this(int.MaxValue, int.MaxValue, int.MaxValue, WProtoReader.MaxNestingDepth) { }
+
         /// <summary>Creates limits, treating negative values as zero and capping nesting at 64.</summary>
         /// <param name="maximumMessageBytes">Maximum bytes in a root or nested reader region.</param>
         /// <param name="maximumLengthDelimitedBytes">Maximum bytes in any length-delimited field.</param>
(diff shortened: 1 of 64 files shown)


Runtime/Core/Serialization/WallstopProto/WProtoReadLimits.cs
```diff
@@ -0,0 +1,50 @@
...
+    public sealed class WProtoReadLimits
+    {
+        internal static readonly WProtoReadLimits Default = new WProtoReadLimits();
+
+        /// <summary>Creates limits, treating negative values as zero and capping nesting at 64.</summary>
+        /// <param name="maximumMessageBytes">Maximum bytes in a root or nested reader region.</param>
+        /// <param name="maximumLengthDelimitedBytes">Maximum bytes in any length-delimited field.</param>
+        /// <param name="maximumFieldCount">Maximum tags read in one region, including skipped groups.</param>
+        /// <param name="maximumNestingDepth">Maximum combined sub-message and group depth.</param>
+        public WProtoReadLimits(
```
wallstop, line 25: Can we make a builder or something, this technically means there is no default constructor. A default constructor should exist, it should just call the ... correct one with a bunch of parameters.

wallstop: Fixed in f6ed8e7e.

- **Fix:** Added a real public parameterless constructor delegating to the existing four-argument constructor with the intended defaults. Named-argument customization remains available. The generic `new()` regression failed with CS0310 before the change; 62 focused host tests and all 45 native Unity limits tests now pass, with no skips or inconclusive results.
- **Sweep:** Reviewed 251 public runtime class groups across the relevant symbol configurations. No additional missing default constructors in options/limits/configuration objects; existing implicit constructors remain valid.
- **Rule:** Added a narrowly scoped default-construction rule to `.llm/skills/create-csharp-file.md`, including verification through a generic factory.

On the readonly-struct question: this profile remains a reusable immutable class. Current null/default-reader handling and profile-continuity checks use reference semantics. A struct would need distinct semantics for zero-initialized `default` versus explicitly zero limits; `in` would not eliminate copies stored in each reader. A builder is unnecessary for this constructor fix.

Issue #742, opened 2026-09-07
[Feature]: adopt general-purpose helpers that two first-party games keep re-writing


wallstop:
## What this is

DoxReloaded holds utility code that is not about DoxReloaded. We scanned the game for it. This issue lists what we found, ranked.

Every candidate below passes two tests. Its parameters and return use only primitives, `System.*` or `UnityEngine.*`. No game type appears. We also searched this package first, and we say what we searched.

Package read at `3.5.0`. We propose the API. We do not ask you to take our code.

## Ranked candidates

| # | Proposed API | Evidence | Searched |
| --- | --- | --- | --- |
| 1 | `DurableFile.TryWriteAllText/TryWriteAllBytes/WriteAllTextAsync/TryDelete` | **A second first-party game already has a hand copy.** The DoxReloaded file says so in its own doc comment: "Ported from the sibling IshoBoy project's `DurableFile`". Writes a sibling `.tmp`, flushes to disk, then swaps, so a broken write cannot truncate the live file | `Core/Helper/FileHelper.cs` has `InitializePath` and `CopyFileAsync` only. `PathHelper`, `DirectoryHelper` are path resolution. No atomic write |
| 2 | `BitOps.PopCount/TrailingZeroCount/HighestSetBit/Log2/NextPowerOfTwo` | 7 copies in the game. Two `NextPowerOfTwo` bodies are character-identical (`VisitedStateSet.cs:80`, `VisitedStateDepthMap.cs:143` — verified by `diff`). Two `PopCount` bodies are the same SWAR constants over `long` and `ulong` | **This package carries the same private copies**: `Core/DataStructure/BitSet.cs:603`, `ImmutableBitSet.cs:282`, and `IListExtensions.cs:3144/1069`. A public helper removes 4 duplicates here too |
| 3 | `WallMath.Median/Percentile/Mean/StandardDeviation` over `IReadOnlyList<T>` | **Six independent `Median` bodies** over four element types, plus `Percentile`, `Quantile`, two `Mean`, one `StandardDeviation`. Twelve definitions | `WallMath.cs`, `MathHelpers.cs`, `Geometry.cs`, `IEnumerableExtensions.cs`, `IListExtensions.cs`. The only `Median` here is a private pivot helper |
[...]

Issue #648, opened 2026-08-31
#633: profile and harden editor tooling at project scale


wallstop:
Parent: #633

## Outcome

Make editor utilities correct under Undo, prefab, reload, import, cancellation, and multi-object workflows, and keep large-project interaction responsive using measured incremental work and explicit resource lifetimes.

Scope: validators/analyzers, YAML/asset scanning, project windows and inspectors, batch fixers, import callbacks, caches, progress/cancellation, preview resources, and editor-only generators.

## Baseline workloads and contracts

Create deterministic synthetic projects at roughly 1k, 10k, and 100k assets with controlled dependency depth, matching/nonmatching distributions, malformed/unsupported text assets, packages/read-only files, and repeated GUID/fileID references.

For every tool, define:

- exact asset/file/object eligibility and deterministic result/order;
- whether scanning may load assets and the maximum loaded-object/retained-memory budget;
- Undo grouping, dirtying, prefab instance/asset override, multi-object, read-only/package, and scene behavior;
- cancellation and exception atomicity: no half-written batch without an explicit resumable transaction;
- import deferral/reentrancy and no recursive refresh loop;
- domain reload on/off, assembly reload, play-mode transition, window close/reopen, and compilation behavior;
- ownership/destruction of textures, serialized objects, streams, native collections, and callbacks.

### 1. Build a pure, deterministic scan core

- [ ] Separate discovery/parsing/analysis from UI and AssetDatabase mutation so it can consume a synthetic manifest/text stream in fast tests.
- [ ] Compare results with a simple full-scan oracle on every corpus, including duplicate references, malformed YAML, CRLF/LF, BOM, large files, and cancellation at every boundary.
- [ ] Use the shared YAML reader tracked by #631 rather than parallel parsers. Loading Unity objects merely to inspect metadata requires a measured justification.
- [ ] Record stable issue identities so incremental scans can add, change, delete, move, and reimport assets and converge exactly to a clean full scan.
- [ ] Add metamorphic tests: unrelated asset changes do not perturb results; order of import events does not change final state; full and incremental scans agree.

### 2. Prove mutation and lifecycle safety

- [ ] For each fixer, add red tests for Undo/Redo, prefab assets/instances, nested prefabs, multi-object editing, open/closed scenes, read-only/package assets, user cancellation, thrown callbacks, and assembly reload.
- [ ] Group Undo deliberately and call serialized/update/apply/dirty/save APIs only where their documented ownership requires it.
[...]

Issue #638, opened 2026-08-31
#633: make bounded RNG sampling exact and fail closed


wallstop:
Parent: #633

## Context

Several bounded/random-transform paths stop rejecting after a fixed number of attempts and silently return a different distribution:

- `AbstractRandom.NextUint(uint)` and `NextUlong(ulong)` fall back to modulo after their attempt caps.
- `NativePcgRandom.NextUint(uint)` has the same biased terminal branch.
- Infinite-range doubles, full-range doubles, and Gaussian generation eventually substitute a fixed or alternate result.
- `Next()` and `NextLong()` can return the documented exclusive maximum.

A stuck or adversarial source can therefore return an answer that looks valid while violating the API’s distribution contract. Statistical batteries cannot prove this arithmetic property. Lemire rejection must remain rejection: https://arxiv.org/abs/1805.10941

## Contract decisions

- Keep `IRandom` unchanged so external implementations do not break.
- Add bounded non-throwing methods to `AbstractRandom`: `TryNextUint`, `TryNextUlong`, `TryNextDouble`, and `TryNextGaussian`, each with the corresponding bounds/parameters and an `out` value.
- Invalid arguments or exhausted entropy return `false` and write `default`; a `Try` method never substitutes a biased, clamped, fixed, or alternate-distribution answer.
- Existing methods retain their current invalid-bound exceptions. For a valid request that exhausts the existing cap, they fail closed with documented `InvalidOperationException`.
- Trusted `NativePcgRandom` uses exact rejection without a modulo fallback.
- Preserve raw `NextUint()`/`NextUlong()` streams and power-of-two fast paths.
- Correct `Next()`/`NextLong()` to their documented half-open range. This is an intentional correctness change; raw streams and serialized continuation remain stable.
- Do not claim cryptographic suitability.

## Implementation plan

1. Add a scripted `AbstractRandom` test double that emits any chosen 32/64-bit sequence without allocating in the sampling loop.
2. Write red cases for constant rejected samples at both caps, acceptance on the final allowed draw, rejection one draw beyond it, all double/Gaussian fallback branches, and the signed exclusive maxima.
3. Implement one shared 32-bit and one shared 64-bit multiply-high rejection core used by both `Try` and throwing surfaces; do not duplicate the arithmetic.
4. Make higher-level floating/Gaussian samplers propagate the same explicit entropy-failure outcome.
5. Remove the `NativePcgRandom` modulo branch and correct signed half-open methods.
[...]

PR #746, opened 2026-09-08
Bound previews and release editor textures

wallstop:
**Why:** Editor texture caches retained memory, and callers could not pass runtime enum exclusions.

**What:**

- Bound preview caches and release owned editor textures.
- Accept enum exclusion arrays without changing `IRandom`.
- Verify mixed JSON continuation for all managed generators.
- Preserve binary files under `.github`.
- Incorporate verified dependency updates from #744 and #745.

Refs #648, #638, #742.

---

> [!NOTE]
> **Medium Risk**
> Editor changes touch native texture lifetime and domain-reload cleanup across multiple windows and shared GUI caches; incorrect ownership could cause leaks or destroyed previews, while the random extension is a small additive API surface.
> 
> **Overview**
> Introduces a bounded **`SpritePreviewCache`** (128 entries, ~32 MiB estimated texture budget) for **Animation Event Editor** and **Animation Creator**, replacing unbounded dictionaries. Owned preview copies are LRU-evicted and destroyed; Unity **`AssetPreview`** textures stay borrowed. When a readable sprite is too large to copy within budget, previews draw the **source texture crop** at full resolution instead of retaining another copy.
> 
> **Editor lifecycle fixes:** **`EditorCacheManager`** now clears caches on **`beforeAssemblyReload`** and editor quit, including new **`SolidButtonStyles.ClearCache()`** so collection-inspector button textures do not leak across script reloads. Animation event previews track **`resolvedSpriteTime`** and release via the shared cache on disable/close.
> 
> **Runtime API:** Adds **`RandomUtilities.NextEnumExcept`** as an **`IRandom`** extension accepting a runtime **`params`** exclusion array, delegating to existing fixed-arity overloads without changing the interface.
> 
> Also adds JSON mixed-continuation restore coverage for all generators, documents preview/cache behavior, tweaks **`.github/**`** in **`.gitattributes`** (`text=auto`), pins a few GitHub Actions, bumps **cspell**, and extends EOL checker tests for staged PNG bytes under those attributes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00999cb4068f32b63c34db15087d5bfde5299e74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/Sprites/AnimationCreatorWindow.cs
```diff
@@ -224,7 +224,7 @@ public sealed class AnimationCreatorWindow : EditorWindow
         private int _previewFrameIndex;
         private TimeSpan _lastPreviewTick;
         private bool _isPreviewPlaying;
-        private readonly Dictionary<Sprite, Texture2D> _previewTextureCache = new();
+        internal readonly SpritePreviewCache _previewTextureCache = new();
```
wallstop, line 227: Are we clearing these on destruction/disable, here and elsewhere? Are we preventing leaks?

Runtime/Core/Random/RandomUtilities.cs
```diff
@@ -9,6 +9,51 @@ namespace WallstopStudios.UnityHelpers.Core.Random
...
+        /// <summary>
+        /// Selects an enum value excluding a runtime-length array of values.
+        /// </summary>
+        /// <param name="random">The generator, or null to return the default enum value.</param>
+        /// <param name="exceptions">Excluded values; null or empty excludes nothing.</param>
+        /// <remarks>
+        /// Uses the generator's existing enum selection semantics, including its handling of
+        /// duplicate, undefined and fully excluded values. Passing an array does not copy it.
+        /// </remarks>
+        public static T NextEnumExcept<T>(this IRandom random, params T[] exceptions)
```
wallstop, line 22: Is there any danger of this taking higher precedence then the fixed arity methods? The reason the fixed arity methods exist is so that params doesn't allocate. I thought we had like, n-arity, then n-arity + params. Why is this better/worse?

.github/review-artifacts/sprite-preview-copy-baseline.png
wallstop, line 1: These files are corrupt/invalid - why are they checked in?

Runtime/Core/Random/RandomUtilities.cs
wallstop: The fixed-arity methods still win. I compiled calls against the actual package with C# 9 and inspected their IL:

- Interface and concrete calls with 1–4 literal exclusions bind to the existing instance overloads and emit no array allocation.
- Calls with 5 or 8 literals still bind to the existing four-plus-params instance overload and allocate the same one tail array as before.
- Passing an existing array (inferred or explicit generic type, interface or concrete receiver) binds to this extension with no new array. The extension itself contains no array allocation; above four exclusions it forwards the original array to the existing set-based overload, where duplicates are deduplicated.

This fills the runtime-length-array gap without changing `IRandom` or the fixed-arity fast paths. It is not a faster replacement for those paths. Calling `RandomUtilities.NextEnumExcept(random, A, B, C)` explicitly as a static params method can allocate; that was the positive control in the IL probe. Sweep/evidence: all these call forms were checked. No dispatch change was needed.

.github/review-artifacts/sprite-preview-copy-baseline.png
wallstop: You are right: the committed files are corrupt. I initially validated the working copies, which are 251-byte PNGs; `.github/** text eol=lf` removed the CR byte from their PNG signatures when Git staged them, leaving invalid 250-byte blobs.

Fix/sweep/rule, verified locally and staged for the combined review-fix push:

- Remove both diagnostic images and their PR-description table. They were added as before/after pixel evidence, but these tiny synthetic test outputs should not have been checked in. The real native pixel-comparison tests remain.
- Change the GitHub-directory attribute to `text=auto eol=lf`, preserving binary bytes while retaining LF scripts.
- Extend the existing EOL test with the actual repository attributes: staged PNG identity, checkout identity, and LF script behavior. The old rule fails both PNG checks; the correction passes all 24 tests.
- Sweep all 184 tracked binary assets: only these two PNGs were corrupt; all 182 remaining indexed assets have valid headers.

The regression checks the bytes Git actually stores and checks out, addressing the gap in my original verification.

Editor/Sprites/AnimationCreatorWindow.cs
wallstop: Yes. `AnimationCreatorWindow.OnDisable` clears this borrowed-preview cache before disposing serialized state. `AnimationEventEditor.OnDisable` and event refresh clear its mixed cache. Eviction/replacement/clear destroy owned copies, preserve Unity-owned textures, and event rows no longer retain preview texture references.

Sweep/fix/rule, verified locally and staged for the review-fix push:

- Native destruction/reopen tests confirm both windows empty their caches/accounting, destroy owned copies, and leave borrowed textures alive.
- Both retaining `AssetPreview` callers use the shared cache. The Odin inline preview is local-only; SpriteSheetExtractor disposes its per-sheet and failed-generation previews.
- The wider sweep found a separate `SolidButtonStyles` leak. Its owned normal/hover/active textures and dependent styles now clear through `EditorCacheManager` before assembly reload and on editor exit. Startup-only cleanup could no longer reach the old domain's native resources.
- The caching guidance now requires native static owners to release before domain unload, and window owners to clear on disable.

The leak regression failed before the fix. The final native lifecycle fixture passed 8/8, with a further passing rerun covering all three texture-liveness controls and failure-safe test cleanup. All four editor compile configurations pass. This verifies the audited cache lifetimes; it does not claim the whole editor/project-scale audit is complete.

Issue #661, opened 2026-08-31
The Copilot review check has been red on every PR since #238


wallstop:
Noticed while driving [#660](https://github.com/Ambiguous-Interactive/unity-helpers/pull/660) green.

## What is red

The **`Copilot`** workflow (`dynamic/agents/copilot-pull-request-reviewer`, id `295836001`) fails on every run. Its job `copilot-pull-request-reviewer` fails at a single step, **`Processing Request (Linux)`** — the bot's own request handling, before it looks at any diff.

Last eight runs, most recent first:

| Conclusion | Branch |
| --- | --- |
| failure | `dev/wallstop/session-240-correctness-sweep` |
| failure | `dev/wallstop/session-239-unity-footgun-analyzers` (x2) |
| failure | `session-238-wproto-subtype-enum-reservations-validation-window` (x5) |

Note there are **two** Copilot reviewer workflows registered. The older **`Copilot code review`** (`dynamic/copilot-pull-request-reviewer/copilot-pull-request-reviewer`, id `218725903`) has `success` on its last eight runs — but those are all from PRs #251 and #252, so it has not run in a long time. The one that runs now is the one that fails.

## Why it is worth fixing rather than ignoring

Sessions 238, 239 and 240 all shipped with this red. **A check that is always red trains everyone to skim the check list**, which is the only defence the repository has against a check that goes red for a real reason. It is also the reason `pr:feedback`'s "failing check runs" section is never empty, so that section stops being a signal.

Either the failure is real and someone should see the review it is not producing, or the workflow should be disabled until it works.

## Where to start

- The step name suggests an authentication or entitlement problem rather than anything about the diff — the same PR gets no review comment at all, so the bot never reaches the code.
- Whether the two registered workflows conflict is worth checking: `dynamic/` paths are GitHub-managed rather than tracked in this repository, so the fix is a repository or organization setting, not a file in `.github/workflows`.
- If it cannot be fixed, turn it off. A disabled check is honest; a permanently failing one is not.

## Acceptance criteria

- The Copilot review check is either green on a PR, or not registered at all.
- If it stays, one PR shows the review it produces, so the check means something.

wallstop:
The cause is in the review the bot *did* post, which the workflow log does not carry. `npm run pr:feedback -- 660` surfaces it:

```text
copilot-pull-request-reviewer[bot]  COMMENTED
  Copilot was unable to review this pull request because the user who
  requested the review has reached their quota limit.
```

[...]

PR #749, opened 2026-09-08
Stats, texture helpers, Jessesort merge, and #742 helpers

### LINKED ISSUE #747: Jessesort implementation is stale and credits the wrong person
### Page URL or Section

Runtime/Core/Extension/Sorting/IListSortJesse.cs

### Description of Issue

Hey, thanks for including Jessesort in the sorting benchmarks! It's cool to see it being used.

I looked at the implementation here and it seems to be based on a much older version of the algorithm. Jessesort changed quite a bit this month, especially the way it handles already sorted data, natural runs, pile storage, etc. I don't push elements back to piles one at a time anymore, I simulate the games and use blueprints.

I just want to make sure the implementation is current and see if those benchmark numbers are actually representative. The 181ms result for 100k integers seems off. I'd be happy to help bring the C# implementation up to date and rerun the benchmarks in Unity. Would you be open to a PR for that?

Also the Runtime/Core/Extension/Sorting/IListSortJesse.cs comments credit "Jesse Michel" but my name is Jesse Lew.

Thanks!

### Suggested Fix

Update implementation and author in docs

wallstop:
**Why:** Two games keep re-writing the same helpers, and PR review found one shipped defect along the way.

**What:**

- `WallMath` adds mean, median, percentile, standard deviation over `IReadOnlyList`. Arrays and lists skip interface dispatch. Measured: mean over an array is 1.7x faster.
- `Objects.Sha256Hex` and `TrySha256HexOfFile` hash text, bytes, and files. Text encodes through one shared UTF-8 helper.
- Jessesort merges piles pairwise. Every measured shape got faster; sorted input is now ~9x ahead of `Array.Sort`.
- `SpriteHelpers` adds `RotateTexture90`, `RotateTexture180`, `ExtractSpriteRect`. Also fixes `ColorExtensions` reading channel-swapped pixels from `ARGB32`/`BGRA32` textures.
- Editor windows share one `IsInvokedByTestRunner()`. `pr-feedback` prints human threads first.
- Pools: a writer's pool now releases its rented buffer on purge. Follow-ups are filed as #752, #753, #754, #755.

refs #742, fixes #747

---

> [!NOTE]
> **High Risk**
> Changes licensed Unity CI gates, triggers, and required-check aggregation (org-wide release risk), alongside public API additions and a serialization pool lifecycle fix.
> 
> **Overview**
> This PR tightens **licensed Unity CI** to an enrollment contract and ships several **runtime/editor helpers** plus bug fixes.
> 
> **CI / workflows:** Weekly cron triggers are removed from Unity Tests and Unity Benchmarks; licensed work is limited to protected-branch pushes and manual dispatch. Release and smoke/export paths **pin Unity `2022.3.45f1` literally** for editor ensure and license return, with **fail-closed drift checks** against `unity-versions.json`. Runner preflight drops the “no org credentials → skip matrix” path: fork/Dependabot guards stay at the job level, and availability checks run unconditionally so missing secrets **fail** instead of silently skipping. **Unity Tests** adds a static **`test-mode` matrix** (`editmode` / `playmode` / `standalone`) so each leg provisions the right editor profile (IL2CPP on standalone) and holds its own lock suffix. New **always-on aggregate jobs** enforce preflight + licensed job results so skipped/cancelled legs cannot vanish from required checks.
> 
> **Library:** `WallMath` gains **median, percentile, mean, and standard deviation** over `IReadOnlyList` (pooled sorted copies for order stats). **`Objects.Sha256Hex`** / **`TrySha256HexOfFile`** centralize hashing; sprite tooling uses the file helper. **`SpriteHelpers`** adds non-mutating **90°/180° rotation** and **sprite rect extraction**. **`ColorExtensions`** fixes channel-swapped averages on **ARGB32/BGRA32** by restricting the raw fast path to **RGBA32**. **JesseSort** replaces the heap k-way merge with **bottom-up adjacent run merging**. **Protobuf buffer pooling** registers **`onDisposal`** so purged writers return rented arrays. Editor windows share **`EditorUtilities.IsInvokedByTestRunner()`**.
> 
> Docs/skills add STE-simple copy rules and minor context updates; tests cover SHA-256, color formats, and the test-runner helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f3ff082464c934566a24dbaf86c422b58bfbd8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/Utils/EditorUtilities.cs
```diff
@@ -20,6 +20,35 @@ public static string GetCurrentPathOfProjectWindow()
...
+        /// </summary>
+        /// <returns>
+        /// True when any command-line argument mentions <c>runTests</c>, <c>testResults</c> or
+        /// <c>testPlatform</c>, case-insensitively.
+        /// </returns>
+        /// <remarks>
+        /// Test-runner launches open window <c>OnEnable</c> during the run itself, which can
+        /// consume fixtures or produce logs the test session never asked for.
+        /// </remarks>
+        internal static bool IsInvokedByTestRunner()
```
wallstop: Should this (and others) be public, for re-use of consumers?

Runtime/Core/Helper/Objects.cs
```diff
@@ -873,6 +876,97 @@ public static uint StableHash32V1(ReadOnlySpan<byte> bytes, uint seed)
...
+        /// </remarks>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="text"/> is null.</exception>
+        public static string Sha256Hex(string text)
+        {
+            if (text == null)
+            {
+                throw new ArgumentNullException(nameof(text));
+            }
+
+            return Sha256Hex(Encoding.UTF8.GetBytes(text));
```
wallstop: I think we have like a safe version of this - should we use that? Why force UTF8? Why not rely on one SSOT, throughout the package?

Runtime/Core/Helper/Objects.cs
```diff
@@ -873,6 +876,97 @@ public static uint StableHash32V1(ReadOnlySpan<byte> bytes, uint seed)
...
+        /// <returns>The 64-character lowercase hexadecimal SHA-256 digest.</returns>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="bytes"/> is null.</exception>
+        public static string Sha256Hex(byte[] bytes)
+        {
+            if (bytes == null)
+            {
+                throw new ArgumentNullException(nameof(bytes));
+            }
+
+            using SHA256 sha256 = SHA256.Create();
```
wallstop, line 924: Should we provide a pooled mechanism? I thought creation/destruction was expensive (and allocates), if we utilize a pool (buffers?) that can trigger disposal, should we be good?

On that thought, do we need intelligent pooling/buffered tech for IDisposable objects? If we buffer them, and cache them, do they ever trigger their dispose?

Runtime/Core/Helper/Objects.cs
```diff
@@ -873,6 +876,97 @@ public static uint StableHash32V1(ReadOnlySpan<byte> bytes, uint seed)
...
+        /// When this method returns true, the 64-character lowercase hexadecimal SHA-256 digest of
+        /// the file's contents. When it returns false, null.
+        /// </param>
+        /// <returns>
+        /// True when the file was read and hashed; false when <paramref name="filePath"/> is null
+        /// or empty, the file does not exist, or reading it failed.
+        /// </returns>
+        public static bool TrySha256HexOfFile(string filePath, out string hex)
+        {
+            if (string.IsNullOrEmpty(filePath) || !File.Exists(filePath))
```
wallstop: Should we be using NullOrWhitespace, here and elsewhere?

Runtime/Core/Helper/Objects.cs
```diff
@@ -873,6 +876,97 @@ public static uint StableHash32V1(ReadOnlySpan<byte> bytes, uint seed)
...
+            {
+                hex = null;
+                return false;
+            }
+
+            try
+            {
+                using SHA256 sha256 = SHA256.Create();
+                using FileStream stream = File.OpenRead(filePath);
```
wallstop, line 951: I think we want filestream before sha, since there's a TOCTOU race here against the file exists above. So we might be creating the sha only to throw it away.

Runtime/Core/Helper/Objects.cs
```diff
@@ -873,6 +876,97 @@ public static uint StableHash32V1(ReadOnlySpan<byte> bytes, uint seed)
...
+            catch
+            {
+                hex = null;
+                return false;
+            }
+        }
+
+        private static string ToHex(byte[] digest)
+        {
+            StringBuilder builder = new(digest.Length * 2);
```
wallstop: should we be using a buffered version?

Runtime/Core/Helper/SpriteHelpers.cs
```diff
@@ -51,5 +52,186 @@ public static void MakeReadable(this Texture2D texture)
...
+            if (!texture.isReadable || sourceWidth < 1 || sourceHeight < 1)
+            {
+                texture.LogError($"RotateTexture90 requires a readable texture with pixels.");
+                return null;
+            }
+
+            Color32[] source = texture.GetPixels32();
+            int rotatedWidth = sourceHeight;
+            int rotatedHeight = sourceWidth;
+            Color32[] rotated = new Color32[source.Length];
```
wallstop, line 84: Buffered version?

Runtime/Core/Helper/SpriteHelpers.cs
```diff
@@ -51,5 +52,186 @@ public static void MakeReadable(this Texture2D texture)
...
+            int sourceWidth = texture.width;
+            int sourceHeight = texture.height;
+            if (!texture.isReadable || sourceWidth < 1 || sourceHeight < 1)
+            {
+                texture.LogError($"RotateTexture180 requires a readable texture with pixels.");
+                return null;
+            }
+
+            Color32[] source = texture.GetPixels32();
+            Color32[] rotated = new Color32[source.Length];
```
wallstop, line 124: Buffered version?

Runtime/Core/Helper/SpriteHelpers.cs
```diff
@@ -51,5 +52,186 @@ public static void MakeReadable(this Texture2D texture)
...
+            )
+            {
+                sprite.LogError($"ExtractSpriteRect requires a sprite rect inside its texture.");
+                return null;
+            }
+
+            int width = endX - startX;
+            int height = endY - startY;
+            Color32[] allPixels = texture.GetPixels32();
+            Color32[] region = new Color32[width * height];
```
wallstop, line 195: Buffered version? Here and elsewhere?

.github/workflows/unity-tests.yml
```diff
@@ -362,6 +320,15 @@ jobs:
...
           - 6000.5.2f1
+        # Enrollment contract (Workflow contract item 3): this reviewed STATIC
+        # test-mode axis lets the central editor gate provision the
+        # StandaloneWindowsIl2Cpp profile on standalone legs, so IL2CPP player
+        # builds are verified at gate time. Values must stay the exact reviewed
+        # triple; the central analyzer rejects dynamic or extra modes.
+        test-mode:
+          - editmode
+          - playmode
+          - standalone
```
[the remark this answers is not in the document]
wallstop: The reported shape was already fixed on this branch by `dec47ab6` ("Exclude unselected modes from the Unity test matrix"), which predates the currently reviewed head `738ce69b`. The thread's cited locations (`resolve-test-matrix.js#L44-L47`, `unity-tests.yml#L415-L425`) reference the pre-fix commits `27e6bcb`/`592e232a`, and the flagged `#L44-L47` range in the current source is now the exclude-emitting block itself:

- `resolve-test-matrix.js` `matrix-exclude` carries `{ "test-mode": <mode> }` entries for every unselected mode (same mechanism as unselected versions), and `unity-tests.yml` consumes them via `exclude: ${{ fromJSON(needs.matrix-config.outputs.matrix-exclude) }}` — an unselected leg never starts, so there is no editor provisioning, no organization-lock acquisition, and no way for an unused leg to fail the job after the requested mode passed.
- Contract coverage on the reviewed head: `test-unity-grouped-modes.js` pins the resolver's exclude shape across every version × mode × acceptance combination including the dispatch CLI path (167 controls, green in this run's Contract Suites), and the matrix contract's message states the reviewed rule as "exclude unselected versions and unselected modes through matrix-config".

No further change needed; the per-step mode guards remain as defense in depth.

.github/workflows/unity-tests.yml
```diff
@@ -589,7 +557,9 @@ jobs:
 
       - name: Run additional native acceptance
         id: run_acceptance
-        if: ${{ !cancelled() && github.event_name == 'workflow_dispatch' && inputs.acceptance != '' && inputs.acceptance != 'none' && steps.unity_lock.outputs.acquired == 'true' }}
+        # Acceptance runs once per version, in the editmode leg, independent of
+        # the mode selection: it is requested explicitly per dispatch.
+        if: ${{ !cancelled() && matrix.test-mode == 'editmode' && github.event_name == 'workflow_dispatch' && inputs.acceptance != '' && inputs.acceptance != 'none' && steps.unity_lock.outputs.acquired == 'true' }}
```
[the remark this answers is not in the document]
wallstop: Also already fixed before the currently reviewed head `738ce69b`: `2e218c04` ("fix(ci): run native acceptance on the IL2CPP-provisioned standalone leg") moved the native `intmap`/`serialization` acceptance steps onto the `standalone` leg — the only leg whose editor gate provisions the `StandaloneWindowsIl2Cpp` profile — and `resolve-test-matrix.js` now selects `standalone` whenever acceptance is requested, so the leg always exists for any dispatch pinning (`editmode`/`playmode` + acceptance both fan out to include `standalone`). The thread's cited locations (`unity-tests.yml#L559-L562`, `#L349-L350`) reference the pre-fix commit `592e232a`.

Contract coverage on the reviewed head: `test-unity-acceptance-workflow.js` evaluates every acceptance predicate against the standalone leg (2391 controls, green in this run's Contract Suites), and `test-unity-grouped-modes.js` pins the resolver's acceptance-to-standalone selection across every mode/acceptance combination including the dispatch CLI path.

No further change needed.

Editor/Utils/EditorUtilities.cs
wallstop: **Fix + sweep.** Yes — made public. `EditorUtilities` is already consumer-facing (`public static class`, `GetCurrentPathOfProjectWindow()` public), and the triad it composes with is two-thirds public: `Helpers.IsRunningInBatchMode` and `Helpers.IsRunningInContinuousIntegration` are public documented environment predicates, so the lone internal member was the odd one out. A consumer writing their own editor windows hits the identical test-runner-launch problem and would have written copy #8 of the helper this PR just deduplicated.

The question implied more than a visibility flip, and that part is now fixed too: every one of the seven first-party callers wrapped the call in `try { } catch { }` because `Environment.GetCommandLineArgs()` can refuse on a restricted host. The guard now lives inside the helper — the whole body is covered, so the public contract is "never throws" — and the five window static constructors dropped their now-dead try/catch. `EditorUi`'s static constructor keeps its guard, because its third leg (`IsRunningInContinuousIntegration`, an environment-variable scan) can still throw. The XML summary now states the precise semantics: this detects Test Runner **command-line arguments**; a run started from the in-editor Test Runner window is not detected.

Evidence: new EditMode test `IsInvokedByTestRunnerAgreesWithCommandLineTokens` (does not throw, agrees with the documented tokens), `EditorUiSuppressionTests` 10/10, typecheck battery green, docs section in the editor tools guide, CHANGELOG Added entry.

Runtime/Core/Helper/Objects.cs
wallstop: **Fix.** Yes — `StringExtensions.GetBytes(this string)` was exactly the safe version to use. `Objects.Sha256Hex(string)` now encodes through it after its own null throw; behavior is provably identical (the SSOT's null branch is unreachable behind the existing `ArgumentNullException`, and its empty branch returns `Array.Empty<byte>()`, which hashes to the same empty digest).

Why UTF-8 at all: a digest over a string must pick an encoding, and the in-memory representation is UTF-16 — hashing that would produce a digest no other tool agrees with. UTF-8 is what makes the digest interoperable, and it is now the one spelling the package uses.

Sweep, in three parts: swept where the SSOT contract is identical — `StringExtensions.ToBase64` (same class, already guarded) and the `WProtoUriFormatter.Write` byte path; swept the docs that taught the old spelling (the `StableHash32V1` example in the helper utilities guide and the JSON string-parse example in the serialization guide). Refused where the contract differs: `UnityRandom.EncodeEngineState` maps null → null (the SSOT maps null/empty → empty array), and `WProtoWriter` encodes from a span through a write buffer with no string allocation to begin with. Both divergences now have a recorded reason rather than being accidental.

Runtime/Core/Helper/Objects.cs
wallstop: **Answered with a measurement: no pooling.** Desktop CLR (.NET 9, 256 KiB payload, 2000 hashes per leg, both orderings twice): per-call `SHA256.Create()` 0.0505–0.0523 ms/hash vs a thread-cached, `Initialize()`-reset instance 0.0506–0.0514 ms/hash — the instance cost is ~1 µs against a ~50 µs hash, inside run-to-run noise, and the file path adds multi-ms of I/O on top. The editor was busy with the CI fleet during this pass, so the number is desktop-CLR labeled; per the repo's measurement rule, specializing this call site would need an in-editor number first, and there is nothing to specialize yet.

On the deeper question — pooling IDisposable objects: the pool that fits this package (`SystemArrayPool`) rents buffers, not stateful algorithms. A cached-for-life `HashAlgorithm` would never dispose; that is *safe* for `SHA256` (a managed implementation holding no unmanaged handles, one instance per thread, collected with the thread) but it is a new lifetime policy the package would carry for a win that measures as noise. If a hot path ever proves otherwise, the shape is `[ThreadStatic] SHA256` with `Initialize()` before every `ComputeHash` — mandatory only because `ComputeHash` resets on the success path but not after a mid-stream exception — and the measurement recorded where the specialization lands.

Runtime/Core/Helper/Objects.cs
wallstop: **Fix (here) + measured refusal (elsewhere).** `TrySha256HexOfFile` now uses `string.IsNullOrWhiteSpace`. A whitespace-only path already failed behaviorally (`File.Exists(" ")` is false), but it now refuses before touching the filesystem at all, and there is a `"""/   """` test case in the data-driven set.

Elsewhere — `DurableFile` and `FileHelper` path gates keep `IsNullOrEmpty` deliberately: their whitespace paths already fail closed through directory-creation and open failures, and changing their gate is a behavior surface (staging/cleanup ownership) that this digest fix should not ride. If you want the sweep there, it should be its own reviewed change.

Runtime/Core/Helper/Objects.cs
wallstop: **Fix.** The stream now opens before the hasher exists, and the `File.Exists` pre-check is gone entirely — it was both the TOCTOU race and the extra stat call, and a file vanishing between the check and the open lands in the same `catch` as any other read failure, still returning `false` with a null out. The `<returns>` contract now reads "false when null or whitespace, or when the file does not exist or cannot be read" — same user-visible behavior, no wasted allocation, no check-then-act window.

Runtime/Core/Helper/Objects.cs
wallstop: **Fix.** `ToHex` now writes into a `char[digest.Length * 2]` through a const hex-digit table and returns `new string(chars)` — one char array and one string, no `StringBuilder`, and no per-byte `ToString("x2")` allocation (the old path allocated 32 short strings plus the builder per digest). Same lowercase output; the known-digest vectors pin it.

Runtime/Core/Helper/SpriteHelpers.cs
wallstop: **Investigated on the live editor; the read side cannot be buffered, the write side was prototyped and rejected on measurement, and the sweep found a real bug instead.**

- There is nothing to rent into: `Texture2D` declares `GetPixels32()` and its mip-level overload and nothing else — both return a fresh array. The array-filling overload belongs to `WebCamTexture`, not `Texture2D` (verified on 6000.4.6f1).
- The write side (`new Color32[...]` for the rotated/region buffers) I prototyped as a `SystemArrayPool<Color32>` rent and rejected: `ArrayPool<Color32>.Shared` (TLS-over-per-core-locked-stacks) **stops pooling above 2²⁰ elements** — `Rent(1048576)` returns pooled, `Rent(1050625)` allocates fresh — so a 1024×1024 or larger texture gains nothing, and below that the pool retains up to 2× the requested size (a 513×513 sprite pins a 2 MB array for 1 MB of pixels) for process lifetime. Net worse for the sizes that matter, and the block-write tolerance the pooled shape depends on is only verified on one editor. The plain allocation is scoped, GC-cheap, and honest.

**The sweep paid off elsewhere**: `ColorExtensions`' four `GetRawTextureData<Color32>` averaging fast paths (LAB/HSV/Weighted/Dominant) checked only byte count, so `ARGB32`/`BGRA32` textures — same 4 bytes per pixel, different channel memory order — were being averaged **channel-swipped**. Those fast paths now require `TextureFormat.RGBA32` before reinterpreting raw bytes as `Color32`, with regression cases for every averaging method on both formats (54/54 PlayMode), and a `Fixed` CHANGELOG entry since 3.5.1 shipped it. The rule is recorded in the Unity API costs skill: raw fast paths must check the format, because byte-count equality is not channel-layout equality.

Runtime/Core/Helper/SpriteHelpers.cs
wallstop: Same answer as the thread above (shared shape): the read side has no array-filling overload to rent into (live-verified), the write-side pool was rejected on measured retention/oversize grounds, and the sweep's real find — the `ColorExtensions` `ARGB32`/`BGRA32` channel-swap bug — is fixed with regression cases. See the reply on the `RotateTexture90` thread for the full evidence.

Runtime/Core/Helper/SpriteHelpers.cs
wallstop: Same answer as the two threads above, plus the "elsewhere": the only other `Runtime/` sites are `SpriteMaskReader` (one read per `Try` call — same no-fill-overload limit) and `ColorExtensions`, where the sweep caught and fixed the actual defect (channel-swapped averages on `ARGB32`/`BGRA32` via the raw fast path). The `Editor/Sprites` one-shot tools allocate per invocation in interactive editor sessions only; leaving those alone rather than pooling what a human clicks once.

Runtime/Core/Helper/Objects.cs
wallstop: No, I mean like, does pooling "IDisposable" objects in our Buffers class (in this project) break - do we leak/not dispose things? Should we upgrade/type-check Disposable stuff there on pool clear/clean?

Runtime/Core/Helper/Objects.cs
wallstop: **Correction to my sweep list above.** `WProtoUriFormatter.Write` is back on the open-coded `Encoding.UTF8.GetBytes(value.OriginalString)` spelling — committing the swap red CI's `generator` job, because the proto oracle (`Generator~/WallstopStudios.UnityHelpers.Proto.Generator.Tests`) compiles the serializer sources **standalone** with an explicit file list that does not include `StringExtensions`, so `using ...Core.Extension` does not resolve there (`CS0234`). That compile-surface constraint is now the recorded reason the URI formatter keeps its local encoding; the sweep stands as: `Sha256Hex(string)` → `GetBytes()`, `StringExtensions.ToBase64` → `GetBytes()`, both docs samples teaching the SSOT spelling, `UnityRandom` refused on null contract, `WProtoWriter` refused on the span pipeline, and `WProtoUriFormatter` refused on the oracle's standalone compile surface.

Runtime/Core/Helper/Objects.cs
wallstop: **Audited, fixed where real, and answered with the rule — no type-check on clear.** A full read of `Buffers`/`DisposalLease`/`PoolBuckets`/`TrackedObjectPool` plus every pool construction site:

**Nothing in this PR leaks.** First, the premise: `Objects.Sha256Hex` pools nothing — both sites are `using SHA256 sha256 = SHA256.Create()`, per-call with deterministic release. And `Buffers` itself pools no IDisposable type at all (List, HashSet, Queue, Stack, SortedSet, LinkedList, Dictionary, StringBuilder, Stopwatch, arrays — none implement it). The disposal plumbing is complete: all fifteen permanent-drop paths in both `WallstopGenericPool<T>` variants (pool `Dispose`, `Purge`, `ForceFullPurge`, `PurgeForBudget`, memory-pressure `PurgeCritical`, idle-timeout purge, return-into-a-disposed-pool) route through `InvokeOnDispose`, which fires the constructor's `onDisposal` callback. It runs **outside the pool's lock** on every retirement path, and the six existing `WallstopGenericPoolTests` onDisposal cases (including throw, re-entrant-dispose and return-racing-dispose retire-exactly-once) pin it — 125/125 still green.

**But the audit found one real caller defect, now fixed:** `Serializer`'s `PooledArrayBufferWriter` pool passed only `onRelease: w => w.Reset()` — no `onDisposal` — so every purge (budget enforcement, memory-pressure `PurgeCritical`, explicit `PurgeAllPools`, pool `Dispose`) dropped a writer that still owned an `ArrayPool<byte>.Shared` rental. Honest magnitude: the rent becomes GC-collectable garbage and the shared pool re-allocates on the next miss — no unbounded growth, no "permanent shrink" — but it silently lost reusable capacity on the hot serialization path. One line fixes it (`onDisposal: w => w.Dispose()`), matching its siblings `PooledBufferStream` and the UTF-8 JSON writer pool.

**Why not the upgrade you asked about — a type-check on clear:** ownership is a property of the drop **path**, not the type, and a blanket `is IDisposable → Dispose()` would introduce bugs at exactly this layer. The decisive counter-example is in this package: the comparer-keyed caches store `WallstopGenericPool` values, which *are* `IDisposable`, and callers hold those pools directly — auto-disposing on LRU eviction or `Cache.Clear` would dispose a pool a caller is actively renting from. Same reason `Destroy*Pool` disposes but comparer-cache eviction deliberately does not (documented in the pooling guide: losing a cached pool costs one construction). The correct failure mode is with the pool author, and that is now documented rather than automated: the `onDisposal` parameter doc states every path that fires it and its execution context (it may run while the global budget registry holds its lock, and again from `onRelease`), and the pooling guide gained a [Pooling IDisposable Objects](../blob/main/docs/features/utilities/pooling-guide.md#pooling-idisposable-objects) section with the owning example. Also worth naming the footgun behind the question: `PooledResource<T>.Dispose()` means *return to pool*, not *dispose T* — that is now stated on the type.

**Thread-safety (your implicit second question):** favorable — `InvokeOnDispose` runs outside the pool's own lock on every multithreaded retirement path (snapshot-then-dispose), and the lease machinery elects exactly one disposer per `Get` via a generational CAS. Under the `SINGLE_THREADED` define the CAS is intentionally compiled out, which is correct for that build.

Recorded for follow-up (not ridden by this fix): `GlobalPoolRegistry.EnforceBudget()` currently invokes user purge callbacks while holding its own registry lock — pre-existing, worth its own change.

Editor/Utils/EditorUtilities.cs
wallstop: **Correction to the "five window static constructors drop their now-dead try/catch" claim above — the guards were not dead, and CI proved it.** Removing them redded two Unity legs with `TypeInitializationException`: `Application.isBatchMode` throws `UnityException` when called from a ScriptableObject static initializer under the editor Test Runner ("get_isBatchMode is not allowed to be called from a ScriptableObject constructor (or instance field initializer)"). The guard was protecting the **first** operand of the composition, not the test-runner probe — my try/catch analysis checked the wrong operand's failure modes. All five static constructors carry the guard again, now with a comment naming the actual reason, and the four fixtures that poisoned on the type initializers (`EditorWindowSerializedStateTests`, `EditorWindowNativeResourceTests`, `EditorProgressCleanupTests`, `AnimationClipContentEqualityTests` — 72 cases) pass again in the editor. The visibility promotion, whole-body guard on `IsInvokedByTestRunner` itself, and the semantics-precise docs all stand unchanged.

Runtime/Core/Helper/Objects.cs
wallstop: The follow-ups recorded at the bottom of my answer above are now tracked issues so they cannot evaporate with the PR:

- #752 — `GlobalPoolRegistry.EnforceBudget()` invokes author purge/disposal callbacks while holding `RegistryLock` (snapshot-then-dispose needed, matching the pool's own MT purge shape).
- #753 — `PooledArrayBufferWriter` use-after-dispose spins forever in `EnsureCapacity` (`newSize * 2` on 0); latent, fail-closed hardening.
- #754 — the four stale "SetPixels32 requires exact-length arrays" comments in `Editor/Sprites`, which overstate the API (only the whole-texture overload rejects oversized), plus the floor-editor (2021.3) measurement for the block overload.
- #755 — the comparer-keyed caches' eviction-ownership decision (values are themselves IDisposable pools; needs the `OnEviction` + `TransferOwnershipOnRemoval` + `GetOrAdd`-race single-owner decision before any resource-owning value type enters those caches).

All four carry file:line evidence and acceptance criteria; none is ridden by this PR's changes.

scripts/pr-feedback.sh
```diff
@@ -128,8 +130,37 @@ for comment in inline:
...
+            ", ".join(sorted({comment["user"]["login"] for comment in human_threads})) or "none",
+            len(bot_threads),
+        )
+    )
+for comment in human_threads + bot_threads:
+    print_thread(comment)
+
 heading("2. REVIEW SUBMISSIONS  (approve / request-changes / comment bodies)")
-reviews = [r for r in get("/pulls/%s/reviews?per_page=100" % NUMBER) if (r.get("body") or "").strip()]
+reviews = get("/pulls/%s/reviews?per_page=100" % NUMBER)
```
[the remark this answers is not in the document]
wallstop: **Fix.** The helper now filters empty bodies at the point of consumption: `said_reviews` is filtered to reviews with a non-blank body *before* the `[:1]` grab, with a comment explaining why the filter must live here (section 2 prints empties deliberately, so the helper can no longer assume a bot's first review carries text). The #661 scenario — Copilot's quota/refusal reason printed beside the red check — survives even when the bot's latest review body is empty, because the helper walks to the first review that *has* a body. Also added the follow-up-instruction update to `.llm/context.md` in the same push, requiring session remainders to become tracked issues (#752–#755 went in under exactly that rule).

Runtime/Core/Helper/WallMath.cs
```diff
@@ -808,6 +1106,168 @@ internal static bool WithinTolerance(Vector3 lhs, Vector3 rhs, float tolerance)
...
+
+            return sum;
+        }
+
+        private static double Sum(IReadOnlyList<int> values, int count)
+        {
+            double sum = 0.0;
+            for (int i = 0; i < count; ++i)
+            {
+                sum += values[i];
```
wallstop: Should any of these use intelligence as to the type? Like be super smart about arrays, lists, spans, etc? Unity doesn't de-virtualize IReadonlyList access (here and throughout)

wallstop: **Fix, driven by measurement.** You're right about the mechanism — Unity's Mono JIT does not de-virtualize `IReadOnlyList<T>` access, so every `Sum`/`Variance` pass paid one interface dispatch per element (and standard deviation pays two passes). Each overload now dispatches on the actual receiver:

- **`T[]`** → direct indexing, no copy. Measured on 6000.4.6f1 (100k floats, medians): mean 0.372 → **0.218 ms** (1.7×).
- **`List<T>`** → one pooled `SystemArrayPool` rent, staged with `List<T>.CopyTo` (a single block copy — no per-element interface reads), then both the mean and the squared-difference passes run on the array. Mean 0.372 → **0.301 ms**; standard deviation 0.645 → **0.585 ms**.
- **Any other `IReadOnlyList`** keeps the existing interface loop — for genuinely opaque receivers there is no cheaper read.

**Median and percentile deliberately stay on the pooled-`List` copy**: the obvious `T[] + Array.Sort` alternative measured *slower* (16.83 vs 18.54 ms at 100k) — `List.Sort` already delegates to `Array.Sort` on its backing array, and the pooled-array variant added rent/return overhead on top. No change without a win.

Spans: adding `Span`-taking overloads would widen the API surface; arrays already get the zero-copy path through the same signatures, and span overloads for the sort family remain the deferred scope decision on #742 (candidate 11), which this slot now shares.

Correctness: `WallMathStatisticsTests` 40/40 in the editor — including the non-mutation contract (`StatisticsDoNotMutateTheCallerList`) and every seeded reference cross-check — since the fast paths only change how elements are read, never the arithmetic.

PR #761, opened 2026-09-10
Keep pool cleanup safe under pressure

### LINKED ISSUE #752: Pool purge callbacks run while GlobalPoolRegistry holds its lock
### What happens

`GlobalPoolRegistry.EnforceBudget()` takes `lock (RegistryLock)` at `Runtime/Utils/PoolPurgeSettings.cs:1879` and, while still holding it, calls `pool.PurgeForBudget(toPurge)` at :1927 for each over-budget pool. `PurgeForBudget` is a user-visible purge path: it runs the pool's `onPurge` callback and, for pools constructed with `onDisposal`, the owner's disposal callback for every retired instance (`Runtime/Utils/Buffers.cs` `InvokeOnDispose`, snapshot-then-dispose on the multithreaded variant).

So author-supplied `onPurge`/`onDisposal` callbacks execute **inside the registry's lock** whenever global budget enforcement fires — reachable by default (`_budgetEnforcementEnabled = 1`, `GlobalMaxPooledItems = 50000`), on `EnforceBudget()`, `TryEnforceBudgetIfNeeded()`, and memory-pressure paths.

### Why it matters

- A callback that re-enters the registry (reads `GlobalPoolRegistry.GetStatistics()`, configures a pool, rents/returns) deadlocks or corrupts the loop's assumptions.
- It serializes every pool's purge behind one lock, so one slow `onDisposal` (a file handle close, a native release) stalls budget enforcement for all pools — on whatever thread happened to trip enforcement.
- The same class of re-entrancy hazard is already documented inside `WallstopGenericPool` itself: "Renting a pooled list here can recursively trigger this purge and overflow the stack" (`Buffers.cs:2529-2530`, :2603) — but the registry-level lock is undocumented.

### What to do

Snapshot-then-dispose, mirroring the pool's own multithreaded purge shape: under `RegistryLock`, collect the pools and the per-pool purge counts; release the lock; then run `PurgeForBudget` outside it. Alternatively document the callback-context contract on `onPurge`/`onDisposal` and enforce it — but the snapshot shape matches what the pool already does internally and needs no contract.

### Acceptance

- `EnforceBudget()` invokes no author callback while holding `RegistryLock` (assertable with a callback that calls `Monitor.TryEnter` on the registry lock object or re-enters the registry).
- Existing budget tests (`PoolPurgeSettings`/registry suites) stay green; `unity-tests-single-threaded` leg unaffected.
- The `onDisposal`/`onPurge` parameter docs updated if the contract changes.

### Provenance

Found in the PR #749 Buffers/IDisposable pooling audit: https://github.com/Ambiguous-Interactive/unity-helpers/pull/749#discussion_r3974112769 — recorded as follow-up there; deliberately not ridden by the pooled-writer fix.

### LINKED ISSUE #753: PooledArrayBufferWriter use-after-dispose spins forever in EnsureCapacity
### What happens

`Serializer.PooledArrayBufferWriter.Dispose()` (PR #749, `Runtime/Core/Serialization/Serializer.cs`) sets `_buffer = Array.Empty<byte>()` (length 0) and `_written = 0`. If any code path reaches `EnsureCapacity` afterwards, the growth loop spins forever:

```csharp
int newSize = _buffer.Length;              // 0 after dispose
while (newSize < required)                 // required >= 1
{
    newSize = newSize < 1024 ? newSize * 2 : newSize + (newSize >> 1);   // 0 * 2 == 0
}
```

`newSize * 2` on 0 is 0, so `GetSpan`/`GetMemory`/`Preallocate` on a disposed writer **hangs** instead of failing. The path is unreachable today — a writer reaches `Dispose` only through the pool's `onDisposal` (purge/pool-dispose), after which no live reference should exist — but nothing enforces that, and a consumer-held `PooledResource` copy of a writer dropped by a purge would hang on next use rather than fail.

### What to do

Fail closed, per the repo's never-hang rule: either throw `ObjectDisposedException` from the public members when `_disposed` is set, or make the growth loop total (e.g. `newSize = newSize < 1024 ? System.Math.Max(newSize * 2, 64) : ...`) so it always terminates. Throwing is preferable — it names the contract violation instead of silently renting a fresh buffer for a dead writer.

### Acceptance

- A disposed writer's `GetSpan`/`GetMemory`/`Preallocate`/`Advance` fails fast (or provably terminates) rather than hanging; a regression test pins it.
- The live serialization paths keep passing (serializer JSON/protobuf suites; the writer's pool-disposal fix and its `onDisposal` wiring stay intact).

### Provenance

Noticed while wiring `onDisposal: w => w.Dispose()` on the writer pool in PR #749 (https://github.com/Ambiguous-Interactive/unity-helpers/pull/749#discussion_r3974112769) — recorded as follow-up there, deliberately not ridden by that fix.

### LINKED ISSUE #755: Comparer-keyed pool caches drop evicted pools without disposal; decide the ownership contract
### What happens

`Buffers.CreateCache` (`Runtime/Utils/Buffers.cs:285-297`) builds every comparer-keyed pool cache (`GetHashSetPool`, `GetSortedSetPool`, `GetDictionaryPool`, `GetSortedDictionaryPool`) with **no** `CacheOptions.OnEviction` and default `TransferOwnershipOnRemoval`, so when a cached `WallstopGenericPool` value leaves the cache it is dropped without disposal:

- LRU capacity eviction past `ComparerPoolMaxDistinctEntries` (default 256) — `Cache.EvictOne`, `Cache.cs:1199-1207`
- Resize-driven eviction when the bound is lowered — `CacheResizeRegistry.ResizeAndPublish` → `Cache.Resize`, `Cache.cs:692-725`
- Lost `GetOrAdd` race — the losing racer's pool self-registered with `GlobalPoolRegistry` (`Buffers.cs:1429`/`:2383`) and is then orphaned (`Cache.cs:355-362` `QueueEviction(..., Replaced)`)

This is **benign today and documented as intended**: the pooled value types (HashSet/SortedSet/Dictionary/SortedDictionary) are not IDisposable, the pooling guide states losing a cached pool costs one construction, `GlobalPoolRegistry` holds `WeakReference<IPurgeable>` and prunes dead slots, and `Destroy*Pool` is the explicit dispose path.

### Why track it

The shape is leak-by-construction: the cache's *values are themselves IDisposable* (`WallstopGenericPool<T> : IDisposable`). The moment anyone pools a value type that owns unmanaged or native resources through these caches, eviction leaks it outright — and the PR #749 review established the package's rule that disposal ownership belongs to the drop path, not to a blanket `is IDisposable` type check on clear. A blanket check is wrong here precisely because callers hold the cached pools directly (eight call sites across `Editor/Sprites` and `Editor/Tools`); auto-disposing on eviction would kill a live pool.

So the correct wiring, if/when it is wanted, is opt-in and ownership-aware, not automatic.

### What to do (design decision, then wiring)

1. Decide whether comparer-cache eviction should wire `CacheOptions.OnEviction` **plus** `.TransferOwnershipOnRemoval()` (the combination matters: `Cache.TryRemove` already fires eviction with transfer disabled by default, and the four `Destroy*Pool` methods `TryRemove` then explicitly `Dispose` — naive wiring would double-dispose there, see `Buffers.cs:946/:967/:1133/:1159`).
2. If wired, the callback must dispose only on true eviction (capacity/LRU/resize), not on `TryRemove`-then-explicit-`Dispose`, and must not dispose a pool handed out in a lost `GetOrAdd` race that the winner also registered — the race needs a single-owner decision.
3. Alternatively record the refusal permanently in the pooling guide if the decision is "eviction intentionally abandons; Dispose remains explicit" — the guide already says this in prose; the decision just needs to be a recorded one rather than an omission.

### Acceptance

- Either the eviction path disposes pool values exactly once (with the TryRemove/race interactions pinned by tests in the existing `CacheEvictionOwnershipTests` suite, 15 tests today), or the intentional-abandonment decision is recorded in the pooling guide's comparer-keyed section.
- No behavior change for the current non-IDisposable value types.

### Provenance

From the PR #749 Buffers/IDisposable pooling audit and its adversarial verification (https://github.com/Ambiguous-Interactive/unity-helpers/pull/749#discussion_r3974112769): three auditors and three verifiers converged on "no blanket type check; the hook is `onDisposal`/`OnEviction` at the owning layer" — this is the owning layer that does not yet have the hook.

### LINKED ISSUE #758: LLM Text Attribution / Skill Update
It is critical that, when posting comments, issues, and descriptions, if they are LLM-generated, to disclose them immediately, at the top, as such. All content flows through my github credentials. I do not want to trick people into thinking they are conversing with me, the human.

If there are open issues, commits, PRs or other forms of input from outside contributors, we need to upgrade our skills to _not_ auto-respond or deal with this as me, but, instead, prompt the user (me) for input.

We should never be auto-replying to other users. This cannot happen again. This happened in #747 

It is disrespectful to communicate with human beings via LLM generated text.

wallstop:

**Why:**

Pool cleanup could deadlock callbacks, and disposed writers could spin forever. GitHub agents also lacked enforced attribution and outside-contributor gates.

**What:**

- Run budget callbacks outside registry locks with single-flight enforcement.
- Fail fast when disposed pooled writers are reused.
- Document comparer-cache disposal ownership.
- Enforce LLM disclosure and outside-contributor approval.

Fixes #752
Fixes #753
Fixes #755
Fixes #758

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [ ] Refactor (code change that neither fixes a bug nor adds a feature)

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Budget enforcement changes threading and callback ordering in shared pool infrastructure; agent policy is docs/scripts only but alters how automation interacts with GitHub.
> 
> **Overview**
> Fixes **global pool budget enforcement** so purge and `onDisposal`/`OnPurge` callbacks no longer run under `GlobalPoolRegistry`'s lock, avoiding deadlocks when callbacks touch the registry. Enforcement is **single-flight** (reentrant and concurrent overlapping calls return 0), with new tests and doc updates that callbacks run after the lock is released.
> 
> **Serialization pooling:** `PooledArrayBufferWriter` now **throws `ObjectDisposedException`** on use after dispose instead of misbehaving silently; changelog and contract tests cover this.
> 
> **Agent/GitHub governance:** LLM instructions require a first-line **`DISCLOSURE: LLM-GENERATED TEXT`**, login-based **outside-human pause** before remote action, and PR/issue body validation in shipping guidance. **`scripts/pr-feedback.sh`** adds an authorship audit (viewer login, paginated commits/co-authors, trusted bots only), fail-closed HTTP/JSON handling, `--self-test`, and replaces misleading `--unresolved-only` with **`--root-comments-only`**. **`lint-llm-instructions.ps1`** gains `-AuthorshipPolicyOnly` and CI runs that check; the full lint step order in the workflow is adjusted accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e97c0aae73a813c26e77f5d3492198806e2239dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Runtime/Utils/PoolPurgeSettings.cs
+++ b/Runtime/Utils/PoolPurgeSettings.cs
@@ -1782,7 +1782,7 @@ namespace WallstopStudios.UnityHelpers.Utils
                 {
                     // One pool failure must not prevent cleanup of the others.
 #if UNITY_EDITOR || DEVELOPMENT_BUILD
-                    Debug.LogWarning($"[PoolPurgeSettings] Failed to purge pool: {e.Message}");
+                    Debug.LogWarning($"[PoolPurgeSettings] Failed to purge pool: {e}");
 #endif
                     _ = e;
                 }
@@ -1839,9 +1839,7 @@ namespace WallstopStudios.UnityHelpers.Utils
                 {
                     // One pool failure must not prevent cleanup of the others.
 #if UNITY_EDITOR || DEVELOPMENT_BUILD
-                    Debug.LogWarning(
-                        $"[PoolPurgeSettings] Failed to force-purge pool: {e.Message}"
-                    );
+                    Debug.LogWarning($"[PoolPurgeSettings] Failed to force-purge pool: {e}");
 #endif
                     _ = e;
                 }
@@ -2078,9 +2076,7 @@ namespace WallstopStudios.UnityHelpers.Utils
                 {
                     // One pool failure must not prevent cleanup of the others.
 #if UNITY_EDITOR || DEVELOPMENT_BUILD
-                    Debug.LogWarning(
-                        $"[PoolPurgeSettings] Failed to purge pool for budget: {e.Message}"
-                    );
+                    Debug.LogWarning($"[PoolPurgeSettings] Failed to purge pool for budget: {e}");
 #endif
                     _ = e;
                 }
--- a/Runtime/Utils/PoolPurgeSettings.cs
+++ b/Runtime/Utils/PoolPurgeSettings.cs
@@ -1687,6 +1687,7 @@ namespace WallstopStudios.UnityHelpers.Utils
 
         private static long _globalMaxPooledItems = DefaultGlobalMaxPooledItems;
         private static int _budgetEnforcementEnabled = 1;
+        private static int _budgetEnforcementInProgress;
         private static float _budgetEnforcementIntervalSeconds =
             DefaultBudgetEnforcementIntervalSeconds;
         private static float _lastBudgetEnforcementTime;
@@ -1864,6 +1865,9 @@ namespace WallstopStudios.UnityHelpers.Utils
         /// and items are purged starting from the oldest pools. Each pool's
         /// <see cref="PoolOptions{T}.MinRetainCount"/> is respected.
         /// </para>
+        /// <para>
+        /// Pool purge and disposal callbacks run after the registry releases its lock.
+        /// </para>
         /// </remarks>
         public static int EnforceBudget()
         {
@@ -1873,77 +1877,20 @@ namespace WallstopStudios.UnityHelpers.Utils
                 return 0;
             }
 
-            long currentTotal = 0;
-            int totalPurged = 0;
-
-            lock (RegistryLock)
+            if (Interlocked.CompareExchange(ref _budgetEnforcementInProgress, 1, 0) != 0)
             {
-                BudgetEnforcementPools.Clear();
-
-                for (int i = RegisteredPools.Count - 1; 0 <= i; i--)
-                {
-                    if (!RegisteredPools[i].TryGetTarget(out IPurgeable pool))
-                    {
-                        RegisteredPools.RemoveAt(i);
-                        continue;
-                    }
-
-                    if (pool is IPoolStatistics stats)
-                    {
-                        BudgetEnforcementPools.Add(stats);
-                        currentTotal += stats.CurrentPooledCount;
-                    }
-                }
-
-                if (currentTotal <= maxItems)
-                {
-                    BudgetEnforcementPools.Clear();
-                    return 0;
-                }
-
-                long excess = currentTotal - maxItems;
-
-                SortPoolsByLastAccessTime(BudgetEnforcementPools);
-
-                long remaining = excess;
-
-                for (int i = 0; i < BudgetEnforcementPools.Count && 0 < remaining; i++)
-                {
-                    IPoolStatistics pool = BudgetEnforcementPools[i];
-                    int poolCount = pool.CurrentPooledCount;
-                    if (poolCount <= 0)
-                    {
-                        continue;
-                    }
-
-                    int toPurge = (int)System.Math.Min(remaining, poolCount);
-                    if (toPurge <= 0)
-                    {
-                        continue;
-                    }
-
-                    try
-                    {
-                        int purged = pool.PurgeForBudget(toPurge);
-                        totalPurged += purged;
-                        remaining -= purged;
-                    }
-                    catch (Exception e)
-                    {
-                        // One pool failure must not prevent cleanup of the others.
-#if UNITY_EDITOR || DEVELOPMENT_BUILD
-                        Debug.LogWarning(
-                            $"[PoolPurgeSettings] Failed to purge pool for budget: {e.Message}"
-                        );
-#endif
-                        _ = e;
-                    }
-                }
-
-                BudgetEnforcementPools.Clear();
+                return 0;
             }
 
-            return totalPurged;
+            try
+            {
+                return EnforceBudgetSingleFlight(maxItems);
+            }
+            finally
+            {
+                BudgetEnforcementPools.Clear();
+                Volatile.Write(ref _budgetEnforcementInProgress, 0);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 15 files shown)


Runtime/Utils/PoolPurgeSettings.cs
```diff
@@ -2068,6 +2015,80 @@ internal static void Clear()
...
+                    int purged = pool.PurgeForBudget(toPurge);
+                    totalPurged += purged;
+                    remaining -= purged;
+                }
+                catch (Exception e)
+                {
+                    // One pool failure must not prevent cleanup of the others.
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+                    Debug.LogWarning(
+                        $"[PoolPurgeSettings] Failed to purge pool for budget: {e.Message}"
```
wallstop: Whenever we log exceptions, can we like, log the full exception, instead of just the message? We lose valuable information, like stack traces. Here and throughout. If changes are too invasive, file a tracking issue.

wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed: the budget warning now logs the full exception. Swept: the other two exception logs in `PoolPurgeSettings` now do the same. Rule: repository-wide sites span unrelated systems, so #762 tracks that broader diagnostic sweep. Default and single-threaded typechecks pass, and Unity recompiles without errors.

PR #776, opened 2026-09-12
Create Unity packages without a license

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

**Why:**

Release publication currently spends a licensed Windows runner and organization lock on an archive format that can be produced deterministically without Unity.

**What:**

- Create deterministic `.unitypackage` archives and checksums on `ubuntu-latest`.
- Keep native Unity package compile/export smoke coverage in the test workflow.
- Incorporate #772's statically auditable Unity-version matrix while preserving one-version dispatch filtering.
- Bound release-preparation PR bodies and remove vestigial Claude hook settings.

Fixes #774
Refs #775
Supersedes #772

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [x] Refactor (code change that neither fixes a bug nor adds a feature)

## Checklist

- [x] I have added tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [ ] I have updated the CHANGELOG (not needed: release/CI tooling only)
- [x] My changes do not introduce breaking changes, or breaking changes are documented

---

> [!NOTE]
> **Medium Risk**
> Release artifacts are produced by a new deterministic packer instead of Unity’s native export, though smoke coverage remains on licensed runners; mis-packaging would affect every published release.
> 
> **Overview**
> **Release Publish** no longer exports `.unitypackage` files on licensed self-hosted Windows runners. It builds deterministic gzip/tar archives on `ubuntu-latest` via `scripts/unity/create-unitypackage.js` (staging through existing `stage-unitypackage.js`, stable GUIDs, SHA256 sidecars) and drops the release `runner-preflight`, Unity lock, license return, and aggregate export gate jobs.
> 
> **Unity Tests** still smoke native compile/export on the Windows fleet, but the matrix now lists Unity versions as static text with a `selected-version` axis and full mismatch exclusions so one-version `workflow_dispatch` filtering stays auditable. The `unity-tests` job also requires successful matrix-config and runner-preflight and skips when cancelled.
> 
> **Release Prepare** points reviewers at `CHANGELOG.md` instead of pasting release notes into the PR body, and fails before branch push if the body exceeds GitHub’s 65,536-character limit.
> 
[...]

PR #780, opened 2026-09-12
Improve gameplay helper correctness

### LINKED ISSUE #754: Stale SetPixels32 exact-length comments in the sprite tools; block overload tolerates oversized rents
### What happens

Four `Editor/Sprites` sites carry why-comments claiming `Texture2D.SetPixels32` requires exact-length arrays:

- `Editor/Sprites/SpriteCropper.cs:936` — "SetPixels32 requires exactly width * height elements, so an oversized pool buffer cannot be passed directly."
- `Editor/Sprites/SpriteSheetExtractor.cs:7078` — "SetPixels32 requires exact-length arrays; WallstopFastArrayPool supplies them without oversized buckets."
- `Editor/Sprites/SpriteSheetExtractor.cs:7571` and :7708 — "SetPixels32 requires exact-length arrays, so oversized SystemArrayPool buffers cannot be used directly."

The claim is true only of the **whole-texture** overload, which throws `ArgumentException` on an oversized array. Measured on 6000.4.6f1, the block overload `SetPixels32(0, 0, width, height, oversizedArray, mip)` accepts an oversized rent and writes exactly the requested block — byte-identical to the exact-array call at 5x3, 16x16 and 1x1, mip chain on and off, across RGBA32/ARGB32/RGB24.

### Why it matters

The comments are the package's recorded reason to avoid pooled rents at these sites, and at least one (`SpriteSheetExtractor.cs:7078`) is used to justify a bespoke exact-size pool (`WallstopFastArrayPool`), whose per-size bucket retention the pooling skill flags as the pool shape to avoid. The stated constraint does not hold, so the sites may be able to use the ordinary `SystemArrayPool` rent and delete the exact-size machinery.

### What to do

1. Reword the four comments to the true statement: the whole-texture `SetPixels32` requires exact-length arrays; the block overload takes a count and tolerates oversized rents.
2. Measure `SpriteCropper`/`SpriteSheetExtractor` against the block overload on the floor editor (**2021.3.45f1** — the oversized tolerance is verified on 6000.4.6f1 only, and the 2021.3 floor is the typecheck pin), then switch the sites if clean.
3. If the switch lands, delete the exact-size `WallstopFastArrayPool` usage it existed for and record the measurement next to the block-overload row in `.llm/skills/unity-api-costs.md`.

### Acceptance

- No comment in `Runtime/` or `Editor/` states the exact-length rule without naming the whole-texture overload.
- The block-overload measurement carries a floor-editor (2021.3) data point, not only 6000.4.6f1.
- Sprite tooling output is byte-identical before/after on real sprites (existing sprite editor tests plus a before/after comparison).

### Provenance

From the PR #749 buffered-pixels review thread (https://github.com/Ambiguous-Interactive/unity-helpers/pull/749#discussion_r3972837853): the write side was prototyped as a pooled rent and rejected there on retention/cap grounds for `SpriteHelpers` itself, but the audit also surfaced that these four editor sites' justifying comments overstate the API. Deliberately left alone in #749 to keep the change scoped to runtime code.

### LINKED ISSUE #777: [Feature]: ColorContrast.Composite, so a translucent colour can be measured
## The gap

`ColorContrast.ContrastRatio(first, second)` measures two opaque colours. A translucent colour has no contrast ratio until it is composited over what is behind it, and 3.6.0 has no helper for that step.

IshoBoy keeps a local `WcagContrast` only because of this. After adopting 3.6.0 it could delete `Of` and `RelativeLuminance` in favour of `ColorContrast`, but not `Composite`, which its occlusion outline and its contrast tests call before every measurement:

```csharp
public static Color Composite(Color tone, Color background)
{
    return new Color(
        (tone.r * tone.a) + (background.r * (1f - tone.a)),
        (tone.g * tone.a) + (background.g * (1f - tone.a)),
        (tone.b * tone.a) + (background.b * (1f - tone.a)),
        1f
    );
}
```

## Why it belongs beside `ContrastRatio`

WCAG's contrast ratio is defined on the colours a viewer sees. For a translucent foreground that is the composite, so a caller that passes the raw colour gets a number for a colour nobody sees. Putting the step beside the ratio makes the correct order the easy one.

## Ask

- `ColorContrast.Composite(Color tone, Color background)`: alpha-over onto an opaque background, returning an opaque colour.
- A note on `ContrastRatio` that a translucent argument must be composited first.

Found while adopting 3.6.0 in Ambiguous-Interactive/IshoBoy#905.

### LINKED ISSUE #778: [Feature]: a report context on StartFunctionAsCoroutine for a job hosted elsewhere
## The gap

3.6.0 fixed #359: `StartFunctionAsCoroutine` now catches a throwing tick and reports it once. The report is filed against the `MonoBehaviour` that hosts the coroutine.

That host is not always the component whose job it is. A job that must outlive its own component is started on a long-lived host, such as `CoroutineHandler.Instance`. Its one console line is then filed under the host, and names neither the component nor the work.

IshoBoy keeps a local `PeriodicJobExtensions.StartPeriodicJob` for that reason only. It takes an optional `UnityEngine.Object context` for the report. One caller needs it: the save loop, which hosts its job on `CoroutineHandler` so that it outlives its component, and whose failure matters most.

## Ask

An optional `UnityEngine.Object context = null` on `StartFunctionAsCoroutine`, used as the log context of the once-only report, and the host when it is null.

With it, IshoBoy can delete `PeriodicJobExtensions` and its gate, as that type's own documentation already asks.

Found while adopting 3.6.0 in Ambiguous-Interactive/IshoBoy#905.

wallstop:

## Why

Translucent colors, shared coroutines, and varied sprites lacked reliable helper behavior.

## What

- Add opaque alpha compositing for reliable contrast choices.
- Attribute recurring coroutine errors to their owning Unity object.
- Reuse varied sprite pixel buffers without exact-size retention.
- Gate row-copy parallelism on measured size/row thresholds and use capture-free job types.

## Verification

- Final head `6910a2e93f640fed9e4aa867fab0fe5d3b93aca1`: 25 successful checks and 5 expected skips; every Unity lane passed from 2021.3 through Unity 6, including single-threaded and package smoke.
- All 14 Unity result artifacts preserve every baseline test name and add one passing case per artifact: 155,512 aggregate cases versus 155,498, with identical failed/inconclusive/skipped counts. The 13,824-value contrast sweep remains intact. The repository does not publish line/branch coverage, so this claim is deliberately limited to the executed matrix, test names, and asserted input domain.
- Final six licensed Unity jobs totaled 3,599 seconds versus 3,811 seconds on adjacent baseline #776 (-212 seconds, -5.6%). Artifact-reported test time totaled 2,144.503 seconds versus 2,357.192 (-9.0%). This proves no increase in that matched observation; runner variance means it is not a claim of a stable causal speedup.
- Unity MCP on the final checkout passed the pixel-buffer fixture (1/1), `ColorContrastTests` (11/11 PlayMode), and the broader helper filters (69/69 EditMode, 168/168 PlayMode).
- Exact row-kernel medians support sequential copies below 1,048,576 pixels or 512 rows and capture-free parallel jobs only above both thresholds.

Fixes #777
Fixes #778
Fixes #754


Editor/Sprites/SpriteCropper.cs
```diff
@@ -932,57 +932,65 @@ Dictionary<string, bool> originalReadable
...
-                cropHeight,
-                y =>
-                {
-                    int destRow = y * cropWidth;
-                    int srcY = visibleMinY + y;
-                    if (srcY < 0 || height <= srcY || srcY < srcY0 || srcY1 < srcY)
+                Parallel.For(
+                    0,
+                    cropHeight,
+                    y =>
```
wallstop: Can we ensure that this, and other Parallel.For operations, are benchmarked for usefulness, and, ideally, perform 0 allocations/closures, here and throughout?

Editor/Sprites/SpriteSheetExtractor.cs
```diff
@@ -7568,30 +7608,33 @@ int index
...
+                    {
+                        using PooledArray<Color32> destinationLease = SystemArrayPool<Color32>.Get(
+                            pixelCount,
+                            out Color32[] destPixels
+                        );
+
+                        Parallel.For(
+                            0,
+                            height,
+                            destY =>
```
wallstop: Similar - can we avoid closures here?

Editor/Sprites/SpriteCropper.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Measured on Unity 6000.4.6f1. Parallel crop copies were 23.958x slower at 16x16, 5.770x at 64x64, and 1.670x at 256x256. A noisy 1024x1024 case saved only about 0.135 ms.

Fix: replaced this closure and scheduler with a sequential, precomputed row loop.

Sweep: audited all ten production `Parallel.For` sites and removed the same copy pattern from both extractor paths.

Rule: documented that row copies stay sequential unless multi-size measurements prove a meaningful crossover. #782 tracks the seven heavier remaining sites.

Editor/Sprites/SpriteSheetExtractor.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Measured on Unity 6000.4.6f1. Parallel extraction was 45.445x slower at 16x16, 10.966x at 64x64, 4.284x at 256x256, and 6.375x for 4096x16. Its 1024x1024 gain was only about 0.02 ms.

Fix: both extraction paths now call one sequential `Array.Copy` row helper with no closure.

Sweep: audited all ten production `Parallel.For` sites and fixed the cropper copy loop too.

Rule: documented the measured threshold requirement. #782 tracks the seven heavier remaining sites.

wallstop: Should we do, like, size tests? that is, if > some threshold, do parallel? Here and elsewhere? What about like 4kx4k?

Editor/Sprites/SpriteCropper.cs
wallstop: Similar - should we have a branch?

wallstop: DISCLOSURE: LLM-GENERATED TEXT

Yes — the follow-up now adds a conservative size branch in 6910a2e9: parallel only at >=1,048,576 pixels and >=512 rows. The row floor keeps wide, short images such as 4096x16 on the sequential path.

I re-ran the exact final row-level Array.Copy kernels in .NET 9 Release on this 24-logical-CPU host (seven trials). Median parallel/sequential ratios for extractor and padded crop were 2.80/2.12 at 256² (parallel loses), 0.317/0.322 at 1024², 0.363/0.208 at 2048², and 0.342/0.343 at 4096². Thus 4K is a clear parallel win here. The rule records that this evidence is platform-specific and should be remeasured before generalization.

The small path remains a plain loop with no closure. The large path uses an explicit sealed job; its job/delegate/scheduler can allocate, but there is no lexical capture. The existing Fast editor test now executes both large kernels, compares every pixel (including padded clears), and pins the exact pixel/row boundaries without adding a test case. The repository rule was updated, and #782 tracks the same measurement/threshold sweep for all remaining Parallel.For sites. Unity MCP stopped responding even to status calls, so the supported editor matrix remains CI-verified rather than claimed locally.

Editor/Sprites/SpriteSheetExtractor.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Yes — the follow-up now adds a conservative size branch in 6910a2e9: parallel only at >=1,048,576 pixels and >=512 rows. The row floor keeps wide, short images such as 4096x16 on the sequential path.

I re-ran the exact final row-level Array.Copy kernels in .NET 9 Release on this 24-logical-CPU host (seven trials). Median parallel/sequential ratios for extractor and padded crop were 2.80/2.12 at 256² (parallel loses), 0.317/0.322 at 1024², 0.363/0.208 at 2048², and 0.342/0.343 at 4096². Thus 4K is a clear parallel win here. The rule records that this evidence is platform-specific and should be remeasured before generalization.

The small path remains a plain loop with no closure. The large path uses an explicit sealed job; its job/delegate/scheduler can allocate, but there is no lexical capture. The existing Fast editor test now executes both large kernels, compares every pixel (including padded clears), and pins the exact pixel/row boundaries without adding a test case. The repository rule was updated, and #782 tracks the same measurement/threshold sweep for all remaining Parallel.For sites. Unity MCP stopped responding even to status calls, so the supported editor matrix remains CI-verified rather than claimed locally.

Issue #782, opened 2026-09-12
Benchmark and bound remaining Parallel.For sites


wallstop:
DISCLOSURE: LLM-GENERATED TEXT

## Problem

A sweep from PR #780 found seven remaining `Parallel.For` call sites outside the three row-copy loops fixed there. Several run unconditionally, capture state, or lack a measured crossover threshold.

## Evidence

Unity 6000.4.6f1 measurements on 24 logical CPUs compared parallel row work with sequential `Array.Copy`. Parallel execution was 45.445x slower at 16x16, 10.966x at 64x64, 4.284x at 256x256, and 6.375x for 4096x16. It was about 10.5% faster at 1024x1024, an absolute best-case saving near 0.02 ms. A padded crop was 23.958x, 5.770x, and 1.670x slower through 256x256, then about 2.75x faster at 1024x1024 with an absolute saving near 0.135 ms.

## Scope

- Benchmark `SpriteCropper.ComputeCrop` against a sequential nested scan.
- Benchmark both `ImageBlurTool` passes across image size and blur radius.
- Benchmark both `SpritePivotAdjuster` center-of-mass paths.
- Profile `LayeredImage` around its existing 2,048-pixel threshold.
- Confirm whether `ValidationWorkspaceReport` benefits at realistic finding counts.

## Acceptance criteria

- Use calibrated, counterbalanced timings at multiple adjacent sizes.
- Keep parallel work only above a stable, meaningful crossover.
- Add a sequential path below every retained threshold.
- Remove lexical captures from retained hot-path delegates.
- Use allocation controls that account for worker-thread visibility limits.
- Preserve output parity and existing test coverage.

Follow-up to #780 and the inline performance review.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #788 removes the `ValidationWorkspaceReport` parallel site.

On Unity 6000.4.6f1 with 24 logical CPUs, counterbalanced median batches found 24 workers slower than one-worker `Parallel.For` at every measured size:

| Findings | 1 worker | 24 workers | 24/1 |
| ---: | ---: | ---: | ---: |
| 16 | 197 µs | 237 µs | 1.20x |
| 64 | 757 µs | 1,065 µs | 1.41x |
| 256 | 3.06 ms | 4.51 ms | 1.47x |
| 1,024 | 15.4 ms | 21.6 ms | 1.40x |
| 4,096 | 49.8 ms | 84.7 ms | 1.70x |
[...]

Issue #781, opened 2026-09-12
chore: Scan/Correct/Minimize Closures and Allocations


wallstop:
We should do a sweep of production code (test code exempt) to attempt to eliminate:

- LINQ
- Algorithmically expensive operations (O(n^2) or worse)
- Closures
- Parallel.For usage v normal for (and, if using Parallel.For, to see if we can optimize)
- IReadonlyList/indexed operations v foreach
- foreach loops on non-value-typed enumerators

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #788 advances this sweep in `ValidationWorkspaceReport`:

- Removes a captured `Parallel.For` delegate.
- Removes two Boolean arrays sized to the finding count.
- Preserves report order with a new regression.
- Removes the obsolete report-worker setting and UI.
- Passes all 28 PR checks across every supported Unity version.

The broader production allocation sweep remains open.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #789 advances this issue with a measured reflection hot-path fix.

RCA: `SerializableType.Resolve` eagerly initialized the complete descriptor catalog even when `ReflectionHelpers.TryResolveType` could resolve the supplied assembly-qualified name directly. The fix checks the published descriptor dictionary, attempts direct resolution, and only constructs the catalog for fallback lookup/filtering. A regression test proves resolving a known type does not initialize the descriptor catalog.

The cache publication protocol was also made concurrency-safe: invalidation clears the volatile dictionary sentinel first under the cache lock, and initialization publishes that sentinel only after descriptors, name/display/tooltip arrays, and the empty filter-cache entry are ready. The full PR matrix is green. The broader profiling backlog in this issue remains open.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #790 advances the closure/allocation sweep in `ImageBlurTool`.

RCA: two production `Parallel.For` calls captured surrounding state on every blur, including tiny inputs. The implementation now uses named bound methods and a small-work direct loop, removing lexical captures and avoiding parallel scheduling below the measured 256-pixel threshold or for one-partition passes. Exact forced-path parity tests and the full Unity matrix pass. This addresses the Image Blur slice; the broader production sweep remains open.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #793 advances the closure and allocation sweep in Sprite Pivot Adjuster.

[...]

Issue #775, opened 2026-09-12
chore: refine/make release publish less manual, simpler, faster (more zero-touch)


wallstop:
The release publish workflow can be improved in a few ways:

1. The `unitypackage` format does not actually require unity, or a hosted seat, to run. From my understanding, it's really just a zip/tarball/npm package. We should research + verify this, and then just replicate it, removing the need for a licensed seat
2. We should support "auto" publish, that is, still workflow_dispatch, but by default, hydrates values or pulls values from current main, branch, or (drop down) tag. We can still support manual versions and such, but, if we have a staged release from "prepare release", manually entering in info is lossy and annoying
3. Any other simplifications, from both a UI/UX perspective, and reliability. For example, if we run "Prepare Release", is it possible to, after main merge is green and valid, auto-run "Publish"? Or similar?

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

#776 addresses the immediate release-preparation failure mode: the generated PR body is now deliberately concise, links to the canonical changelog, and is rejected before branch push if it exceeds GitHub's 65,536-character limit. The broader release-PR UX work remains tracked here and in PLAN.md.

Issue #764, opened 2026-09-10
chore: Optimize Source Generators


wallstop:
We should analyze our source generator code and ensure that we're being as efficient in terms of both CPU and I/O. Things like, utilizing StringBuilders (pooled? cached?), using the appropriate builder methods (AppendLine instead of manually inserting "\n" characters), ordering checks, reducing allocations, etc

Additionally, we should ensure that we're shipping the highly optimized, release generator dll instead of the debug version.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #790 fixes the shipped-analyzer overwrite footgun identified in this performance umbrella.

RCA: both analyzer projects copied build outputs into the committed package payload for Debug and Release, so an ordinary Debug build could silently replace optimized shipped DLLs. Copying now defaults on only for Release, an explicit non-Release copy fails before writing, and a 25-check contract exercises the real copy target for both projects. Shipped analyzer bytes are unchanged. The broader performance umbrella remains open.

PR #784, opened 2026-09-13
Reduce runtime and CI performance debt

### LINKED ISSUE #778: [Feature]: a report context on StartFunctionAsCoroutine for a job hosted elsewhere

### LINKED ISSUE #783: tests: reduce runtime, if possible
Right now, our detect asset changed processor and similar tests take a super long amount of time. We should do a sweep of these and other long running tests and see if we can aim for a 50% reduction in test runtime, without sacrificing any coverage (and, ideally, gaining coverage). We need to research available techniques, like moving logic to nunit testable code, less I/O, etc.

wallstop:

**Why:**

Asset-change tests and several production paths were paying avoidable I/O/allocation costs, while coroutine jobs lacked a programmable way to observe every failure and the release graph contained a redundant relay job.

**What:**

- Reuse detached asset-watcher fixture snapshots, preserving isolation and null-state behavior while reducing the measured fixture runtime by 77.4%.
- Add a coroutine exception-handler overload with invocation index, resolved Unity context, continuation semantics, and once-only fallback logging.
- Remove source-generator split/LINQ hot-path allocations, add exact output and diagnostic-order regressions, and rebuild the shipped Release analyzer.
- Remove the release input relay job and bind validated publish stages directly to `verify-tag`, with both focused and broad workflow contracts.
- Evaluate the remaining `Parallel.For` sites and defer unmeasured changes until low-core, single-threaded, IL2CPP, and crossover evidence exists.

Fixes #778
Fixes #783
Refs #764
Refs #775
Refs #781
Refs #782

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [x] Refactor (code change that neither fixes a bug nor adds a feature)

## Checklist

## Validation

- Unity 6000.4: asset processor 35/35 in 50.46s; prefab/scene 21/22 in 12.43s with the same explicit skip.
- Unity 6000.4 PlayMode coroutine regression: 1/1.
- WallstopProto oracle suites: 832/832 (protobuf v2) and 833/833 (protobuf v3).
- Unity typecheck matrix: runtime/editor/tests/integrations, default/legacy/Odin/single-threaded, plus 327 documentation samples.
- Contract suite: 88/88 with bounded local concurrency; agent preflight and shipped-analyzer verification pass.

---

> [!NOTE]
> **Medium Risk**
> Release publishing now depends on a reshaped workflow graph, and editor asset-watcher plus repeating-coroutine failure paths changed behavior that tests cover but production could surface subtly.
> 
> **Overview**
> Cuts CI and editor hot-path overhead while extending repeating coroutine jobs and simplifying the release graph.
> 
> **Release workflow** drops the `release-ready` relay job; `validate-package`, `unitypackage`, `prepare-tag`, and `publish` now read `verify-tag` outputs directly. Contract tests in `scripts/tests` were updated to enforce that wiring.
> 
> **`StartFunctionAsCoroutine`** gains overloads with an optional `Action<ulong, Object, Exception>` handler (invocation index, context, exception). The loop keeps running when the action or handler throws; handler failures log once via `ReportRepeatingJobHandlerFailure`. Docs and PlayMode tests cover the new path.
> 
> **`DetectAssetChangeProcessor`** avoids LINQ/`Split`-style allocations in watcher discovery and uses deep-cloned fixture snapshots (`CloneWatchers`, `ClonePendingChange`, etc.) so tests reset faster without sharing mutable watcher state. Editor integration tests were reworked around `_fixtureSettings`.
> 
> **WallstopProto generator** replaces `Writer.Line` string splitting with index-based appends, scans attributed methods without LINQ, adds `WriterTests` and a diagnostic-order regression for partial types with hook attributes, and exposes internals to the test assembly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1b57ae330d0f4d3062c65a1c6423d678c9517b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/AssetProcessors/DetectAssetChangeProcessor.cs
```diff
@@ -1065,10 +1073,10 @@ private static void BuildWatchers()
                 MethodInfo[] methods = type.GetMethods(flags);
                 foreach (MethodInfo method in methods)
                 {
-                    DetectAssetChangedAttribute[] attributes = method
-                        .GetCustomAttributes(typeof(DetectAssetChangedAttribute), true)
-                        .OfType<DetectAssetChangedAttribute>()
-                        .ToArray();
+                    object[] attributes = method.GetCustomAttributes(
```
wallstop: Don't we have ReflectionHelper optimized versions of this?

Editor/AssetProcessors/DetectAssetChangeProcessor.cs
```diff
@@ -1150,6 +1169,82 @@ out AssetWatcher watcher
...
+                CopyPaths(source.Deleted),
+                CopyPaths(source.Moved),
+                CopyPaths(source.MovedFrom)
+            );
+        }
+
+        private static string[] CopyPaths(IReadOnlyList<string> source)
+        {
+            string[] copy = new string[source.Count];
+            for (int index = 0; index < source.Count; ++index)
```
wallstop, line 1241: There's no specialized bulk operations?

Editor/AssetProcessors/DetectAssetChangeProcessor.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in d1b57ae3: watcher discovery now uses `GetAllAttributesSafe<DetectAssetChangedAttribute>()`, preserving multiple-attribute ordering while using the repository’s safe typed reflection path. Sweep: I also removed the redundant object cast and type filter. Rule: I did not add a lint rule because choosing this reflection helper is semantic and context-dependent; the existing shared helper already expresses the repository policy.

Editor/AssetProcessors/DetectAssetChangeProcessor.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in d1b57ae3: array-backed paths now use `Array.Copy`, other `ICollection<string>` implementations use `CopyTo`, and arbitrary `IReadOnlyList<string>` implementations retain the indexed fallback. Sweep: adjacent `KnownAssetPaths` cloning now uses `HashSet.UnionWith`. Rule: I did not add a lint rule because indexed copying remains the correct fallback for interface-only collections, so a broad static rule would flag valid code.

Issue #636, opened 2026-08-31
#633: establish trustworthy correctness and performance evidence gates


wallstop:
Parent: #633

## Context

The repository already has unusually good foundations—four Unity versions, Edit/Play coverage, Windows IL2CPP, WProto differential tests, RNG quality batteries, and `BenchmarkProtocol` counterbalancing—but the evidence behind a repository-wide correctness/performance claim is not yet trustworthy enough:

- `perf-results/baseline.json` has no usable metrics, while the weekly workflow is report-only and immediately rolls the current run forward even after a regression.
- Most performance fixtures still use one sequential `Stopwatch` aggregate; only newer fixtures use the paired `ABBABAAB` protocol.
- Benchmarks run in Editor/Mono, not a shipping-like player, and do not track IL2CPP binary size.
- `Generator~/Directory.Build.props` deliberately disables Microsoft code-quality analyzers; a fresh exploratory `latest-recommended` scan produced hundreds of findings, including real defects and Unity-specific false positives.
- #616 documents that no local check project compiles `Tests/Editor/**` with the shipped analyzers. Analyzer DLL changes can also leave stale MSBuild `obj` results falsely green.
- There is no instrumented changed-branch or source-mutation gate, and some randomized tests use nondeterministic global seeds without replay data.

This issue establishes the experimental controls needed by every other #633 child. It must land before performance-only changes are accepted; P0 red correctness tests may proceed in parallel.

## Outcome and non-goals

Create a reproducible evidence pipeline that can answer: “Did this exact change preserve the contract, reduce runtime cost on a relevant player, avoid new allocations/retention, and avoid IL2CPP code-size growth?”

This is not a request to make all current analyzer findings fatal, require an arbitrary global coverage percentage, or run the full Unity matrix twice on every PR.

### 1. Repair local gate coverage

- Implement #616 with an editor-test check project using the pinned Unity editor references and the shipped analyzer DLLs.
- Extend the asmdef-reference linter/probes to the new project; list unavoidable exclusions one per line with a reason.
- Make every check project treat the analyzer binaries and relevant configuration as build inputs, or clean the affected `obj` directories before analysis, so a rebuilt analyzer always reruns.
- Add negative controls that introduce a known `WPROTO###`, `WUH###`, and compiler error in each source tree and prove the local command catches them.

### 2. Upgrade `BenchmarkProtocol`

[...]

PR #787, opened 2026-09-13
Reduce editor and local validation overhead

### LINKED ISSUE #786: perf(editor): settings load scans every loaded type on each domain reload
## What happens

`UnityHelpersSettings.OnEnable` calls `ApplyRuntimeConfiguration`, which calls `SerializableTypeCatalog.WarmPatternStats(patterns)`. For each ignore pattern, `GetPatternStats` builds a `Regex` and runs `CountTypesMatchingRegex` over every loaded type.

The settings asset loads on every domain reload. In the reload measured below, the first reader is the static constructor of `FailedTestsExporter`, so the scan runs inside `ProcessInitializeOnLoadAttributes`.

## Measured

Unity 6000.5.2f1, unity-helpers 3.6.0, a project with 66,512 loaded types and the 6 default patterns.

| measurement | result |
| --- | --- |
| `EnableDomainReloadTimings`: `FailedTestsExporter..cctor` → `LoadSerializedFileAndForget(ProjectSettings/UnityHelpersSettings.asset)` | 1208 ms, of which `Loading.AwakeFromLoad` is 1201 ms and the file read is 7 ms |
| `ProcessInitializeOnLoadAttributes` in the same reload | 2519 ms |
| `WarmPatternStats` with `PatternStatsCache` cleared, in a warm domain, 3 runs | 382 ms, 348 ms, 341 ms |
| the same 6 counts, with and without `RegexOptions.Compiled` | 337 ms, 348 ms |

A consuming project's batchmode CI (self-hosted Windows) spends 8411 ms in `ProcessInitializeOnLoadAttributes` per editor launch, and pays it on every launch.

## The only reader of the result

`GetPatternStats` has one other caller: the settings window's match count beside each pattern (`UnityHelpersSettings.cs:2073`). That call already computes on demand and caches. So every domain reload pays for a label that is drawn only when the settings window is open.

## Suggested fix

- Remove `WarmPatternStats` from `ApplyRuntimeConfiguration`. The window computes each count the first time it draws the pattern.
- If a pre-warmed count is still wanted, warm it from the settings provider's activate handler instead.
- `RegexOptions.Compiled` does not help a one-shot count. The measured cost was the same without it.

## Not measured

About 850 ms of the 1201 ms `AwakeFromLoad` is not attributed here. It can include JIT of the settings type. A fresh-domain measurement after the fix would say how much of the reload the warm-up was.

Related: #648.

wallstop:

**Why:** Domain reloads and local checks repeated expensive whole-workspace work.

**What:**

- Remove eager settings type scans.
- Bound and parallelize local validation safely.
- Reduce analyzer work and make name matching ordinal.
- Derive release versions from the selected source.

Fixes #786
Refs #636, #764, #775, #781


Generator~/WallstopStudios.UnityHelpers.Proto.Generator/WProtoGenerator.cs
```diff
@@ -3591,12 +3596,22 @@ private static bool IsPartialEverywhere(INamedTypeSymbol symbol)
...
-                        && declaration.Modifiers.Any(modifier => modifier.ValueText == "partial")
-                    )
+                    if (!(reference.GetSyntax() is TypeDeclarationSyntax declaration))
+                    {
+                        continue;
+                    }
+
+                    foreach (Microsoft.CodeAnalysis.SyntaxToken modifier in declaration.Modifiers)
+                    {
+                        if (modifier.ValueText == "partial")
```
wallstop: Should this, or any other use string.Equals, comparisons, Ordinal?

wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed, swept, and ruled. The partial-token check now uses `SyntaxKind.PartialKeyword`. Both analyzer assemblies now use explicit ordinal semantics for compiler names, metadata keys, and name-keyed collections. The agent guidance now requires `SymbolEqualityComparer.Default` when both sides are symbols and explicit ordinal comparers otherwise.

Issue #633, opened 2026-08-30
chore: correctness + perf sweep/audit


wallstop:
We should do a full audit of the code base in mind for:

- Correctness, in all cases
- Performance (can we reduce allocations? Buffers? Pooling?)
- Performance (can we optimize any algorithms? Batching? Without sacrificing correctness?)
- Performance (can we de-virtualize any calls, reduce branches, shrink code size?)
- Performance (can we operate on unsigned values, reduce bounds checks, etc?)
- Safety (no UB, no unsafe code, maximum warnings as errors, all static analyzers, etc)
- Simplicity (how easy is it for clients to use this code and have success?)

Areas of interest:

- Relational Components
- Wallstop Proto / JSON processing
- Buffers/Pooling
- IList/sorting operations
- Random number generators
- Reflection tech
- Singleton (SO/Runtime)
- Tags/Effects system
- Spatial trees
- Logging
- String operations
- Enum operations
- Hashcodes
- Editor Tooling

## Execution campaign

This issue is the umbrella and decision ledger for the audit. Child issues own implementable contracts, red tests, measurements, backend matrices, and definitions of done. Do not mix unrelated fixes into one large optimization PR.

### Scientific workflow for every task

1. **Observe:** record the current source path, public contract, suspected failure/cost, and representative caller/workload.
2. **Red:** add an independent reference model, fixed edge case, deterministic property/metamorphic check, hostile input, seeded mutant, or replayable schedule that fails for the expected reason.
3. **Baseline:** use #636 to capture merge-base and candidate in paired/randomized order on the same machine. Save raw samples and environment metadata.
4. **Green:** make the smallest safe change. Correctness, exception atomicity, ownership, AOT/linker behavior, and public compatibility precede speed.
5. **Challenge:** run differential tests, minimized replay seeds, branch/define variants, and the floor/latest Mono + IL2CPP matrix. WebGL gets compile/smoke coverage where runtime execution is impractical.
6. **Measure:** report time/op or throughput, bytes/op, peak/retained bytes, output/quality metrics, and binary/build-time cost when relevant. Diagnose the mechanism; do not accept CoreCLR/editor-only microbenchmarks as player proof.
7. **Decide:** accept a default-path optimization only when its declared representative workload improves by at least 5% on both Unity 2021.3 floor and latest supported IL2CPP and no other representative workload regresses by more than 5%, unless the child defines a stricter threshold. Material allocation reductions may be accepted with latency within 5%. Publish negative results.
[...]

PR #795, opened 2026-09-14
Add stable weighted gameplay helpers and utility hardening

### LINKED ISSUE #792: [Feature]: weighted pick that stays stable when a weight changes

### Feature Description

Add a weighted index draw that stays stable when a weight changes. It draws one uniform per slot and returns the slot whose clock `-ln(1 - u) / weight` is smallest. Each slot wins with probability `weight / total`, the same as `NextWeightedIndex`.

### Use Case / Motivation

- **Mutation and tuning.** `NextWeightedIndex` adds weights in order. When one weight changes, every later interval moves, so the same uniform picks another slot. In DoxReloaded's level mutation (20 types, 30 % of weights moved by up to ±0.2), the cumulative pick flipped **14.9 %** of draws and the race flipped **4.9 %** (200,000 paired draws). A mutated level's board distance fell from 0.67 to 0.19 of an unrelated level ([DoxReloaded#848](https://github.com/Ambiguous-Interactive/DoxReloaded/pull/848)).
- **A/B with common random numbers.** Two loot or spawn tables that differ in one weight should differ only near that weight.
- **Double weights.** DoxReloaded's NRPA solver writes its own cumulative softmax pick over `double[]` (`Solver.Nrpa.cs`), because `NextWeightedIndex` takes only `float`.

**Copies today.** DoxReloaded: `ButtonDistribution.SampleByRace` (race over floats), `ButtonDistribution.Sample` and `SampleArity` (cumulative over floats), and `Solver.Nrpa` (cumulative over doubles).

**Searched.** `RandomExtensions.NextWeightedIndex(IReadOnlyList<float>)` and `(float[])`, `NextWeighted` and `WeightedShuffleBag` on `main` (`9bab9008`), and the 3.6.0 changelog. All are cumulative or bag-based. No double or span overload exists.

### Proposed Solution

```csharp
public static int NextWeightedIndexByRace(this IRandom random, ReadOnlySpan<float> weights);
public static int NextWeightedIndexByRace(this IRandom random, ReadOnlySpan<double> weights);
public static int NextWeightedIndex(this IRandom random, ReadOnlySpan<double> weights);

// k = destination.Length picks without replacement: the k smallest clocks.
public static void NextWeightedSubsetByRace(
    this IRandom random, ReadOnlySpan<double> weights, Span<int> destination);
```

- Draw one uniform for **every** slot, zero weights included. The draw count then never depends on which weights are positive, and a weight that crosses zero moves no other slot's draw.
- A non-positive weight is an infinite clock. Ties go to the lower index.
- Fail closed on NaN, infinity or no positive weight, as #638 does for bounded sampling.
- `Math.Log` can differ in its last bit between runtimes, so a near tie can flip. Document it.
- The subset variant is Efraimidis–Spirakis. Keep it allocation-free with a partial selection.
- Tests: a seeded chi-square on frequencies; on a seeded panel of weight changes, the race flips fewer draws than the cumulative pick; draws consumed equal the slot count; NaN, infinity, zero and single-slot cases; a golden sequence.

### Alternatives Considered

- Cumulative draw: unstable, as measured above.
- Alias method: fast, but a rebuilt table reorders picks, so it is unstable too.
- Gumbel-max: the same decision as the race, written with a second log.

### Package Version

3.5.0 in DoxReloaded. `main` read at `9bab9008`.

Related: #742, #638. Requested in [review](https://github.com/Ambiguous-Interactive/DoxReloaded/pull/848#discussion_r4006967052).

wallstop:

## Summary

- add overflow-safe cumulative and stable exponential-race weighted span selection, including allocation-free weighted subsets without replacement
- add reusable command-line parsing, RectTransform/BoxCollider2D synchronization, and defensive DPI-scaled drag thresholds
- promote the bounded-random arithmetic proof into the Ubuntu Local Gates job with pinned Z3, explicit ownership tests, and bounded native execution
- document the public contracts and add changelog coverage

## Validation

- Unity MCP recompile: 0 errors
- `RandomExtensionsTests`: 79/79 (0.62s final run)
- `UnityExtensionsComprehensiveTests`: 135/135 (2.05s)
- `HelpersTests`: 75/75 (2.49s)
- `ValidationReportingTests`: 32/32
- runtime, editor, test, and editor-test typecheck variants: 0 warnings / 0 errors
- bounded proof: all 8/16/32/64-bit obligations, production `MulHi64`, 3/3 mutants killed, 4,294,967,040 reduced-width subjects
- fast contract aggregate: 88/88
- agent preflight, content/tooling/format/spelling/EOL/changelog/UPM gates: passed
- adversarial reviews for #792, #742, and #638: zero remaining revisions

Closes #792
Refs #742, #781, #638, #633

---

> [!NOTE]
> **Medium Risk**
> New weighted-selection paths affect procedural gameplay RNG semantics (including platform-dependent `Math.Log` near-ties), while validation CLI parsing is a mechanical move to shared helpers with regression tests.
> 
> **Overview**
> Adds **span-based weighted random APIs** on `IRandom`: overflow-safe cumulative `TryNextWeightedIndex` for `ReadOnlySpan<double>`, exponential-race single picks (`TryNextWeightedIndexByRace`), and **without-replacement subsets** (`TryNextWeightedSubsetByRace`) with overlap checks, optional caller scratch, and pooled scratch above 1,024 winners.
> 
> **UI/helper utilities** gain `RectTransform.TrySyncBoxCollider2D`, DPI-scaled `EventSystem` drag thresholds (`CalculatePixelDragThreshold` / `TryApplyPixelDragThresholdForCurrentDpi`), and shared **`Helpers` command-line parsing**; `ValidationBatch` drops its local parsers and calls those helpers instead.
> 
> **CI/docs** pin and run the bounded-random Z3 proof in Ubuntu Local Gates (Python 3.11, `requirements-random-quality.txt`, 2m timeout, stricter shell flags); the proof script adds compile/run timeouts, and a contract test asserts the workflow wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f28de9fafec2b855d07be47fd86ea5541e54ef14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/local-gates.yml
```diff
@@ -95,15 +95,30 @@ jobs:
         with:
           dotnet-version: "9.0.306"
 
+      - name: Setup Python
```
wallstop, line 98: Should we use `uv`, here and everywhere that python is necessary?

Runtime/Core/Extension/Partials/UI.cs
```diff
@@ -56,6 +68,112 @@ public static void SetBottom(this RectTransform rt, float bottom)
...
+        /// </returns>
+        public static int CalculatePixelDragThreshold(
+            int baseThreshold,
+            float dpi,
+            float referenceDpi = 160f
+        )
+        {
+            int clampedBase = Mathf.Max(0, baseThreshold);
+            if (
+                !float.IsFinite(dpi)
```
wallstop, line 122: Is this NaN safe, here and elsewhere?

.github/workflows/local-gates.yml
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Not as a blanket rule. The docs workflows use pinned `uv==0.12.6` because that larger requirements set was measured to save 10–15 seconds, while this gate installs one cached wheel and adding the `uv` bootstrap would add another install and pin to maintain. The repository also still uses direct `pip` for the single-package wiki-link test. I’m keeping this proof on `setup-python` + its requirements-keyed pip cache unless we measure a net benefit for this specific one-wheel workload; a repository-wide Python runner standardization would be a separate scoped change.

Runtime/Core/Extension/Partials/UI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Yes. Both floating inputs are rejected unless finite and strictly positive, and `clampedBase` is non-negative. Therefore `scaled` cannot be NaN or negative. Its largest possible value for finite `float` inputs is still within `double` range; even if a platform produced positive infinity, `int.MaxValue <= scaled` catches it before `Math.Round` or the cast. The rounded value is likewise checked before casting. The surrounding collider path separately validates both `rect.center` and `rect.size` with `float.IsFinite` before either collider field is mutated.

Runtime/Core/Extension/RandomExtensions.cs
```diff
@@ -1297,6 +1300,301 @@ public static int NextWeightedIndex(this IRandom random, float[] weights)
...
+        )
+        {
+            if (!TryGetWeightScale(random, weights, out double scale))
+            {
+                index = default;
+                return false;
+            }
+
+            double normalizedTotal = 0d;
+            for (int weightIndex = 0; weightIndex < weights.Length; weightIndex++)
```
wallstop, line 1330: Is this better or worse than foreach? Here and other span loops

wallstop: DISCLOSURE: LLM-GENERATED TEXT

For `Span<T>`, both indexed `for` and `foreach` are allocation-free; `foreach` uses the span’s ref-struct enumerator. Here the algorithm needs the index as the selected result, last-positive fallback, stable tie key, and subset destination value, so `foreach` would still need a separately maintained counter and would be less direct without a demonstrated speedup. I’d keep indexed loops where the index is algorithmic state. For span loops that only consume values, `foreach` is equally reasonable; I would choose by clarity and benchmark a hot path before making a repository-wide performance claim.

Runtime/Core/Extension/RandomExtensions.cs
```diff
@@ -1480,6 +1778,212 @@ private static int ClampSkewedInteger(float value, int min, int max)
...
+                {
+                    return;
+                }
+
+                SwapWeightedRaceEntries(scores, indices, parentIndex, entryIndex);
+                entryIndex = parentIndex;
+            }
+        }
+
+        private static void SiftWeightedRaceDown(
```
wallstop, line 1828: Should any of this be partial classes, for maintainability?

Runtime/Core/Extension/RandomExtensions.cs
wallstop: Alright, please file a follow-up issue for tracking utilizing foreach throughout the prod/editor code base

.github/workflows/local-gates.yml
wallstop: Please file a follow up item to do a repo-wide python standardization (and bias towards `uv`)

wallstop: DISCLOSURE: LLM-GENERATED TEXT

Filed #797 to inventory every Python path, measure cold/warm behavior, standardize on pinned `uv` where compatible and beneficial, and document any justified exceptions: https://github.com/Ambiguous-Interactive/unity-helpers/issues/797

Runtime/Core/Extension/RandomExtensions.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Filed #796 for the Runtime/Editor audit, with the distinction captured explicitly: prefer `foreach` for value-only span traversal when it improves clarity, retain indexed loops where the index is algorithmic state, and benchmark before making performance claims: https://github.com/Ambiguous-Interactive/unity-helpers/issues/796

Runtime/Core/Extension/RandomExtensions.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Yes. At roughly 2,276 lines, `RandomExtensions` is large enough that cohesive partial files would improve navigation. The weighted-selection family and its private validation/heap helpers are a natural first boundary. I kept that mechanical refactor out of this already-green functional change so draw order, documentation, and review scope stay easy to verify, and filed #798 with behavior-, allocation-, API-, and docs-neutral acceptance criteria: https://github.com/Ambiguous-Interactive/unity-helpers/issues/798

Issue #794, opened 2026-09-14
[Feature]: Odin Conversions


wallstop:
### Feature Description

It would be great if we had some tooling that clients/users could run to convert bits/swaths of their project from Odin tech to the Unity Helpers equivalent, seamlessly, without data loss.

### Use Case / Motivation

Onboarding client repos that currently depend on Odin, but want to offboard.

### Proposed Solution

Needs research + design, likely editor tools (sibling repo DxMessaging does similar tooling for upgrading from older version)

### Unity Version

None

### Package Version

None

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #802 adds a conservative preview/apply tool for exact, globally qualified Odin inspector attributes under `Assets`. It preserves supported encodings and newline styles, creates byte-for-byte backups, checks for concurrent edits, attempts safe rollback, and reports serialized-state blockers without rewriting them.

This issue should remain open: migrating Odin-owned serialized bases and collections without data loss needs a separately designed and validated data-migration stage.

wallstop:
This should be for all inspector attributes, serializable data, as well as anything Sentinel can handle (AND extra Unity Helpers stuff in general)

wallstop:
> Authored by OpenAI Codex (GPT-5.4) on behalf of @wallstop.

PR #805 advances this feature with a safety-focused follow-up to the initial tool.

RCA: the source analyzer treated properties and nonserialized fields as eligible migration targets, even though removing Odin can change whether Unity draws or persists those members. It also recognized only a fixed subset of globally qualified Odin attributes, and a failed/cancelled filesystem scan could still leave a partial Apply plan.

The fix limits automatic edits to public fields or fields carrying exact `global::UnityEngine.SerializeField[Attribute]` / `SerializeReference[Attribute]`, leaves unsupported member shapes and Odin-owned serialized state manual, reports every `global::Sirenix.OdinInspector.*` attribute, and disables Apply after cancellation or any scan failure. The migration guide now states those boundaries explicitly.

Evidence: Unity 6000.4.6f1 passed all 35 `OdinMigrationToolTests` after recompilation; default and Odin editor-test typechecks pass with zero warnings/errors. The broader seamless-conversion feature remains open.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #806 advances the migration report for exact `ShowIf(nameof(...))` and `HideIf(nameof(...))` conditions.

[...]

PR #812, opened 2026-09-17
Report possible Odin data in authored assets

wallstop:

**Why:** The Odin migration preview did not show possible Odin state in authored assets.

**What:**
- Report possible serialized-data keys and prefab override paths with file and line numbers.
- Scan Unity text scenes, prefabs, and assets under `Assets` and report unreadable files.
- Keep serialized data untouched and preserve safe source-only Apply.
- Cover plain and quoted YAML keys with Editor tests.

Refs #794

---

> [!NOTE]
> **Low Risk**
> Editor-only preview/reporting with no serialized asset writes; apply semantics for scripts are unchanged aside from an additional read pass over project assets.
> 
> **Overview**
> **Extends the Odin migration preview** so it also scans Unity text `.unity`, `.prefab`, and `.asset` files under `Assets` for possible Odin `serializationData` / `_serializationData` payloads and prefab `propertyPath` overrides that target those fields, with file and line in the report as **DATA REVIEW** items.
> 
> Adds `OdinMigrationSerializedDataScanner` and wires it into `OdinMigrationTool.BuildPlans` alongside the existing C# script pass. **Apply still rewrites only source scripts**; serialized-data findings and data-scan failures are informational and do not block safe apply (source scan failures and cancellation still do).
> 
> **Authored asset YAML parsing** now normalizes quoted mapping keys so checks match Unity’s on-disk YAML; docs and changelog describe the broader preview behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bb9c7eba950f3d1c480dcf227d8d1fd924c0337. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/Tools/OdinMigration/OdinMigrationSerializedDataScanner.cs
```diff
@@ -0,0 +1,268 @@
...
+            if (0 <= source.IndexOf('\0'))
+            {
+                return false;
+            }
+            SkipWhitespace(source, ref position, source.Length);
+            return StartsWith(source, position, "%YAML 1.1")
+                || StartsWith(source, position, "--- !u!");
+        }
+
+        private static void AnalyzeLine(
```
wallstop: Should this utilize our YAMLAsset/Parser thing at all? This seems extremely brittle

Editor/Tools/OdinMigration/OdinMigrationTool.cs
```diff
@@ -95,6 +106,60 @@ out string decodeFailure
                 );
                 result.Replacements += analysis.ReplacementCount;
             }
+
+            if (result.Cancelled)
+            {
+                return result;
+            }
+
+            for (int index = 0; index < serializedAssetPaths.Count; index++)
```
wallstop: Can we/should we lift .Count/.Length out of loops, here and throughout the codebase (and enforce into the future)?

Editor/Tools/OdinMigration/OdinMigrationTool.cs
```diff
@@ -95,6 +106,60 @@ out string decodeFailure
...
+                string assetPath = serializedAssetPaths[index];
+                if (context.ShouldCancel(assetPath, assetPaths.Count + index, totalFiles))
+                {
+                    result.Cancelled = true;
+                    break;
+                }
+
+                try
+                {
+                    byte[] bytes = context.ReadAllBytes(context.GetFullPath(assetPath));
```
wallstop: Can we un-nest, for being able to understand failures? Are there async versions that we can/should use?

Editor/Tools/OdinMigration/OdinMigrationSerializedDataScanner.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

I searched this package for `YAMLAsset` and YAML parser code; neither is present here. The scan is read-only and reports possible Odin state, so a false match asks for review but cannot change data. I added quoted mapping-key coverage and a test. Full YAML parsing would need a parser dependency or a shared implementation; I left this thread open for that design choice.

Editor/Tools/OdinMigration/OdinMigrationTool.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed this scan: both list counts are read once, and the cached values drive progress and loop bounds. I checked the wider rule. A blanket `.Count`/`.Length` hoist can change loops that grow or shrink a collection, while array and `List<T>` bounds are cheap. I would not enforce it across the codebase without a narrower rule tied to measured cost.

Editor/Tools/OdinMigration/OdinMigrationTool.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed: the asset loop now calls a small scan method with early returns for decode and YAML failures. The exception path still records the asset name. This editor command uses synchronous file reads with per-file progress and cancellation; an async file read would not reduce Unity main-thread parsing work and would need a separate progress/cancellation design. No broad rule is warranted for this command shape.

Editor/Tools/OdinMigration/OdinMigrationSerializedDataScanner.cs
```diff
@@ -0,0 +1,278 @@
...
+                && position + 1 < end
+                && char.IsWhiteSpace(source[position + 1])
+            )
+            {
+                position++;
+                SkipWhitespace(source, ref position, end);
+            }
+
+            char keyQuote =
+                source[position] == '\'' || source[position] == '"' ? source[position++] : '\0';
```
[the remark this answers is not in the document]
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in `c997648f`: the scanner returns when a list marker and whitespace consume the line. The quoted-key test now ends with `- ` after three findings, so an overrun fails that test. I checked the other indexed reads in this method; their bounds guards remain in place. No wider rule is needed for this local parser boundary.

Editor/Tools/OdinMigration/OdinMigrationSerializedDataScanner.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Correction: my earlier reply was wrong. I missed `Editor/Validation/AuthoredAssetYaml.cs`, which is a reusable Unity document reader in this package. Commit `4bb9c7eb` now uses it for the Odin data scan. I added simple quoted-key support to that shared reader, kept the small `propertyPath` scalar check, and made files with no Unity document report an incomplete scan. I swept the parser and scanner cases; 44 focused tests pass. The shared reader is now the rule for both checks.

Issue #817, opened 2026-09-19
chore: Production Code Base Sweep


wallstop:
We should sweep the code base to ensure:

- Optimal algorithmic use (for owned code, where we are not implementing explicit reference algorithms)
- Higher performance
- Less (minimal) allocations
- Zero LINQ
- Less (minimal) closures, ideally zero
- TOCTOU bugs (File.Exists / File.Delete style)

All backed by red-green verification of improvements across a variety of data sets.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

I found two user-facing editor save paths that checked `File.Exists(destination)` and later chose `File.Replace` or `File.Move`. A destination created or removed between those calls could make the save fail. Commit b6910e49 on PR #825 routes Texture Resizer PNG saves and Analyzer Policies ruleset saves through `DurableFile.TryWriteAllBytes`, which stages the bytes and handles that race. The former hand-written replacement branches are gone; the shared writer is now the only direct `File.Replace`/`File.Move` user under `Editor` and `Runtime`.

Focused Unity 6000.4 editor fixtures passed after the change (TextureResizerWizardTests 6/6; AnalyzerPolicyWindowTests 1/1). A probe confirmed the ruleset serialization produces the same bytes. CI is running for the new commit. This addresses these two save paths only; the broader production sweep and its data-set performance evidence remain open.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

The remaining filesystem sweep found two user-facing write races:

- `Editor/Tools/ImageBlurAPI.cs`: `TryWriteAsset` loops on `File.Exists` to choose a free name, then uses `File.WriteAllBytes`. A file created at that name between the probe and write can be overwritten.
- `Editor/Sprites/SpriteSheetExtractionAPI.cs`: `Extract` honors `overwriteExisting == false` with `File.Exists`, but `TryWrite` later uses `File.WriteAllBytes`. A newly created output can be overwritten despite the option.

A follow-up should reserve no-overwrite outputs with `FileMode.CreateNew` and handle an occupied path without truncating it. For the overwrite path, stage the encoded bytes before replacement. Tests should cover existing output preservation, name collision handling, and failure cleanup. I am leaving this for the next focused slice while PR #828's Unity matrix runs.

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

[...]

Issue #815, opened 2026-09-18
Consuming 3.6.0 outside Unity is not zero-touch


wallstop:
## What happened

DoxReloaded compiles a subset of this package's `Runtime` sources into a plain `netstandard2.1` project so its solver/generator code can be benchmarked and tested without Unity. Moving 3.5.0 → 3.6.0 cost 246 compile errors and four hand-written workarounds.

Filed at the request of the consuming repo's owner, who asked whether upstream can make this zero-touch.

## The three touches

**1. The generated registrar needs `ModuleInitializerAttribute`, which netstandard2.1 does not have.**

`WProtoGeneratedRegistrar.g.cs` emits `[ModuleInitializer]`, so the consumer must hand-write:

```csharp
namespace System.Runtime.CompilerServices
{
    [AttributeUsage(AttributeTargets.Method, Inherited = false)]
    internal sealed class ModuleInitializerAttribute : Attribute { }
}
```

Could the generator emit this polyfill itself when the compilation does not already define it? That is the usual pattern for generators targeting netstandard2.x, and it would make this touch disappear.

**2. `[WProtoContract]` on `AbstractRandom` cascades to every subclass, including test doubles.**

Any type deriving from `AbstractRandom` now must be `partial` (`WPROTO001`), have a parameterless constructor (`WPROTO011`), and be at least `internal` so the generated registrar can name it (`WPROTO040` → `CS0122`). Five throwaway EditMode test fakes — a counting random, a fixed random, a prefixed random and two cancel-after-N randoms — needed all three, purely so a formatter could be generated for types that are never serialized.

Is there an opt-out? `[WProtoIgnore]` or `[WProtoNotSerialized]` on a subclass that should not participate would cover this. If one exists, the diagnostics do not mention it, and the message text is otherwise excellent.

**3. Unity-only types reach non-Unity consumers through `Core/Random` and `Core/Serialization`.**

`UnityRandom` needs `UnityEngine.JsonUtility` and `Random.State`; `ProtobufUnitySurrogates` needs `RectInt`, `BoundsInt` and `Resolution`; `WProtoBootstrap` needs `RuntimeInitializeOnLoadMethod`. `AbstractRandom` references `UnityRandom` directly, so a consumer cannot simply exclude the Unity-bound file. All of these were shimmed by hand.

## Not a request to change the design

The upgrade is worth it and the package works. This is only about the cost of consuming it from a non-Unity target, which item 1 alone would mostly remove.

[...]

PR #820, opened 2026-09-19
Fix settings load and modernize agent tooling

### LINKED ISSUE #607: migrate to agentic skills
Our generic .llm harness is great, except for two things:

- It does not follow the current standard of agentic skills (SKILL.md): https://agenticskills.io/
- Many skills are too verbose and could be split up into research/code-samples/etc
- The skills might not be auto-detected by codex, claude, copilot, opencode

We should keep the generic .llm harness, but:

- Migrate to the current, modern standard of skills and agent context
- Ensure single-source-of-truth
- Ensure full compat with claude code, codex, opencode
- Ensure context files, including skills, are <200 lines

---

## Measured, so the four asks can be priced separately

`.llm/` holds **120 Markdown files, 112 of them skills, 31,949 lines**.

| Ask | Status | Evidence |
| --- | --- | --- |
| Single source of truth | **Already done** | Five entry points -- `CLAUDE.md`, `AGENTS.md`, `.cursorrules`, `.github/copilot-instructions.md` -- are 3 lines each and all say "See [AI Agent Guidelines](./.llm/context.md)". Nothing is duplicated. |
| Auto-detection | **Not done, and it is the real gap** | `.claude/skills/` does not exist. Claude Code discovers skills only there, as `<name>/SKILL.md` with YAML frontmatter. None of the 112 are discoverable; an agent reaches them only by reading `.llm/context.md` and following a link. |
| SKILL.md format | **Not done** | A skill today opens `# Skill: Create Test` then an HTML-comment trigger line. No YAML frontmatter, no per-skill directory. |
| <200 lines | **Not done, at scale** | **75 of 112 skills are over 200 lines**, median **300**, longest 476. `.llm/context.md` itself is 494. 78 of the 120 `.llm` files exceed the target. |

Distribution, so the split work is not a guess:

| Over | Files |
| ---: | ---: |
| 100 | 102 |
| 150 | 88 |
| **200** | **75** |
| 300 | 55 |
| 400 | 27 |

## Why this needs a decision before a migration

The three asks pull in different directions, and picking wrong is expensive to undo:

1. **Auto-detection wants content in `.claude/skills/<name>/SKILL.md`.** Single source of truth then means `.llm/skills/` either *becomes* that tree, or is generated from it, or the two diverge. Generating 32,000 lines into a second committed tree is the option that looks cheapest and is worst -- the repository already treats "two hand-maintained copies of one document" as a defect ([#441](https://github.com/Ambiguous-Interactive/unity-helpers/issues/441), the `README.md` / `docs/readme.md` finding).
2. **Codex and opencode do not read `.claude/skills/`.** Codex reads `AGENTS.md`; `opencode.json` here configures MCP and nothing else. So "full compat with claude code, codex, opencode" cannot be satisfied by the Claude layout alone -- a shared tree plus per-tool thin pointers is the only shape that serves all three, and that is what `.llm/` already is. The question is whether `.llm/skills/` **moves** to `.claude/skills/` with the others pointing at it, or stays and `.claude/skills/` holds pointers.
3. **The <200-line rule turns 75 files into ~150-220 files.** `lint-skill-sizes.ps1` enforces 500 today; tightening it to 200 fails 75 files at once, so the limit has to be lowered in step with the splitting rather than up front.

## Suggested first slice, if the layout is settled

1. Decide the tree: `.claude/skills/<name>/SKILL.md` as the source, with `.llm/skills/` retired and `.llm/context.md` pointing at the new paths -- or the reverse.
2. Add YAML frontmatter (`name`, `description`) generated from the existing trigger comments, which already carry both. `scripts/generate-skills-index.ps1` parses them today, so the mapping exists.
3. Split the 27 files over 400 lines first, into `SKILL.md` plus `reference.md` / `examples.md` in the same directory. That is where the density is worst and the split is most obvious.
4. Lower `lint-skill-sizes.ps1` one band at a time -- 400, then 300, then 200 -- so each tightening is a green commit rather than a 75-file cliff.

**Not started**, because step 1 is yours: every later step is cheap to do and expensive to redo against the wrong tree.

### LINKED ISSUE #797: chore: standardize repository Python tooling around uv

## Context

Python setup currently varies across repository workflows: some documentation jobs use pinned `uv`, while the proof and wiki-link jobs use `pip`. Issue #338 researched `uv`; PR #795 prompted a maintainer request to standardize the repository and bias toward `uv`.

## Scope

Inventory every Python installation and execution path in CI, hooks, and local scripts. Define one documented policy, preferring pinned `uv` unless measurements or compatibility requirements justify an exception.

## Acceptance criteria

- [ ] Inventory Python setup, dependency installation, caching, and invocation across workflows and scripts.
- [ ] Measure representative cold and warm runs before changing runners.
- [ ] Standardize on a pinned `uv` version where it is compatible and beneficial.
- [ ] Document justified exceptions and the update procedure.
- [ ] Keep requirements lock/pin behavior reproducible.
- [ ] Verify local gates, documentation builds, proof checks, and wiki-link checks remain green.

Follow-up to #338. Requested in PR #795 review: https://github.com/Ambiguous-Interactive/unity-helpers/pull/795#discussion_r4010543332

### LINKED ISSUE #814: [Bug]: May not pass in already-persistent Objects
### Description

Found when booting Unity and not entering Safe Mode.

Image

### Steps to Reproduce

Boot Unity without safe mode, unclear which objects were throwing the persistence error.

### Expected Behavior

Not to get this error

### Actual Behavior

Threw error when booting in console, wanted me to boot into Safe Mode

### Severity

Minor (feature broken, workaround exists)

### Unity Version

Unity 6.3 (6000.3)

### Package Version

3.6.0

### Platform

- [x] Editor
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] iOS
- [ ] Android
- [ ] WebGL
- [ ] Other

### Regression

- [x] Yes, this is a regression (worked in a previous version)

### Additional Context

_No response_

wallstop:

**Why:** Settings loading can trigger a Unity persistence error, and the project's Python and agent guidance workflows need reproducible, discoverable maintenance.

**What:**

- Defer settings migration saves until Unity finishes loading, with an editor regression test.
- Report invalid file deletion targets accurately and retain random subtype serialization in Unity while allowing a smaller plain .NET source set.
- Pin documentation and wiki Python dependencies and document measured install choices.
- Generate Codex and Claude skill discovery files from canonical `.llm` guidance, split long guides below 200 lines, and verify regeneration.
- Update guidance size tests and CI paths so new entries and limits stay checked.

Fixes #814
Fixes #797
Fixes #607
Refs #815
Refs #817

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [x] Refactor (code change that neither fixes a bug nor adds a feature)

## Checklist

---

> [!NOTE]
> **Low Risk**
> Mostly generated docs and CI wiring; the editor settings deferral is localized but touches project-settings persistence at startup.
> 
> **Overview**
> Defers Unity Helpers **settings migration saves** until the editor finishes loading (with a regression test), addressing startup persistence errors ([#814](https://github.com/Ambiguous-Interactive/unity-helpers/issues/814)). The PR description also covers related runtime/tooling fixes (invalid delete reporting, random subtype serialization) that are not in the excerpted diff.
> 
> **Agent guidance** is reorganized so **`.llm/skills` stays canonical**: `scripts/generate-agent-skill-stubs.mjs` emits matching discovery stubs under **`.agents/skills`** and **`.claude/skills`** (now tracked via `.gitignore` exceptions). Oversized skills are split into **`.llm/references/*` parts**, and **`.llm/context.md`** is shortened with pointers to new `context-*.md` reference files.
> 
> **Python/CI** pins docs installs to **`requirements-docs.lock`**, bumps **`uv` to 0.12.10** in workflows and the devcontainer (including `uv pip` for tool requirements and a `/usr/local/bin/uv` symlink), and installs wiki test deps from **`requirements-wiki.txt`**. **LLM Instructions Lint** watches the new agent paths and runs **`test-agent-skill-migration.mjs`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17c9f88ed02048e970056edaf009b9fe853fa877. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/Settings/UnityHelpersSettings.cs
```diff
@@ -4135,6 +4128,21 @@ KeyValuePair<string, WButtonCustomColor> entry in _wbuttonCustomColors.ToArray()
...
+                return;
+            }
+
+            EditorApplication.delayCall -= SaveAfterLoad;
+            _saveAfterLoadQueued = false;
+            if (this != null)
+            {
+                SaveSettings();
+            }
+        }
```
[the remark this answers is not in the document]
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in ed29443a: the queued save now runs from `EditorApplication.update` only when `isUpdating` and `isCompiling` are false. A busy callback keeps the save queued; an idle callback saves once. The native editor regression test covers both passes. I swept the other `ScriptableSingleton` settings load path: `ValidationWorkspaceSettings.OnEnable` normalizes without saving, so this was the sole load-time save site. The focused test is the guard for this Unity lifecycle behavior.

Editor/Settings/UnityHelpersSettings.cs
```diff
@@ -3990,36 +4000,19 @@ internal void OnEnable()
...
-                SaveSettings();
-                shouldApplyRuntimeConfig = false;
-            }
-            if (EnsureSerializableSetTweenDefaults())
-            {
-                SaveSettings();
-                shouldApplyRuntimeConfig = false;
-            }
-            if (shouldApplyRuntimeConfig)
+            settingsChanged |= EnsureFoldoutTweenDefaults();
```
wallstop: Can we avoid bitwise logic like this? Can we ensure that these functions are just always called?

wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in ad609a39. Each default helper now has its own `if`, so every call runs and any changed result sets `settingsChanged`. I swept direct method calls in boolean `|=` assignments across Runtime and Editor and converted the remaining settings and dispatcher sites; assignments that combine already computed booleans remain. A new lint rule would need C# type analysis to distinguish this pattern from bitmask accumulation, so the code sweep is the guard here. The focused native editor test passed after the change.

Issue #827, opened 2026-09-20
chore: Reliability


wallstop:
- Scan production code for TOCTOU (concurrent check + do thing, File Exist -> Do, etc), replace with atomic operations
- See if we can apply DurableFile concepts throughout any production file operations
- Sweep all code for try/finally - see if we can replace with value-based IDisposable concepts (that are copy-safe, and steady-state allocation-free)
- Sweep code that SHOULD have try/finally or IDisposable concepts (do a thing, perform maybe-throwing operation, undo a thing)
- Optimize Buffer paths (can we avoid locks? Research + A/B/A test various strategies for increasing perf and reducing allocations + ensuring sweep is extremely fast)

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #828 addresses one data-loss path in this reliability sweep. `FileHelper.CopyFileAsync` opened an existing destination with `FileMode.Create` before copying; cancellation or a read error could erase the previous file. It now uses the staged `DurableFile` copy and retains its buffer-size option. The same review found that both `DurableFile` copy paths could leak an opened source handle if destination directory creation failed; that is fixed with sync and async regressions.

Unity 6000.4.6f1 passed 45 `DurableFileTests` and 25 `FileHelperTests`, and the local typecheck matrix passed. PR CI is still running. The broader issue remains open for other file operations, cleanup paths, and measured buffer strategy work. On platforms without `File.Replace`, the existing delete-then-move fallback can still leave the destination absent if the final move fails; the user guide now states that limit.

wallstop:
https://github.com/Ambiguous-Interactive/unity-helpers/pull/829#discussion_r4058615219

https://github.com/Ambiguous-Interactive/unity-helpers/pull/829#discussion_r4058615786

wallstop:
DISCLOSURE: LLM-GENERATED TEXT

PR #830 is merged at `c78796c3`. RCA: several editor report exports opened the destination with truncation before rendering or completing the write, so a failed export could destroy a prior valid report. Prefab Checker JSON/CSV, Animation Copier preview, Method Analyzer JSON/Markdown, validation batch/window, and failed-test reports now render first and publish through `DurableFile.TryWriteAllText`. The Prefab Checker CSV keeps its existing output format. Staged-image cleanup failures in Sprite Sheet Extraction and Image Blur are now reported without masking the primary write error.

[...]

PR #828, opened 2026-09-20
Preserve files when async copies fail

wallstop:

**Why:** A failed async copy could erase an existing file.

**What:**

- Stage copies before replacing the destination.
- Close source streams when destination directory creation fails.
- Test cancellation and source cleanup in Unity.

Refs #827

---

> [!NOTE]
> **Medium Risk**
> Changes async copy behavior and centralizes it on DurableFile staging/swap logic, which affects any caller relying on FileHelper.CopyFileAsync for overwrites.
> 
> **Overview**
> **`FileHelper.CopyFileAsync`** no longer opens the destination with `FileMode.Create` up front. It delegates to **`DurableFile`**’s staged copy path so the existing destination is only replaced after a full read and flush to a `.tmp` sibling; read/staging failures and cancellation leave the prior file intact.
> 
> **`DurableFile`** adds **`OpenCopySourceStream`** (open source, then create destination directories, disposing the source if directory creation fails) for both sync and async copy. **`CopyAsync`** gains a public overload that forwards to an internal implementation with configurable **`bufferSize`** and validation. **`TryDelete`** is repositioned with clearer docs: success means **`File.Delete`** completed without throwing, not merely “nothing left at the path.”
> 
> Tests cover source handle cleanup when destination directory creation fails, cancellation preserving destination contents, and no leftover temp files. Changelog and helper-utilities docs describe the new copy semantics and platform swap caveats.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8edcfc19dabab2b730f137c5405e0109442b0c12. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Helper/DurableFile.cs
```diff
@@ -340,12 +339,50 @@ public static bool TryCopy(string sourcePath, string destinationPath, out Except
...
         /// <param name="cancellationToken">Optional cancellation token.</param>
         /// <returns>Null on success, otherwise the failure.</returns>
-        public static async ValueTask<Exception> CopyAsync(
+        public static ValueTask<Exception> CopyAsync(
             string sourcePath,
             string destinationPath,
             CancellationToken cancellationToken = default
         )
         {
+            return CopyAsync(sourcePath, destinationPath, DefaultBufferSize, cancellationToken);
```
wallstop, line 348: Should these be async +await (is that better or worse for stack traces?)

Runtime/Core/Helper/DurableFile.cs
```diff
@@ -340,12 +339,50 @@ public static bool TryCopy(string sourcePath, string destinationPath, out Except
...
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                return false;
+            }
+
+            try
+            {
+                File.Delete(path);
+                return true;
```
wallstop, line 367: Even with TOCTOU, what does bool actually mean here? If the file didn't exist, this will return true, right? Is that expected/intended?

Runtime/Core/Helper/DurableFile.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Decision: keep the direct `ValueTask` return. The inner async overload catches I/O exceptions and returns them as the method result, so adding `async`/`await` to this forwarding wrapper would not improve an exception stack trace. It would add another async state machine/frame for a path that does no work after delegation. I checked the wrapper and the inner catch path; no code change is needed for this question.

Runtime/Core/Helper/DurableFile.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fix: `true` means `File.Delete` completed without throwing; that includes an already absent file. This is intentional idempotent cleanup, and `DeleteReportsSuccessWhenNothingIsThere` already tests it. A concurrent writer can recreate the path afterward, so the previous “no file remains” XML return text was too strong. Commit `8edcfc19` clarifies the contract. Sweep: I checked the other `TryDelete` references and delete tests; no caller relies on a post-return absence guarantee. No new rule is needed for this API-specific wording.

Issue #821, opened 2026-09-19
[Feature]: Generate sprite atlases from code, not only the window


wallstop:
### Feature Description

A public, static, prompt-free way to do what the Sprite Atlas Generator window does: scan a `ScriptableSpriteAtlas` config's folders, generate the `.spriteatlas` it describes, and pack. Add a way to ask whether an atlas is still what its config would generate. Then the configs can be the source of truth from scripts, tests and batch mode as well as the window.

Today all of it lives on `ScriptableSpriteAtlasEditor`, an `EditorWindow` (3.6.0):

| member | access |
| --- | --- |
| `GenerateSingleAtlas(ScriptableSpriteAtlas, bool)` | `private` |
| `ScanFoldersForConfig(ScriptableSpriteAtlas)` | `private` |
| `LoadAtlasConfigs()`, `GenerateAllAtlases()`, `PackAllProjectAtlases()` | `internal`, instance, reading the window's `_atlasConfigs` |
| `TexturePlatformNameHelper.DefaultPlatformName` | `internal` |

So a consumer can drive it only by opening the window and pressing buttons, or by reflecting into members the package is free to rename.

### Use Case / Motivation

IshoBoy (Ambiguous-Interactive/IshoBoy#971) raised the padding on three atlases so an affordance outline stops inside the padding in a packed build. The first attempt edited the `.spriteatlas` files directly. The next "Generate + Pack All" would have undone it, because the configs still said 4 and 8. An earlier IshoBoy change had done the same to a fourth atlas. Neither was caught, because nothing compared an atlas with its config.

The fix set the padding in the configs and regenerated with the tool, through reflection:

```csharp
Type editor = typeof(ScriptableSpriteAtlasEditor);
BindingFlags flags = BindingFlags.Instance | BindingFlags.NonPublic;
ScriptableObject window = ScriptableObject.CreateInstance(editor);
editor.GetMethod("LoadAtlasConfigs", flags).Invoke(window, null);
editor.GetMethod("GenerateAllAtlases", flags).Invoke(window, null);
Object.DestroyImmediate(window);
```

[...]

PR #829, opened 2026-09-20
Protect editor asset outputs

wallstop:

**Why:** Editor output could overwrite another file, and animation event saves required opening a window.

**What:**

- Publish blurred and extracted images only to free output paths.
- Stage Sprite Cropper PNG writes before replacing existing files.
- Expose animation event saves to editor scripts with Undo.
- Add collision, failed-write, and Undo tests.

Refs #817, #821, #827

---

> [!NOTE]
> **Medium Risk**
> Touches multiple editor code paths that write project assets; behavior changes (skip vs overwrite, naming) could surprise batch workflows, though changes are defensive and heavily tested.
> 
> **Overview**
> Editor asset writes are hardened so concurrent or failed operations are less likely to clobber existing files, and animation event persistence is exposed for scripts.
> 
> A new **`AnimationEventSaveAPI.TrySave`** centralizes validation, Undo, dirty marking, and asset save; the Animation Event Editor now calls it instead of writing the clip inline. **`ExclusiveFilePublisher`** publishes staged bytes with exclusive `CreateNew` so a destination that already exists is left untouched.
> 
> **Image Blur**, **sprite sheet extraction**, and **sprite cropping** now stage PNG output (or use **`DurableFile`**) before replacing files. Blur keeps occupied names and picks the next free suffix; extraction skips outputs that appear mid-run unless `overwriteExisting` is set. Cropper rejects empty PNG encodes and failed durable writes without changing prior bytes. Texture resizer reports encode failures instead of writing empty data. Changelog, editor-tools docs, EditorCheck shims/exclusions, and edit-mode tests cover publish races, skip/overwrite behavior, and Undo for the new save API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02589356ceeab160c12567e177055260c27d8e31. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Editor/Tools/ImageBlurAPI.cs
```diff
@@ -267,6 +307,19 @@ out string error
...
+            try
+            {
+                File.Move(stagedPath, destinationPath);
+                return true;
+            }
+            catch (IOException) when (File.Exists(destinationPath))
+            {
+                return false;
+            }
+        }
```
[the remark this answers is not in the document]
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in 3af89abd. `File.Move` documents failure for an occupied destination, but its Unix implementation can check the path before a rename, leaving a creation race. Both blur and extraction now use `File.Copy(..., overwrite: false)` from the staged image; the destination is created exclusively, and a concurrent publisher cannot replace it. I added a concurrent two-writer Unity test alongside the existing occupied-path cases; the extraction fixture passes 12/12 and blur passes 25/25. I swept both `TryPublishNewFile` implementations. The docs now state the remaining limitation: an interrupted copy to a new path can leave an incomplete new file.

Editor/Sprites/SpriteCropperAPI.cs
```diff
@@ -552,7 +553,20 @@ out Color32[] croppedPixels
...
+                            out Exception writeError
+                        )
+                    )
+                    {
+                        return new CropResult(
+                            CropStatus.RetryableError,
+                            null,
+                            $"Failed to write cropped texture at '{outputPath}': {writeError.Message}"
+                        );
+                    }
```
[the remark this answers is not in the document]
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in 3af89abd. Cropper now rejects null or empty PNG bytes before calling `DurableFile`; the regression test verifies both cases leave an existing file unchanged, and the cropper fixture passes 3/3. I swept editor PNG writes through `DurableFile`: Sprite Sheet Extraction and Image Blur already reject null encoding; Texture Resizer had the same gap, so it now rejects null or empty bytes and reports an error. Its fixture passes 6/6. The rule is to validate encoder output at each call site because `DurableFile` intentionally treats null bytes as empty for its general API.

Editor/Sprites/SpriteSheetExtractionAPI.cs
```diff
@@ -366,6 +375,20 @@ out bool sourceWasReadable
             return result;
         }
 
+        internal static bool TryPublishNewFile(string stagedPath, string destinationPath)
+        {
+            try
+            {
+                File.Copy(stagedPath, destinationPath, overwrite: false);
+                File.Delete(stagedPath);
```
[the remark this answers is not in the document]
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in 02589356. Both callers now share `ExclusiveFilePublisher`, which catches an occupied destination only around the atomic `FileMode.CreateNew` claim. Copy and flush failures propagate as errors; a completed output stays successful if staged-file deletion fails, avoiding a skip or retry under a second name. I swept both blur and extraction callers and kept their stage cleanup. The extraction regression now asserts that an output-path error is reported rather than skipped, and the concurrent test checks that only the winner's stage is removed. Unity 6000.4.6f1 passed extraction 12/12 and blur 25/25 after the change. The guide states that failed publishing only attempts temporary-file cleanup.

Editor/Sprites/SpriteSheetExtractionAPI.cs
```diff
@@ -523,21 +541,80 @@ out Color32[] pixels
...
+                            {
+                                skipped = true;
+                                return false;
+                            }
+                        }
+                        finally
+                        {
+                            if (stagedOwned)
+                            {
+                                File.Delete(stagedPath);
```
wallstop, line 601: Should this be in its own try/catch?

Editor/Tools/ImageBlurAPI.cs
```diff
@@ -134,9 +126,58 @@ public static bool TryWriteAsset(
...
+                                            "No available blurred output name was found."
+                                        );
+                                    }
+                                }
+                            }
+                            finally
+                            {
+                                if (stagedOwned)
+                                {
+                                    File.Delete(stagedPath);
```
wallstop, line 178: Should this be in its own try catch?

Editor/Sprites/SpriteSheetExtractionAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fix: Yes. In #830 (`6b43a712`), extraction cleanup now has its own try/catch. A cleanup failure is reported without hiding the extraction error or being counted as an occupied output.

Sweep: I fixed the same pattern in Image Blur.

Rule: cleanup in `finally` must preserve the primary failure and report its own failure separately.

Editor/Tools/ImageBlurAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fix: Yes. In #830 (`6b43a712`), Image Blur cleanup now has its own try/catch. Its error is appended to the write error instead of replacing it.

Sweep: I fixed the same pattern in sprite sheet extraction.

Rule: cleanup in `finally` must preserve the primary failure and report its own failure separately.

PR #842, opened 2026-09-22
Let scripts resize textures without a window

wallstop:

**Why:** Texture resizing required a wizard, and subset sampling copied common sources twice.

**What:**

- Add a prompt-free texture resize API with explicit inputs and safety checks.
- Copy queue and set sources once during random subset setup.
- Remove duplicate file checks from two runtime readers.
- Cover behavior with Unity tests and update user guides.

Refs #821, #781, #817, #827

---

> [!NOTE]
> **Medium Risk**
> Editor code can overwrite PNG assets and temporarily flip texture importer readability; validation and staged writes reduce but do not eliminate partial-run or data-loss risk on real runs.
> 
> **Overview**
> Adds **`TextureResizerAPI.TryResizeTextures`** so editor scripts can resize or dry-run PNGs from explicit texture lists and asset folders without opening the Texture Resizer wizard. The wizard now only gathers folder paths and delegates to that API.
> 
> The API **validates settings, folders, and output-name collisions before any file or importer changes**, blocks overwriting textures outside `Assets/`, processes textures in natural name order, and caps growth at 16k. Failures are contained with try/catch at the public entry (progress cleanup tests updated accordingly).
> 
> **`RandomExtensions.NextSubset`** snapshots queues, stacks, hash sets, and linked lists in one copy instead of double materialization, and no longer trusts an untrusted `ICollection.Count` when building the fallback list.
> 
> **`DirectoryHelper.ReadPackageIdFromRoot`** and **`Serializer.TryReadFromJsonFile`** drop redundant `File.Exists` checks and rely on read/parse failures instead. Docs and integration tests cover the API, subset behavior, and wizard edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1d411d2825a8e75f73cd1187b4eeec95bf407f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Runtime/Core/Extension/RandomExtensions.Subsets.cs
```diff
@@ -66,8 +67,41 @@ int count
...
+                        nameof(count)
+                    );
+                }
+
+                T[] snapshot = new T[queue.Count];
+                queue.CopyTo(snapshot, 0);
+                return NextSubsetIterator(random, snapshot, count);
+            }
+
+            if (items is HashSet<T> set)
```
wallstop, line 100: Why aren't we buffering here? Why can't we do a more general collection, or similar? Why are we allocating temporary arrays?

Here and all other specializations.

Editor/Sprites/TextureResizerAPI.cs
```diff
@@ -0,0 +1,636 @@
...
+            int numResizes,
+            ResizeAlgorithm scalingResizeAlgorithm,
+            int pixelsPerUnit,
+            float widthMultiplier,
+            float heightMultiplier,
+            string outputDirAssetPath,
+            bool dryRun
+        )
+        {
+            if (outputDirAssetPath != null && string.IsNullOrWhiteSpace(outputDirAssetPath))
```
wallstop: Is this... correct? Why do we have like, multiple null/empty/whitespace checks? Why does this exist at all if we have a stronger one below?

Editor/Sprites/TextureResizerAPI.cs
```diff
@@ -0,0 +1,636 @@
...
+                        || outputDirAssetPath.StartsWith("Assets/", StringComparison.Ordinal)
+                    )
+                )
+            )
+            {
+                Debug.LogError($"The output folder is invalid: {outputDirAssetPath}.");
+                return false;
+            }
+
+            if (
```
wallstop, line 87: Should these be checked before AssetDB query, for perf? Can we do a perf scan + optimization pass of this tech?

Editor/Sprites/TextureResizerAPI.cs
```diff
@@ -0,0 +1,636 @@
...
+                foreach (Texture2D texture in selectedTextures)
+                {
+                    textures.Add(texture);
+                }
+            }
+
+            using PooledResource<HashSet<string>> sourcePathsResource = Buffers<string>.HashSet.Get(
+                out HashSet<string> sourcePaths
+            );
+            {
```
wallstop, line 143: Why is this scope here?

Editor/Sprites/TextureResizerAPI.cs
```diff
@@ -0,0 +1,636 @@
...
+                    continue;
+                }
+
+                if (distinct.Add(t))
+                {
+                    ordered.Add(t);
+                }
+            }
+
+            ordered.Sort(static (a, b) => string.Compare(a.name, b.name, StringComparison.Ordinal));
```
wallstop: Isn't there, like, a name comparer that we have implemented and use that is awesome and robust? Why can't we use that here, and everywhere we sort/compare names?

Editor/Sprites/TextureResizerAPI.cs
```diff
@@ -0,0 +1,636 @@
...
+                            return false;
+                        }
+
+                        _ = sourcePaths.Add(path);
+                    }
+                }
+
+                if (0 < sourcePaths.Count)
+                {
+                    string[] sourceFolders = new string[sourcePaths.Count];
```
wallstop: Should any of this be buffered? Here and anywhere else we make arrays? Using one of our array pools? Is that better or worse?

Editor/Sprites/TextureResizerAPI.cs
```diff
@@ -0,0 +1,636 @@
...
+                                    );
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            finally
+            {
+                Utils.EditorUi.ClearProgress();
```
wallstop, line 505: Can we make this into a re-usable, zero-alloc, using + IDisposable concept for use internally and externally?

Editor/Sprites/TextureResizerAPI.cs
```diff
@@ -0,0 +1,636 @@
...
+            }
+
+            // Outside Assets, ensure the physical output directory exists even when registration is unavailable.
+            string dirAsset = Path.GetDirectoryName(assetPath)?.SanitizePath();
+            if (string.IsNullOrEmpty(dirAsset))
+            {
+                return;
+            }
+            string fullDir = ToFullPath(dirAsset);
+            if (!Directory.Exists(fullDir))
```
wallstop: Is there TOCTOU here?

Editor/Sprites/TextureResizerWizard.cs
```diff
@@ -79,466 +72,43 @@ public static void ResizeTextures()
...
+            _ = TextureResizerAPI.TryResizeTextures(
+                textures,
+                sourceFolders,
+                numResizes,
+                scalingResizeAlgorithm,
+                pixelsPerUnit,
+                widthMultiplier,
+                heightMultiplier,
+                outputPath,
+                dryRun
```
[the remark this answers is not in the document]
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in 5453589e: the wizard now passes only valid folder search objects to the direct API, preserving selected-texture processing when a search object is a file. I swept the wizard-to-API boundary; the explicit API still rejects invalid caller-supplied folders. A native Unity regression test covers the file-search/selected-texture case (14/14 resizer fixture). No broader rule is needed beyond that test.

Editor/Sprites/TextureResizerAPI.cs
```diff
@@ -0,0 +1,636 @@
...
+                    || !(
+                        string.Equals(outputDirAssetPath, "Assets", StringComparison.Ordinal)
+                        || outputDirAssetPath.StartsWith("Assets/", StringComparison.Ordinal)
+                    )
+                )
+            )
+            {
+                Debug.LogError($"The output folder is invalid: {outputDirAssetPath}.");
+                return false;
+            }
```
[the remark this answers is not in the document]
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in 5453589e: the API normalizes both output and source asset-folder paths before validation and discovery. I swept both folder inputs and added a native Unity regression using backslash-separated paths; the resizer fixture passes 14/14. The regression test is the rule for this API, so no separate package-wide lint rule is warranted.

Runtime/Core/Extension/RandomExtensions.Subsets.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed/swept in 15e9695b: Queue and HashSet already copy directly into one owned array; I added measured Stack and LinkedList direct-copy paths. At 4,096 items, the new paths were about 3.5× faster than the delivery-based fallback in paired Unity probes. A pooled source snapshot cannot safely outlive this deferred, repeatable `IEnumerable` if the caller never enumerates it; the sampling reservoir itself is already pooled during enumeration. A general `ICollection<T>.Count/CopyTo` path is unsafe for caller-defined collections with forged counts, so the fallback grows only from delivered items. Seeded snapshot/order tests and the documented fallback are the guard; full native fixture 86/86.

Editor/Sprites/TextureResizerAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed/swept in e1d411d2: output validation is now one null-guarded branch. `null` means overwrite; non-null empty/whitespace, outside-Assets, and nonexistent folders fail there. Post-validation output branches now test only null versus non-null. The existing invalid-output regression remains the guard; native resizer fixture 14/14.

Editor/Sprites/TextureResizerAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed/swept in e1d411d2: scalar and algorithm checks now run before any AssetDatabase query; the output path's cheap Assets-prefix check precedes `IsValidFolder`. I also removed a redundant directory/AssetDB probe on the write path. The zero-pass test now supplies a nonexistent output folder and verifies settings validation wins. Native resizer fixture remained 14/14 without an extra case; one 10 s run versus an 11.44 s baseline is not enough to claim a speed gain. No separate rule is needed beyond the ordering test.

Editor/Sprites/TextureResizerAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed in e1d411d2: that extra source-folder scope had no lifetime purpose and is removed. I swept the adjacent pooled-resource lifetimes; each lease still spans its actual use. Existing resizer fixture 14/14 is the guard; no new rule is warranted for a redundant brace scope.

Editor/Sprites/TextureResizerAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed here in e1d411d2: resizer ordering uses `UnityObjectNameComparer<Texture2D>.Instance`, with natural name order documented. I swept the other name sorts: three animation sites intentionally use `EditorUtility.NaturalCompare`. In a native Unity check it orders `Frame2Idle` before `Frame10Idle`, while our shared comparer reverses them because it recognizes only trailing digits; replacing those sites would change frame order. I kept that semantic boundary as the rule rather than making an unsafe global replacement. Resizer fixture 14/14.

Editor/Sprites/TextureResizerAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Swept the resizer's array sites. `AssetDatabase.FindAssets` needs an exact-length `string[]` of folders; `SystemArrayPool` may return a larger array whose extra slots change that input, while exact variable-size pooling retains arbitrary bucket sizes for this small, infrequent editor call. `GetPixels` and `EncodeToPNG` return Unity-owned arrays that this API cannot safely pool. I therefore kept the exact owned folder array rather than adding a pooling rule without a measured benefit; the allocation-sensitive collection scratch already uses existing pools. Native resizer fixture 14/14.

Editor/Sprites/TextureResizerAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Swept the current progress cleanup and kept its zero-allocation `try/finally`, which clears exactly once on cancellation and callback failure (native cleanup fixture 14/14). A reusable public `IDisposable`/`using` scope changes EditorUi's cross-tool contract, especially nested ownership, double-dispose, cleanup exceptions, and boxing. I opened detailed child issue #843 to design and measure it before migrating tools. That issue is the follow-up rule; I did not expand this resizer PR speculatively.

Editor/Sprites/TextureResizerAPI.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Fixed/swept in e1d411d2: I removed the resizer's separate `Directory.Exists`/`CreateDirectory` path and its extra `EnsureAssetParentFolder` probe. The already-used `DurableFile.TryWriteAllBytes` creates the parent idempotently as part of its staged write and returns a write failure to this API. This keeps directory creation at one owner instead of adding a second TOCTOU check; the existing durable-write contract is the rule. Native resizer fixture 14/14.

Editor/Sprites/TextureResizerAPI.cs
wallstop: Yea but what about WallstopArrayPool?

wallstop: DISCLOSURE: LLM-GENERATED TEXT

I checked `WallstopArrayPool<string>` specifically. It returns an exact-length array, but each distinct requested length creates a bucket that retains an array; `sourcePaths.Count` is not bounded to a small set, so pooling this `FindAssets` argument could turn transient allocations into retained memory. In a native Unity 6000.4 benchmark (best of three 100k-iteration runs, filling and consuming the array), direct allocation was faster at lengths 1/4/16/64: 4.97/6.66/18.73/65.23 ms versus 17.19/19.08/28.91/70.53 ms with the pool. I swept the other `FindAssets` callers; they already use exact arrays. No code change here; I added the whole-array API and unbounded exact-size bucket rule to `.llm/skills/use-array-pool.md` in b69f4d37 to prevent this class of mistake.

Runtime/Core/Extension/RandomExtensions.Subsets.cs
```diff
@@ -66,8 +67,71 @@ int count
...
+                }
+
+                T[] snapshot = new T[linkedList.Count];
+                linkedList.CopyTo(snapshot, 0);
+                return NextSubsetIterator(random, snapshot, count);
+            }
+
             using PooledResource<List<T>> lease = Buffers<T>.List.Get(out List<T> materializedList);
-            materializedList.AddRange(items);
+            foreach (T item in items)
```
wallstop, line 131: Is this actually faster than AddRange?

Runtime/Core/Helper/ReflectionHelpers.TypeDiscovery.cs
wallstop, line 50: Follow up: We should ban LINQ here (and everywhere in prod/editor code) and make value-based enumerators (do not handle in this PR, create an issue)

Runtime/Core/Helper/ReflectionHelpers.TypeDiscovery.cs
```diff
@@ -119,28 +123,33 @@ public static Type TryResolveType(string typeName)
...
-                    try
+                    resolved = asm.GetType(typeName, throwOnError: false, ignoreCase: false);
+                    if (resolved != null)
                     {
-                        resolved = asm.GetType(typeName, throwOnError: false, ignoreCase: false);
-                        if (resolved != null)
-                        {
-                            break;
-                        }
+                        TypeResolutionCache[typeName] = resolved;
```
wallstop, line 145: Should any of this be, like, addorupdate or getoradd or similar? Here and elsewhere?

Runtime/Core/Extension/RandomExtensions.Subsets.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

I have not measured this loop as faster than `AddRange`, and am not claiming a speedup. This fallback accepts arbitrary caller-defined `IEnumerable<T>`/`ICollection<T>`; `List<T>.AddRange` takes the `ICollection<T>.Count`/`CopyTo` fast path when available. A source can report an inflated count, and `NextSubsetIgnoresUntrustedCollectionCount` covers exactly that case. The explicit loop sizes the owned snapshot from delivered items only, preserving the deferred sampler's source snapshot contract. The concrete Queue/Stack/HashSet/LinkedList paths already use direct `CopyTo` where their counts are trustworthy; no code change is needed here.

Runtime/Core/Helper/ReflectionHelpers.TypeDiscovery.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

Tracked as #847 (related to the broader allocation sweep in #781). It scopes the production/editor LINQ ban, value-based enumerator alternatives, an enforceable gate, and regression/performance checks. I have not implemented that follow-up in this PR, as requested.

Runtime/Core/Helper/ReflectionHelpers.TypeDiscovery.cs
wallstop: DISCLOSURE: LLM-GENERATED TEXT

I swept this cache and the adjacent reflection lookup caches. `TypeResolutionCache` is a `ConcurrentDictionary` outside `SINGLE_THREADED`; its indexer write is atomic and safe, and the value has already been resolved before the write. `AddOrUpdate` is unnecessary because no update transform is needed; `GetOrAdd` with a precomputed value would not avoid the resolution work (a factory can also run more than once under contention). The adjacent field/property/method/indexer caches already use `GetOrAdd` where resolution is part of the cache lookup. Unresolved and migrated composite names deliberately remain uncached. I found no correctness issue that warrants changing this write or the surrounding caches in this PR.

